PR #44, opened 2015-01-21
Fixes for serialization stream of graph datatypes.

ctsims:
Also added some better outputs during deserialziation process to better catch errors on the serialization/deserialization stream.

### DIFF
--- a/backend/src/org/commcare/suite/model/graph/BubbleSeries.java
+++ b/backend/src/org/commcare/suite/model/graph/BubbleSeries.java
@@ -18,6 +18,13 @@ import org.javarosa.xpath.parser.XPathSyntaxException;
 public class BubbleSeries extends XYSeries {
     private String mRadius;
     private XPathExpression mRadiusParse;
+    
+    /*
+     * Deserialization Only!
+     */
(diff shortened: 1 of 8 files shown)


backend/src/org/commcare/suite/model/graph/BubbleSeries.java
```diff
@@ -18,6 +18,13 @@
 public class BubbleSeries extends XYSeries {
     private String mRadius;
     private XPathExpression mRadiusParse;
+    
+    /*
+     * Deserialization Only!
```
wpride: Can you explain what these do for deserialization?

ctsims: @wpride For Sure!

The deserializtion engine has to be able to inflate a copy of each class using reflection, then it asks that class to deserialize itself against the current position in the input stream, as such it needs to have a constructor with no arguments to be able to instantiate a new copy.

In a world with no 1.3 compatibility, we'd likely want these constructors to be private and then we'd change their permissions through reflection so that the API was more clear, but so it goes.

PR #46, opened 2015-02-03
Multiple apps

amstone326:


czue:
:scream_cat:

orangejenny:
x-requested with https://github.com/dimagi/commcare-odk/pull/131

backend/src/org/commcare/suite/model/Profile.java
```diff
@@ -84,6 +88,21 @@ public Profile(int version, String authRef) {
...
+	 */
+	public String getUniqueId() {
+		return this.uniqueId;
+	}
+	
+	/*
+	 * Return whether this Profile was generated from an old version of profile.ccpr
+	 * (i.e. does not have a uniqueId or displayName)
+	 */
+	public boolean fromOld() {
```
orangejenny: Good that this is commented, but it's pretty vague. How about returning !fromOld and calling it multipleAppsEnabled?

backend/src/org/commcare/xml/ProfileParser.java
```diff
@@ -48,46 +47,57 @@ public Profile parse() throws InvalidStructureException, IOException, XmlPullPar
...
-        
-        Profile profile = new Profile(version, authRef);
+
+        // If this is an old version of the profile file and is therefore missing uniqueId,
+        // get it from the update URL
+        boolean fromOld = false;
+        if (uniqueId == null) {
+            fromOld = true;
+            int startIndex = authRef.indexOf("download") + 9;
+            int endIndex = authRef.indexOf("profile") - 1;
```
orangejenny: Paranoid, but perhaps this should use the indexOf(String str, int fromIndex) version and start from startIndex.

orangejenny:
Looks fine, just minor comments.

backend/src/org/commcare/suite/model/Profile.java
```diff
@@ -66,9 +69,12 @@ public Profile() {
      * is obsoleted by it. 
      * 
      */
-    public Profile(int version, String authRef) {
+    public Profile(int version, String authRef, String uniqueId, String displayName, boolean fromOld) {
         this.version = version;
         this.authRef = authRef;
+        this.uniqueId = uniqueId;
```
ctsims: these flags should be in the serialization, right?

phillipm:
Looks good after you address @ctsims serialization comment.

backend/src/org/commcare/suite/model/Profile.java
```diff
@@ -81,6 +92,30 @@ public int getID() {
...
+     * use the app name from Localization strings instead)
+     */
+	public String getDisplayName() {
+	    return this.displayName;
+	}
+
+    /**
+     * @return if this object was generated from an old version of the profile.ccpr file
+     */
+	public boolean oldVersion() {
```
ctsims: should be isOldVersion for conventions

tests/test/org/commcare/model/test/ProfileTests.java
```diff
@@ -83,4 +110,23 @@ private Profile getBasicProfile() {
             throw PersistableSandbox.wrap("Error during profile test setup", e);
         }
     }
+
+    // Return a Profile object constructed from a profile.ccpr file explicitly containing the
+    // fields necessary for multiple apps support
+    private Profile getMultipleAppsProfile() {
```
ctsims: This should probably just be the same method as above and take in a string arg that is the path

PR #47, opened 2015-02-04
make jenkins builds work with HQ

wpride:


application/build.xml
```diff
@@ -363,7 +363,6 @@
 		</zip>
 		<taskdef name="py-run" onerror="report" classname="org.pyant.tasks.PythonRunTask" classpath="${dir.tools}/pyAntTasks-1.3.jar"/>
 		<py-run script="${dir.postjad.new}/utilities/submit_build.py"/>
-		<delete dir="${build.zip.dir}"/>
```
ctsims: Do we need to move this to before the <zip> so that workspaces which already have the file around won't be overwritten? Did you try to run it twice in a row locally and make sure there weren't any issues?

wpride: no issues I can tell

PR #50, opened 2015-02-26
Module filtering

orangejenny:
Adds a `relevant` attribute to `<menu>` elements in suite.xml

### DIFF
--- a/application/src/org/commcare/util/CommCareSessionController.java
+++ b/application/src/org/commcare/util/CommCareSessionController.java
@@ -71,16 +71,28 @@ public class CommCareSessionController {
         entryTable.clear();
         menuTable.clear();
         Enumeration en = session.platform.getInstalledSuites().elements();
+        EvaluationContext ec  = session.getEvaluationContext(getIif());
         while(en.hasMoreElements()) {
             Suite suite = (Suite)en.nextElement();
             for(Menu m : suite.getMenus()) {
+                String menuRelevant = m.getRelevancyCondition();
+                if (menuRelevant != null) {
+                    XPathExpression parsed;
+                    try {
+                        parsed = XPathParseTool.parseXPath(menuRelevant);
+                        if(XPathFuncExpr.toBoolean(parsed.eval(ec)).booleanValue() == false) {
+                            continue;
+                        }
+                    } catch (XPathSyntaxException xpse) {
+                        throw new RuntimeException(xpse.getMessage());
+                    }
+                }
                 if(menu.equals(m.getId())){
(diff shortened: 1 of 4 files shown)


orangejenny:
cross-requested with https://github.com/dimagi/commcare-odk/pull/161

orangejenny:
reminding myself to update https://github.com/dimagi/commcare/wiki/Suite20 once this is reviewed/merged/released

application/src/org/commcare/util/CommCareSessionController.java
```diff
@@ -71,16 +71,16 @@ public void populateMenu(List list, String menu, MultimediaListener listener) {
         entryTable.clear();
         menuTable.clear();
         Enumeration en = session.platform.getInstalledSuites().elements();
+        EvaluationContext ec  = session.getEvaluationContext(getIif());
```
ctsims: I think this can have performance implications (IE: Requesting the ec) Maybe we should make it lazily evaluate?

backend/src/org/commcare/suite/model/Menu.java
```diff
@@ -78,6 +75,14 @@ public String getId() {
     }
     
     /**
+     * @return A string representing an XPath expression to determine
+     * whether or not to display this menu.
+     */
+    public String getRelevancyCondition() {
```
ctsims: This name is now pretty confusing with the one below. Maybe we should update getRelevantCondition() below to be something like "getCommandRelevantCondition()" to make it clear that they are doing different things?

backend/src/org/commcare/suite/model/Suite.java
```diff
@@ -77,6 +82,32 @@ public void setID(int ID) {
...
+        Vector<Menu> relevant = new Vector<Menu>();
+        for(Menu m : menus) {
+            String condition = m.getRelevancyCondition();
+            if (condition == null) {
+                relevant.addElement(m);
+            }
+            else {
+                XPathExpression parsed;
+                try {
+                    parsed = XPathParseTool.parseXPath(condition);
```
ctsims: this might have performance implications if we do it frequently. Parsing XPath expressions is a super costly operation. We solve this for the other commands by moving the parse code into the menu object itself and letting it soft cache the parsed out version. Would be good to replicate that process here if possible.

ctsims:
Two High level questions: 

1) Is this documented on the wiki format for the sutie file?
2) Do we have a clear path to getting the session data loaded up for the menus?

orangejenny:
@ctsims addressed PR comments & updated wiki. Holding off on making any changes related to your question #2 for now.

PR #56, opened 2015-04-01
Move element parser

phillipm:
Move ElementParser to javarosa repo, and create CommCareElementParser, which extends ElementParser with parseDisplayBlock, which uses commcare specific models.

Cross request with https://github.com/dimagi/commcare-odk/pull/210 and https://github.com/dimagi/javarosa/pull/65


backend/src/org/commcare/xml/CommCareElementParser.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public static int SEVERITY_UNKOWN = 4;
+    /**
+     * The profile is incompatible with the major version of the current CommCare installation *
+     */
+    public static int REQUIREMENT_MAJOR_APP_VERSION = 1;
+    /**
+     * The profile is incompatible with the minor version of the current CommCare installation *
+     */
+    public static int REQUIREMENT_MINOR_APP_VERSION = 2;
```
phillipm: Moved these static vars here from UnfullfilledRequirementsException.java

PR #58, opened 2015-04-01
Surface XPathSyntax exceptions in graphing code

orangejenny:


orangejenny:
cross-requested with https://github.com/dimagi/commcare-odk/pull/212

PR #65, opened 2015-04-09
Flexible Case Property Fix

ctsims:
The recent heterogeneous path fix changed the mechanism of checking templates from getChild to getChildWithName, which broke the case element's ability to flexibly support missing values.

Fixes for the case instance objects to continue supporting flexible operations with the new template query formats.


orangejenny:
Why do the nodes need to have case_type set when created within the cache, but not in the main getChildrenWithName function?

ctsims:
@orangejenny Because that's part of the template for the <index> ndoes

<index case_type="">

PR #69, opened 2015-04-21
Case callouts

wpride:


wpride:
WIP

backend/src/org/commcare/suite/model/Callout.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.Hashtable;
+import java.util.Vector;
+
+/**
+ * Created by wpride1 on 4/14/15.
+ */
```
phillipm: It would be great if you could add a class level doc explaining what this class is used for.

backend/src/org/commcare/suite/model/CalloutData.java
```diff
@@ -0,0 +1,42 @@
+package org.commcare.suite.model;
+
+import java.util.Hashtable;
+import java.util.Vector;
+
+/**
+ * Created by wpride1 on 4/17/15.
```
phillipm: Class comment

backend/src/org/commcare/suite/model/Callout.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public void addResponse(String key){
+        responses.add(key);
+    }
+
+    public Hashtable<String, String> getExtras(){
+        return extras;
+    }
+
+    public Vector<String> getResponses(){
```
phillipm: getResponses is never used in the code base, is it worth including (also getExtras, getDisplayName, getActionName)?

backend/src/org/commcare/suite/model/Callout.java
```diff
@@ -0,0 +1,100 @@
...
+    private final Hashtable<String, String> extras = new Hashtable<String, String>();
+    private final Vector<String> responses = new Vector<String>();
+
+    public Callout(String actionName, String image, String displayName) {
+        this.actionName = actionName;
+        this.image = image;
+        this.displayName = displayName;
+    }
+
+    @Override
```
phillipm: Explicit override annotations are useful because they statically check that your method signature matches the one you are overriding. Recommend using in the future

PR #81, opened 2015-05-12
Fix nullables displayunit

wpride:


backend/src/org/commcare/suite/model/DisplayUnit.java
```diff
@@ -70,18 +71,18 @@ public Text getAudioURI() {
      */
     public void readExternal(DataInputStream in, PrototypeFactory pf)
             throws IOException, DeserializationException {
-        name = (Text)ExtUtil.read(in, Text.class, pf);
-        imageReference = (Text)ExtUtil.read(in, Text.class, pf);
-        audioReference = (Text)ExtUtil.read(in, Text.class, pf);
+        name = (Text)ExtUtil.read(in, new ExtWrapNullable(Text.class));
```
ctsims: This shouldn't be nullable, should it? (not actually sure...)

wpride: yeah you're right... I think it might be reasonable in the future to allow this to be null IE you want just an image no text (actually discussed this with sravan) but I think at the moment we'd just crash later

PR #82, opened 2015-05-14
Bar graphs

orangejenny:


phillipm:
Gaah, I read those commits in the wrong order...

PR #84, opened 2015-05-25
Remove TransactionParser's name & namespace fields

phillipm:
Since TransactionParser and its extensions didn't use the following, remove them:
 -  'parses' method
 - 'name' and 'namespace' fields

Change TransactionParserFactory's getParser to not take in a name or namespace field. These were rarely used, and in the cases that they were, the name/namespace values were obtained from the parser argument, and hence were redundant.

Cross requested with https://github.com/dimagi/commcare-odk/pull/311

### DIFF
--- a/backend/src/org/commcare/data/xml/DataModelPullParser.java
+++ b/backend/src/org/commcare/data/xml/DataModelPullParser.java
@@ -119,14 +119,11 @@ public class DataModelPullParser extends ElementParser<Boolean> {
             }
 
             String name = parser.getName();
-            String namespace = parser.getNamespace();
-
-            int depth = parser.getDepth();
             if (name == null) {
                 continue;
             }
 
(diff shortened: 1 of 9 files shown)


backend/src/org/commcare/data/xml/DataModelPullParser.java
```diff
@@ -119,14 +119,11 @@ private void parseBlock(String root) throws InvalidStructureException, IOExcepti
...
 
             String name = parser.getName();
-            String namespace = parser.getNamespace();
-
-            int depth = parser.getDepth();
             if (name == null) {
                 continue;
             }
 
-            TransactionParser transaction = factory.getParser(name, namespace, parser);
```
phillipm: Since this is the only place that getParser is actually called, and it is always using the name and namespace from the 'parser' object, I've changed the signature of getParser to only take in the parser as an argument.

PR #86, opened 2015-05-28
Support pixel-based case list column widths

orangejenny:
http://manage.dimagi.com/default.asp?169125

Most of this is just storing style hints as strings instead of numbers, so I can keep the "%" to distinguish between percentages and pixels.

Cross-requested with javarosa and commcare-odk

### DIFF
--- a/application/src/org/commcare/entity/RecentFormEntity.java
+++ b/application/src/org/commcare/entity/RecentFormEntity.java
@@ -56,11 +56,11 @@ public class RecentFormEntity extends Entity<FormInstance> {
         return true;
     }
 
-    public int[] getStyleHints (boolean header) {
+    public String[] getStyleHints (boolean header) {
         if(header) {
-            return new int[] {-1, -1} ;
+            return new String[] {null, null} ;
         } else {
-            return new int[] {-1, 36 };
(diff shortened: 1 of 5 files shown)


application/src/org/commcare/entity/RecentFormEntity.java
```diff
@@ -56,11 +56,11 @@ public boolean match (String key) {
...
     }
 
-    public int[] getStyleHints (boolean header) {
+    public String[] getStyleHints (boolean header) {
         if(header) {
-            return new int[] {-1, -1} ;
+            return new String[] {null, null} ;
         } else {
-            return new int[] {-1, 36 };
+            return new String[] {null, "36%" };
```
orangejenny: @ctsims Do you know offhand what this code is for? J2ME case list header row has a 36% column? I'm looking at https://confluence.dimagi.com/download/attachments/10223880/case%20select.png?version=1&modificationDate=1342853926863&api=v2 but not sure what the 36 is for.

orangejenny:
https://github.com/dimagi/commcare-odk/pull/314
https://github.com/dimagi/javarosa/pull/99

PR #88, opened 2015-06-01
Helper method for mockEasiestRoute

phillipm:
Used for fixing mockEasiestRoute as noted in the cross-requested PR https://github.com/dimagi/commcare-odk/pull/321

### DIFF
--- a/backend/src/org/commcare/resources/model/ResourceTable.java
+++ b/backend/src/org/commcare/resources/model/ResourceTable.java
@@ -136,7 +136,7 @@ public class ResourceTable {
     }
 
     public void addResource(Resource resource, ResourceInstaller initializer,
-                            String parentId) throws StorageFullException {
+                            String parentId) {
         addResource(resource, initializer, parentId, Resource.RESOURCE_STATUS_UNINITIALIZED);
     }
 
--- a/backend/src/org/commcare/resources/model/ResourceTable.java
+++ b/backend/src/org/commcare/resources/model/ResourceTable.java
@@ -126,17 +126,17 @@ public class ResourceTable {
         storage.remove(resource);
     }
 
-    public void addResource(Resource resource, ResourceInstaller initializer, String parentId, int status) throws StorageFullException {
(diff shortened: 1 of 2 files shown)


backend/src/org/commcare/resources/model/ResourceTable.java
```diff
@@ -126,17 +126,17 @@ public void removeResource(Resource resource) {
...
         resource.setParentId(parentId);
         addResource(resource, status);
     }
 
-    public void addResource(Resource resource, ResourceInstaller initializer, String parentId) throws StorageFullException {
+    public void addResource(Resource resource, ResourceInstaller initializer, String parentId) {
         addResource(resource, initializer, parentId, Resource.RESOURCE_STATUS_UNINITIALIZED);
     }
 
-    public void addResource(Resource resource, int status) throws StorageFullException {
```
phillipm: Sneaking this change into this PR:
addResource never throws the StorageFullException runtime exception, since it explicitly swallows it. Hence we should take it out of the method signature completely.

backend/src/org/commcare/util/CommCarePlatform.java
```diff
@@ -346,6 +346,29 @@ public void initialize(ResourceTable global) {
...
+     */
+    public String getModuleNameForEntry(Entry entry) {
+        Vector<Suite> installed = getInstalledSuites();
+
+        for (Suite s : installed) {
+            for (Entry suiteEntry : s.getEntries().values()) {
+                if (suiteEntry.getCommandId().equals(entry.getCommandId())) {
+                    // XXX PLM: unsure what to do when there are multiple
+                    // entries in getMenus()
+                    return s.getMenus().firstElement().getId();
```
phillipm: @ctsims do you know when multiple entries in getMenus would ever occur?

ctsims: So the HQ app builder doesn't really support this, but you can put an <entry> into any number of modules/menus, which is super helpful for certain use cases.

phillipm:
@wpride Can you review this one? Thanks

PR #90, opened 2015-06-10
isViewCommand checking for multiple datums in a session

dcluna:
Use case in [Commcare-167204](http://manage.dimagi.com/default.asp?167204).

Also including a few toString implementations used during debugging.

Cross: https://github.com/dimagi/commcare-odk/pull/327

### DIFF
--- a/backend/src/org/commcare/suite/model/Callout.java
+++ b/backend/src/org/commcare/suite/model/Callout.java
@@ -15,6 +15,7 @@ import org.javarosa.xpath.parser.XPathSyntaxException;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.Vector;
@@ -123,5 +124,16 @@ public class Callout implements Externalizable, DetailTemplate {
 
     public Vector<String> getResponses() {
         return responses;
+    }
+
+    @Override
+    public String toString() {
+        return "Callout{" +
+                "actionName='" + actionName + '\'' +
+                ", image='" + image + '\'' +
+                ", displayName='" + displayName + '\'' +
+                ", extras=" + Arrays.toString(extras.entrySet().toArray()) +
+                ", responses=" + responses +
+                '}';
     }
 }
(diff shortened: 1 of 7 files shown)


backend/src/org/commcare/suite/model/Callout.java
```diff
@@ -124,4 +125,15 @@ public void addResponse(String key) {
...
+    @Override
+    public String toString() {
+        return "Callout{" +
+                "actionName='" + actionName + '\'' +
+                ", image='" + image + '\'' +
+                ", displayName='" + displayName + '\'' +
+                ", extras=" + Arrays.toString(extras.entrySet().toArray()) +
+                ", responses=" + responses +
+                '}';
+    }
```
phillipm: :+1: for line-breaks and override annotation :)

backend/src/org/commcare/util/CommCareSession.java
```diff
@@ -721,6 +721,21 @@ public boolean isViewCommand(String command) {
         // commcare core for release issues
         return (entries.size() == 1 &&
                 (prototype.getXFormNamespace() == null ||
-                        prototype.getXFormNamespace().equals("")));
+                        prototype.getXFormNamespace().equals("")) &&
```
phillipm: In my opinion it is good to call "equals" on the concrete string to avoid null pointer exceptions. For instance "".equals(prototype.getXFormNamespace())

phillipm: (in this case it is obvious that a NPE couldn't happen, but still...)

backend/src/org/commcare/util/CommCareSession.java
```diff
@@ -721,6 +721,21 @@ public boolean isViewCommand(String command) {
         // commcare core for release issues
```
phillipm: This comment makes me sad! Can we figure out if the "".equals check is needed or not and address the comment?

backend/src/org/commcare/util/CommCareSession.java
```diff
@@ -721,6 +721,21 @@ public boolean isViewCommand(String command) {
...
         return (entries.size() == 1 &&
                 (prototype.getXFormNamespace() == null ||
-                        prototype.getXFormNamespace().equals("")));
+                        prototype.getXFormNamespace().equals("")) &&
+                // Commcare-167204 by dcluna on 06/10/2015: If we have more than 1 datum for this Entry, we want to show them all
+                prototype.getSessionDataReqs().size() < 2);
```
phillipm: Is it possible to move this comment into the methods documentation?

ctsims: This should actually compare the session data requirements against the current set of session data.

phillipm:
Solved by https://github.com/dimagi/commcare/pull/150

PR #97, opened 2015-07-02
log and skip missing cases

wpride:


wpride:
fix for http://manage.dimagi.com/default.asp?175352

mostly for @ctsims to review

This will just skip this index IE not try and process it during purging. I don't think this will be necessarily bad, but it seems like we should probably not fail so silently when their OTA restore is this jacked. Ideally this is fixed on the HQ side. Do you think its reasonable to fail the restore completely in this case? Or should we just continue?

phillipm:
Can we close this? It looks like it got fixed on the server side of things.

wpride:
@phillipm The server side check is imperfect - this just came up again yesterday. Think we still want this

ctsims:
Hey,

This is a tough situation to resolve for sure. 

Will: at the very least can you make the "error" tag here a very, very specific and let Cory know (he's working on 412 flagging) what the tag is so we can log it very deliberately on HQ and get ongoing metrics on this?

wpride:
@ctsims @czue added specific tag

PR #99, opened 2015-07-13
Check if commandId is null in getEntriesForCommand to prevent NPE

phillipm:
Trying to solve this crash from the google play store crash reports
```
java.lang.RuntimeException: Unable to start activity ComponentInfo{org.commcare.dalvik/org.commcare.dalvik.activities.EntitySelectActivity}: java.lang.NullPointerException
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:1659)
at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:1675)
at android.app.ActivityThread.access$1500(ActivityThread.java:121)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:943)
at android.os.Handler.dispatchMessage(Handler.java:99)
at android.os.Looper.loop(Looper.java:130)
at android.app.ActivityThread.main(ActivityThread.java:3701)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:507)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:866)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:624)
at dalvik.system.NativeStart.main(Native Method)
Caused by: java.lang.NullPointerException
at org.commcare.util.CommCareSession.getEntriesForCommand(CommCareSession.java:79)
at org.commcare.util.CommCareSession.getEntriesForCommand(CommCareSession.java:79)
at org.commcare.util.CommCareSession.getEntriesForCommand(CommCareSession.java:68)
at org.commcare.util.CommCareSession.getNeededDatum(CommCareSession.java:223)
at org.commcare.dalvik.activities.EntitySelectActivity.onCreate(EntitySelectActivity.java:162)
at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1047)
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:1623)
... 11 more
```

EntitySelectActivity calls getNeededDatum https://github.com/dimagi/commcare-odk/blob/master/app/src/org/commcare/dalvik/activities/EntitySelectActivity.java#L168, which results in an NPE because the currentCmd field is null, resulting in commandId being null https://github.com/dimagi/commcare/blob/master/backend/src/org/commcare/util/CommCareSession.java#L79 

This PR prevents the null pointer exception. I am uncertain if just preventing this crash is the right way to go, perhaps I should be addressing the logical root of the problem?

### DIFF
--- a/backend/src/org/commcare/util/CommCareSession.java
+++ b/backend/src/org/commcare/util/CommCareSession.java
@@ -225,8 +225,12 @@ public class CommCareSession {
      * an entry on the stack
      */
     public SessionDatum getNeededDatum() {
-        Entry entry = getEntriesForCommand(getCommand()).elementAt(0);
-        return getNeededDatum(entry);
+        Vector<Entry> entries = getEntriesForCommand(getCommand());
+        if (entries.size() > 0) {
+            return getNeededDatum(entries.elementAt(0));
+        } else {
+            return null;
+        }
     }
 
     /**
@@ -235,14 +239,13 @@ public class CommCareSession {
      */
     public SessionDatum getNeededDatum(Entry entry) {
         int nextVal = getData().size();
-        //If we've already retrieved all data needed, return null.
         if (nextVal >= entry.getSessionDataReqs().size()) {
+            // we've already retrieved all data needed.
             return null;
         }
 
-        //Otherwise retrieve the needed value
-        SessionDatum datum = entry.getSessionDataReqs().elementAt(nextVal);
-        return datum;
+        // retrieve the needed value
+        return entry.getSessionDataReqs().elementAt(nextVal);
     }
 
     public Detail getDetail(String id) {
@@ -431,7 +434,6 @@ public class CommCareSession {
      * @return Evaluation context for a command in the installed app
      */
     public EvaluationContext getEvaluationContext(InstanceInitializationFactory iif, String command) {
-
         if (command == null) {
             return new EvaluationContext(null);
         }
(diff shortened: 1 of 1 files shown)


backend/src/org/commcare/util/CommCareSession.java
```diff
@@ -715,12 +724,14 @@ public void markCurrentFrameForDeath() {
...
+        if (entries.size() == 1) {
+            Entry prototype = entries.elementAt(0);
 
-        // NOTE: We shouldn't need the "" here, but we're avoiding making changes to
-        // commcare core for release issues
-        return (entries.size() == 1 &&
-                (prototype.getXFormNamespace() == null ||
-                        prototype.getXFormNamespace().equals("")));
+            // NOTE: We shouldn't need the "" here, but we're avoiding making changes to
+            // commcare core for release issues
```
phillipm: @ctsims do you know what is up with this comment?

phillipm:
@ctsims Can you weigh in on whether preventing this crash is the right way to go? Or would it be better to make sure that getNeededDatum is never called when the session's state is such that CommCareSession.currentCmd is null?

ctsims:
@phillipm Hm, a tricky issue for sure.

My guess is that it's going to be dangerous to swallow the error, versus failing in a more explicit/semantically actionable manner. I definitely think of it as being part of the contract for getNeededDatum() that a datum is needed in the current state.

My guess is that the right behavior is to throw an exception here that reflects an invalid operation on the current state and to catch that in the Activity to do something helpful with it. The issue here is almost certainly that the session is getting cleared/restarted and then the Activity is relaunched, I'd imagine?

phillipm:
I'm going to close this but make a note to revisit it at some point

PR #103, opened 2015-07-16
Install/upgrade related comments

phillipm:
Comments for methods that deal with resource installation

### DIFF
--- a/backend/src/org/commcare/util/CommCarePlatform.java
+++ b/backend/src/org/commcare/util/CommCarePlatform.java
@@ -50,7 +50,14 @@ public class CommCarePlatform implements CommCareInstance {
         this.minorVersion = minorVersion;
     }
 
-    public void init(String profileReference, ResourceTable global, boolean forceInstall) throws UnfullfilledRequirementsException, UnresolvedResourceException {
+    /**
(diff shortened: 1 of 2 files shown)


backend/src/org/commcare/util/CommCarePlatform.java
```diff
@@ -50,7 +50,20 @@ public CommCarePlatform(int majorVersion, int minorVersion) {
...
+    /**
+     * Installs resources described by profile reference into the provided
+     * resource table. If the resource table is ready or already has a profile,
+     * don't do anything.
+     *
+     * @param profileReference URL to profile file
+     * @param global           Add profile ref to this table and install its
+     *                         resources
+     * @param forceInstall     Should installation be performed regardless of
+     *                         version numbers?
```
phillipm: Is this the correct comment for 'forceInstall'? It is passed along all the way down to its cryptic use in ProfileParser (https://github.com/dimagi/commcare/blob/master/backend/src/org/commcare/xml/ProfileParser.java#L64)

amstone326: Yes, looks correct to me

PR #111, opened 2015-08-03
Case XML end to end regressions

ctsims:
Tests loading case transactions in/out of XML

### DIFF
--- a/tests/test/org/commcare/cases/test/BadCaseXMLTests.java
+++ b/tests/test/org/commcare/cases/test/BadCaseXMLTests.java
@@ -1,0 +1,41 @@
+package org.commcare.cases.test;
+
+import org.commcare.resources.model.Resource;
+import org.commcare.suite.model.Profile;
+import org.commcare.test.utils.PersistableSandbox;
+import org.commcare.util.mocks.MockDataUtils;
+import org.commcare.util.mocks.MockUserDataSandbox;
+import org.commcare.xml.ProfileParser;
+import org.javarosa.core.util.ArrayUtilities;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.InputStream;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+/**
+ * Test suite to verify end-to-end parsing of inbound case XML
+ * and reading values back from the casedb model
+ * 
+ * @author ctsims
+ */
+public class BadCaseXMLTests {
+
+    MockUserDataSandbox sandbox;
+    
+    @Before
+    public void setUp() {
+        sandbox = MockDataUtils.getStaticStorage();
+    }
+    
+    @Test
+    public void testNoCaseID() {
+        //Expected - Fail silently (TODO: Fix?)
+        MockDataUtils.parseIntoSandbox(this.getClass().getResourceAsStream("/case_create_broken_no_caseid.xml"), sandbox);
+
+        //Make sure that we didn't make a case entry for the bad case though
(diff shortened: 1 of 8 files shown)


phillipm:

tests/test/org/commcare/cases/test/CaseParseAndReadTest.java
```diff
@@ -0,0 +1,79 @@
...
+        Document parsed = XmlComparator.getDocumentFromStream(new ByteArrayInputStream(parsedDb));
+        Document loaded = XmlComparator.getDocumentFromStream(this.getClass().getResourceAsStream(caseDbState));
+
+        try {
+            XmlComparator.compareXmlDOMs(parsed, loaded);
+        } catch (Exception e) {
+            System.out.print(new String(parsedDb));
+
+
+            assertEquals("CaseDB output did not match expected structure(" + e.getMessage() + ")", new String(dumpStream(caseDbState)), new String(parsedDb));
```
phillipm: This should be moved into the body of the ```try```

tests/test/org/commcare/test/utils/XmlComparator.java
```diff
@@ -0,0 +1,91 @@
...
+            parser.setInput(reader);
+            parser.setFeature(XmlPullParser.FEATURE_PROCESS_NAMESPACES, false);
+            document.parse(parser);
+            return document;
+        } catch(Exception e) {
+            throw new RuntimeException(e.getMessage());
+        }
+    }
+
+    public static void compareXmlDOMs(Document a, Document b) throws RuntimeException {
```
phillipm: Please add a comment describing the matching conditions: Does order matter, are values checked, etc.

phillipm: Shouldn't this method return a boolean and be renamed to xmlDOMsEqual? Or you leave the name the same and return -1, 0, 1, making it a comparator

tests/test/org/commcare/cases/test/BadCaseXMLTests.java
```diff
@@ -0,0 +1,31 @@
...
+        sandbox = MockDataUtils.getStaticStorage();
+    }
+
+    @Test
+    public void testNoCaseID() {
+        //Expected - Fail silently (TODO: Fix?)
+        MockDataUtils.parseIntoSandbox(this.getClass().getResourceAsStream("/case_create_broken_no_caseid.xml"), sandbox);
+
+        //Make sure that we didn't make a case entry for the bad case though
+        //assertEquals("Case XML with no id should not have created a case record", sandbox.getCaseStorage().getNumRecords(), 0);
```
phillipm: I've created a ticket for this: 
http://manage.dimagi.com/default.asp?178270

phillipm:
@ctsims Good to go after addressing my inline comments.

tests/test/org/commcare/cases/test/CaseParseAndReadTest.java
ctsims: Hey,

So this is actually basically an "assertFalse" that happens to outline the correct sections of the xml more often than not.

The two strings aren't actually guaranteed to be the same even _if_ the XML is semantically equivalent, but if they _are_ different, the strings definitely will be, and the areas where they are different get highlighted quite nicely in my tests

tests/test/org/commcare/test/utils/XmlComparator.java
ctsims: agreed. Will document

ctsims:
@phillipm Addressed comments and fixed some linting

phillipm:
:+1:

PR #117, opened 2015-08-07
Auto-update checks in background

phillipm:
CommCare related changes to enable ODK auto-update checks to happen in the background.

 - Hoist resource table downloading logic out of ```CommCarePlatform.upgrade``` and into ```prepareUpgradeResources```
 - Move install/update logic into a ```ResourceManager``` class that abstracts most access to resource tables

Cross-requested with https://github.com/dimagi/commcare-odk/pull/507

List of TODOs:
 - Test on J2ME device


backend/src/org/commcare/resources/model/Resource.java
```diff
@@ -75,10 +72,6 @@
     // Resource is in the local environment and ready to install
     public static final int RESOURCE_STATUS_LOCAL = 1;
 
-    // TODO: I think this means the same as "upgade" and should be removed and
-    // replaced by it
-    public static final int RESOURCE_STATUS_PENDING = 2;
```
phillipm: This was pretty much unused: it still had references but nothing setting records to this status.
I've removed the references, and this entry, but I'm not changing the numbering system here because it would require a database migration, which I don't think is worth the effort.

backend/src/org/commcare/resources/ResourceManager.java
```diff
@@ -0,0 +1,377 @@
...
+        if (masterTable.getTableReadiness() != ResourceTable.RESOURCE_TABLE_INSTALLED) {
+            repair();
+
+            if (masterTable.getTableReadiness() != ResourceTable.RESOURCE_TABLE_INSTALLED) {
+                throw new IllegalArgumentException("Global resource table was not ready for upgrading");
+            }
+        }
+    }
+
+    protected void loadProfile(ResourceTable table,
```
amstone326: Maybe rename to loadProfileIntoTable(), for clarity

backend/src/org/commcare/resources/model/ResourceTable.java
```diff
@@ -419,83 +433,64 @@ public void prepareResources(ResourceTable master, CommCareInstance instance)
      * @throws UnfullfilledRequirementsException resource(s) incompatible with
      *                                           current CommCare version
      */
-    public void prepareResources(ResourceTable master,
-                                 CommCareInstance instance,
-                                 String toInitialize)
-            throws UnresolvedResourceException, UnfullfilledRequirementsException {
+    public void prepareResourcesUpTo(ResourceTable master,
```
amstone326: Kind of an awkwardly named/used method. Since it's only getting used once for a very specific purpose (to prepare the table with just the profile), might be good to rename/rewrite to do that explicitly.

backend/src/org/commcare/resources/ResourceManager.java
```diff
@@ -0,0 +1,377 @@
...
+
+    public boolean isUpgradeTableStaged() {
+        return isTableStagedForUpgrade(upgradeTable);
+    }
+
+    /**
+     * @return True if profile argument points to an app version that isn't
+     * any newer than the profile in the upgrade table.
+     */
+    public boolean updateIsntNewer(Resource currentProfile) {
```
amstone326: Could think about renaming this like we talked about. Even updateNotNewer sounds less awkward

backend/src/org/commcare/resources/model/ResourceTable.java
```diff
@@ -141,29 +142,33 @@ public void addResource(Resource resource, ResourceInstaller initializer,
     }
 
     public void addResource(Resource resource, int status) {
-        // only add resource if they don't already exist
         if (storage.getIDsForValue(Resource.META_INDEX_RESOURCE_ID,
```
amstone326: If you're going to delete that comment, maybe factor this check out into a function called like resourceDoesNotExist()

backend/src/org/commcare/resources/model/ResourceTable.java
phillipm: Changing the name to `prepareProfileResource` would make me want to avoid taking in a `toInitialize` argument, which would require the `ResourceTable` to import the profile reference key from `CommCarePlatform`. Importing very  specific info to a generic class like `ResourceTable` feels a little wrong. I'm going to create a helper method in `ResourceManager` with a better name that calls `prepareResourcesUpTo`

PR #118, opened 2015-08-07
Empty Case ID parser fix + Test

ctsims:
Fixed empty caseid attribute being recorded as empty strings instead of null values, and associated test


phillipm:
Nice!

PR #126, opened 2015-08-11
CLI Tool Improvements

ctsims:
Fixed app level fixtures, added updates, and cleaned up a few things.

### DIFF
--- a/util/src/org/commcare/util/CommCareConfigEngine.java
+++ b/util/src/org/commcare/util/CommCareConfigEngine.java
@@ -386,7 +386,6 @@ public class CommCareConfigEngine {
 
     final static private class QuickStateListener implements TableStateListener{
         int lastComplete = 0;
-        int lastTotal = 0;
 
         @Override
         public void resourceStateUpdated(ResourceTable table) {
@@ -411,23 +410,6 @@ public class CommCareConfigEngine {
 
         Profile profileObj = this.getPlatform().getCurrentProfile();
 
-        boolean appInstalled = (profileRef != null &&
-                profileRef.getStatus() == Resource.RESOURCE_STATUS_INSTALLED);
-
-        TableStateListener globalListener = new TableStateListener() {
-            int lastComplete = 0;
-
-            @Override
-            public void resourceStateUpdated(ResourceTable table) {
-
-            }
-
-            @Override
-            public void incrementProgress(int complete, int total) {
-
-            }
-        };
-
         global.setStateListener(new QuickStateListener());
 
         updateTable.setStateListener(new QuickStateListener());
@@ -445,7 +427,6 @@ public class CommCareConfigEngine {
             // app (instead of latest release), add it to the query tags of
             // the profile reference
             if (forceNew &&
-                    (authUrl != null) &&
                     ("https".equals(authUrl.getProtocol()) ||
                             "http".equals(authUrl.getProtocol()))) {
                 if (authUrl.getQuery() != null) {
--- a/util/src/org/commcare/util/CommCareConfigEngine.java
+++ b/util/src/org/commcare/util/CommCareConfigEngine.java
@@ -64,10 +64,10 @@ import java.util.zip.ZipFile;
 public class CommCareConfigEngine {
     private OutputStream output;
     private ResourceTable table;
+    private ResourceTable updateTable;
+    private ResourceTable recoveryTable;
     private PrintStream print;
     private final CommCarePlatform platform;
-    private Vector<Suite> suites;
-    private Profile profile;
     private int fileuricount = 0;
     
     private ArchiveFileRoot mArchiveRoot;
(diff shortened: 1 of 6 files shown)


util/src/org/commcare/util/CommCareConfigEngine.java
```diff
@@ -379,4 +382,123 @@ private void print(Suite s, Entry e, int level) {
             }
         }
     }
+
+
+    final static private class QuickStateListener implements TableStateListener{
+        int lastComplete = 0;
+        int lastTotal = 0;
```
phillipm: Unused

util/src/org/commcare/util/CommCareConfigEngine.java
```diff
@@ -379,4 +382,123 @@ private void print(Suite s, Entry e, int level) {
...
+
+    public void attemptAppUpdate(boolean forceNew) {
+        ResourceTable global = table;
+
+        // Ok, should figure out what the state of this bad boy is.
+        Resource profileRef = global.getResourceWithId(CommCarePlatform.APP_PROFILE_RESOURCE_ID);
+
+        Profile profileObj = this.getPlatform().getCurrentProfile();
+
+        boolean appInstalled = (profileRef != null &&
```
phillipm: Unused

util/src/org/commcare/util/CommCareConfigEngine.java
```diff
@@ -379,4 +382,123 @@ private void print(Suite s, Entry e, int level) {
...
+                profileRef.getStatus() == Resource.RESOURCE_STATUS_INSTALLED);
+
+        TableStateListener globalListener = new TableStateListener() {
```
phillipm: Unused

util/src/org/commcare/util/CommCareConfigEngine.java
```diff
@@ -379,4 +382,123 @@ private void print(Suite s, Entry e, int level) {
...
+                    ("https".equals(authUrl.getProtocol()) ||
+                            "http".equals(authUrl.getProtocol()))) {
+                if (authUrl.getQuery() != null) {
+                    // If the profileRef url already have query strings
+                    // just add a new one to the end
+                    authRef = authRef + "&target=build";
+                } else {
+                    // otherwise, start off the query string with a ?
+                    authRef = authRef + "?target=build";
+                }
```
phillipm: This code is copied from odk's ResourceEngineTask. Should we centralize it in the commcare repo?

util/src/org/commcare/util/CommCareConfigEngine.java
```diff
@@ -379,4 +382,123 @@ private void print(Suite s, Entry e, int level) {
...
+            URL authUrl = new URL(authRef);
+
+            // profileRef couldn't be parsed as a URL, so don't worry
+            // about adding dev flags to the url's query
+
+            // If we want to be using/updating to the latest build of the
+            // app (instead of latest release), add it to the query tags of
+            // the profile reference
+            if (forceNew &&
+                    (authUrl != null) &&
```
phillipm: It isn't possible for authUrl to be null because constructors can never return null values

util/src/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -43,6 +43,10 @@
     private MockUserDataSandbox mSandbox;
     private SessionWrapper mSession;
 
+    boolean mUpdatePending = false;
+    boolean mForceLatestUpdate = false;
```
phillipm: These can be private

util/src/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -94,7 +106,19 @@ private boolean loopSession() throws IOException {
...
+                        mUpdatePending = true;
+
+                        if(input.contains("-f")) {
+                            mForceLatestUpdate = true;
+                        }
+                        return true;
+                    }
 
+                    if(input.equals(":home")) {
+                        return true;
```
phillipm: This method, ```loopSession```,  always returns ```true```. Is this expected? Could we just make it return void instead?

phillipm:
Done reviewing. Good to go after addressing dead code comments, and I'm fine with leaving in the duplicate code because I'm currently refactoring that code on the odk side and can centralize it in my PR.

util/src/org/commcare/util/cli/ApplicationHost.java
ctsims: oh, whoops. "Exit" is actually supposed to return false. will update

util/src/org/commcare/util/CommCareConfigEngine.java
ctsims: they use different URL objects unfortunately, but yes, we should definitely find a wy to centralize this logic

PR #127, opened 2015-08-12
Make case & ledger queries fail silently when no data has been loaded

phillipm:
XPath case and ledger queries should fail silently when no case or ledger data has been loaded but the base path is of the form:
 - ```instance('casedb')/casedb/case[@case_id = 'anything']```
 - ```instance('ledger')/ledgerdb/ledger[@entity-id='anything']```

Should address issues brought up in http://manage.dimagi.com/default.asp?177247


tests/test/org/commcare/cases/test/CaseXPathQueryTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    public void referenceNonExistentCaseId() {
+        MockDataUtils.parseIntoSandbox(this.getClass().getResourceAsStream("/case_create_basic.xml"), sandbox);
+        EvaluationContext ec =
+                MockDataUtils.getInstanceContexts(sandbox, "casedb", CaseTestUtils.CASE_INSTANCE);
+
+        // TODO PLM: Should this expression throw an exception because the
+        // case_id doesn't exist in the casedb?
+        Assert.assertTrue(CaseTestUtils.xpathEval(ec, "instance('casedb')/casedb/case[@case_id = 'no-case']", ""));
```
phillipm: What is the expected behavior of this query? It references a non-existent case id, should it fail silently or not?

cases/src/org/commcare/cases/ledger/instance/LedgerInstanceTreeElement.java
```diff
@@ -93,7 +93,7 @@ public void attachStringCache(Interner<String> stringCache) {
      */
     public LedgerChildElement getChild(String name, int multiplicity) {
         if (multiplicity == TreeReference.INDEX_TEMPLATE) {
-            return null;
+            return LedgerChildElement.TemplateElement(this);
```
phillipm: The fix; not sure if this has widespread implications or not.

Currently changes the code path in ```DataInstance.hasTemplatePathRec``` to detect a template path even when no ledgers have been loaded via ```currentNode.getChildrenWithName```, which in turn calls ```LedgerInstanceTreeElement.getLedgers```

cases/src/org/commcare/cases/instance/CaseInstanceTreeElement.java
```diff
@@ -107,7 +107,7 @@ public void attachStringCache(Interner<String> stringCache) {
      */
     public CaseChildElement getChild(String name, int multiplicity) {
         if (multiplicity == TreeReference.INDEX_TEMPLATE) {
-            return null;
+            return CaseChildElement.TemplateElement(this);
```
phillipm: Fix for case related queries failing in the wrong way when no case data has been loaded

cases/src/org/commcare/cases/ledger/instance/LedgerInstanceTreeElement.java
```diff
@@ -93,7 +93,7 @@ public void attachStringCache(Interner<String> stringCache) {
      */
     public LedgerChildElement getChild(String name, int multiplicity) {
         if (multiplicity == TreeReference.INDEX_TEMPLATE) {
```
ctsims: We should be checking thiis for both name and multiplicity.

IE: we only wanna return the ledger/case template if the request on templates was for a "ledger" template, not a "test" tempalte

cases/src/org/commcare/cases/instance/CaseInstanceTreeElement.java
ctsims: Same as below

tests/test/org/commcare/cases/CaseTestUtils.java
```diff
@@ -0,0 +1,58 @@
...
+import org.javarosa.xpath.parser.XPathSyntaxException;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+public class CaseTestUtils {
+    public static final String CASE_INSTANCE = "jr://instance/casedb";
+    public static final String LEDGER_INSTANCE = "jr://instance/ledgerdb";
+
+    public static byte[] loadCaseInstanceIntoSandbox(MockUserDataSandbox sandbox) {
```
ctsims: This is a confusingly named method. Should be

serializeCaseInstanceFromSandbox

or something similar

tests/test/org/commcare/cases/CaseTestUtils.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public static boolean xpathEval(EvaluationContext evalContext,
+                                    String input,
+                                    Object expectedOutput) {
+        XPathExpression expr;
+        try {
+            expr = XPathParseTool.parseXPath(input);
+        } catch (XPathSyntaxException e) {
+            e.printStackTrace();
```
ctsims: Hm, should this signature just throw these rather than swallow them? Otherwise the traces won't have them

tests/test/org/commcare/cases/CaseTestUtils.java
```diff
@@ -0,0 +1,58 @@
...
+            return false;
+        }
+        return expectedOutput.equals(output);
+    }
+
+    public static void xpathEvalWithException(EvaluationContext evalContext,
+                                              String input)
+            throws XPathSyntaxException {
+        XPathExpression expr = XPathParseTool.parseXPath(input);
+        XPathFuncExpr.unpack(expr.eval(evalContext));
```
ctsims: If the purpose here is to just generally use this when an exception is expected, maybe it should explicitly fail if one isn't?

tests/test/org/commcare/cases/test/CaseXPathQueryTest.java
ctsims: Oh, hm, good question. I think given that it isn't dereferenced to another value, but rather requested raw it is likely fine for it to return "".

util/src/org/commcare/util/mocks/MockDataUtils.java
```diff
@@ -141,8 +141,7 @@ public static FormInstance loadFixture(MockUserDataSandbox sandbox,
     }
 
     /**
-     * A quick way to request an evaluation context with an abstract instance available.
-     *
+     * Create an evaluation context with an abstract instance available.
      */
     public static EvaluationContext getInstanceContexts(MockUserDataSandbox sandbox, String instanceId, String instanceRef){
         InstanceInitializationFactory iif = new CommCareInstanceInitializer(sandbox);
```
ctsims: This should get generalized to call the mehtod below after creating a hashtable with its arguments, I think

tests/test/org/commcare/cases/CaseTestUtils.java
phillipm: Renamed the method to ```xpathEval``` so that it doesn't explicitly have anything to do with exceptions, letting the tests handle all the expected exception logic and failing test reporting.

util/src/org/commcare/util/mocks/MockDataUtils.java
phillipm: Good call

tests/test/org/commcare/cases/CaseTestUtils.java
phillipm: Oops, named that back when I was super confused about what it was actually doing :)

czue:
woot! thanks for quick turnaround!

PR #130, opened 2015-08-13
Session Processing Unification

ctsims:
Updated the way that stack frames are processed so that it runs the same code path on all platforms.

Updated the CLI for usernames and to use the new processing code paths.

### DIFF
--- a/backend/src/org/commcare/util/CommCareSession.java
+++ b/backend/src/org/commcare/util/CommCareSession.java
@@ -613,6 +613,29 @@ public class CommCareSession {
     }
 
     /**
+     * Called after a session has been completed. Executes and pending stack operations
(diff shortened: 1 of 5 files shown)


backend/src/org/commcare/util/CommCareSession.java
```diff
@@ -613,6 +613,29 @@ private void cleanStack() {
...
+
+        //Let the session know that the current frame shouldn't work its way back onto the stack
+        markCurrentFrameForDeath();
+
+        //First, see if we have operations to run
+        if(ops.size() > 0) {
+            executeStackOperations(ops, ec);
+        }
+        return finishAndPop();
+    }
```
phillipm: :+1:

PR #133, opened 2015-08-14
CommCare App Template Test Framework

ctsims:
Added a couple helper and utility classes/methods to make it easy to write tests against small shell applications, so we can write tests about session behavior, and other workflow componenents.

Depends on
https://github.com/dimagi/javarosa/pull/144

### DIFF
--- a/tests/test/org/commcare/cases/test/BadCaseXMLTests.java
+++ b/tests/test/org/commcare/cases/test/BadCaseXMLTests.java
@@ -2,6 +2,8 @@ package org.commcare.cases.test;
 
 import org.commcare.util.mocks.MockDataUtils;
 import org.commcare.util.mocks.MockUserDataSandbox;
+import org.javarosa.xml.util.InvalidStructureException;
+import org.javarosa.xpath.XPathMissingInstanceException;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
@@ -21,13 +23,13 @@ public class BadCaseXMLTests {
         sandbox = MockDataUtils.getStaticStorage();
     }
 
-    @Test
-    public void testNoCaseID() {
-        //Expected - Fail silently (TODO: Fix?)
-        MockDataUtils.parseIntoSandbox(this.getClass().getResourceAsStream("/case_create_broken_no_caseid.xml"), sandbox);
-
-
-        //Make sure that we didn't make a case entry for the bad case though
-        Assert.assertEquals("Case XML with no id should not have created a case record", sandbox.getCaseStorage().getNumRecords(), 0);
+    @Test(expected = InvalidStructureException.class)
+    public void testNoCaseID() throws Exception {
+        try {
+            MockDataUtils.parseIntoSandbox(this.getClass().getResourceAsStream("/case_create_broken_no_caseid.xml"), sandbox, true);
+        }finally {
+            //Make sure that we didn't make a case entry for the bad case though
+            Assert.assertEquals("Case XML with no id should not have created a case record", sandbox.getCaseStorage().getNumRecords(), 0);
+        }
     }
 }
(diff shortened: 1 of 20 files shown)


util/src/org/commcare/util/mocks/MockDataUtils.java
```diff
@@ -41,15 +41,32 @@ public static MockUserDataSandbox getStaticStorage() {
...
+     */
+    public static void parseIntoSandbox(InputStream stream, MockUserDataSandbox sandbox, boolean failfast) throws IOException,
+            UnfullfilledRequirementsException,
+            XmlPullParserException, InvalidStructureException {
+        if(stream == null) {
+            throw new IOException("Parse Stream is Null");
         }
+
+        CommCareTransactionParserFactory factory = new CommCareTransactionParserFactory(sandbox);
+        DataModelPullParser parser = new DataModelPullParser(stream, factory, true, failfast);
```
phillipm: The DataModelPullParser constructor being called has the following declaration:
```DataModelPullParser(InputStream is, TransactionParserFactory factory, boolean failfast, boolean deep)``` https://github.com/dimagi/commcare/blob/master/backend/src/org/commcare/data/xml/DataModelPullParser.java#L51

I'm confused by the fact that the 4th argument being passed into the constructor, called  ```failfast```, is getting named to ```deep``` in the body of the constructor, while the 3rd parameter of the constructor is also named ```failfast```.

Are you sure you didn't mean to name the ```failfast``` variable in the ```parseIntoSandbox``` something else, like ```deep```?

Comment (ctsims) [inline util/src/org/commcare/util/mocks/MockDataUtils.java]: updated

File: tests/test/org/commcare/cases/test/CaseParseAndReadTest.java [hunk 2]
```diff
@@ -41,7 +41,7 @@ public void testReadCaseDB() throws Exception {
 
     private void compareCaseDbState(String inputTransactions,
                                     String caseDbState) throws Exception {
-        MockDataUtils.parseIntoSandbox(this.getClass().getResourceAsStream(inputTransactions), sandbox);
+        MockDataUtils.parseIntoSandbox(this.getClass().getResourceAsStream(inputTransactions), sandbox, true);
```
phillipm: Since ```parseIntoSandbox``` is overridden to set ```failfast``` to true if it isn't provided, you shouldn't need to update existing calls to it, unless you want ```failfast``` set to false.

tests/test/org/commcare/cases/test/BadCaseXMLTests.java
```diff
@@ -21,13 +23,13 @@ public void setUp() {
         sandbox = MockDataUtils.getStaticStorage();
     }
 
-    @Test
-    public void testNoCaseID() {
-        //Expected - Fail silently (TODO: Fix?)
```
phillipm: I can't remember, did this TODO get fixed?

ctsims: It got fixed with failfast being set to true.

phillipm:
This is great! It is really nice to have examples of building up CommCare app instances by hand. Not only does it help with tests and debugging, but also with understanding object dependencies.

ctsims:
Yeah, it's revealed some fairly clear weaknesses and couplings that it's a
bit easier to see how to pull out in this frame.

-Clayton

tests/test/org/commcare/test/utils/MockApp.java
```diff
@@ -0,0 +1,63 @@
...
+    private final SessionWrapper mSessionWrapper;
+
+    /**
+     * Creates and initializes a mockapp that is located at the provided Java Resource path.
+     *
+     * @param resourcePath The resource path to a an app template. Needs to contain a leading and
+     *                     trailing slash, like /path/app/
+     */
+    public MockApp(String resourcePath) throws Exception {
+        if(!resourcePath.startsWith("/") && resourcePath.endsWith("/")) {
```
amstone326: Doesn't this need another set of parentheses so that the ! applies to the whole expression?

ctsims: Good Catch, thanks!

https://github.com/dimagi/commcare/pull/134

PR #137, opened 2015-08-17
Clear localization

amstone326:


wpride:
@amstone326 looks like this test is failing on Jenkins

PR #138, opened 2015-08-17
Tests for navigating through a session

amstone326:
Tests for navigating through a session -- setting commands, datum values, and computed values, and using stepBack()


phillipm:
Nice, definitely helps in understanding sessions!
We're up to 88 tests! 2 weeks ago we had 33

tests/resources/session-tests-template/suite.xml
```diff
@@ -0,0 +1,233 @@
...
+        <text>
+          <xpath function="case_name"/>
+        </text>
+      </template>
+    </field>
+  </detail>
+  <detail id="m1_case_short">
+    <title>
+      <text>
+        <locale id="m1.case_short.title"/>
```
ctsims: These details are either unused or can be cleaned up, right, since these locale strings aren't being loaded? Would be good to remove as much of the extraneous stuff in here as possible to make it easy to extend/adapt.

amstone326: Yep, removed.

tests/resources/session-tests-template/modules-0/forms-0.xml
```diff
@@ -0,0 +1,41 @@
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">
```
ctsims: We don't actually use the data in these forms, right?

Can we replace them all with just hte one template form?

Again, lean/less confusing.

amstone326: @ctsims Yea definitely, sorry for not doing that before. Done now.

PR #140, opened 2015-08-18
Refactor setting computed datum

amstone326:
-move logic for handling and setting a computed datum entirely into CommCareSession, gets passed in the necessary evaluation context
-add hook to SessionWrapper to access the setComputedDatum method, which also passes in the necessary evaluation context

cross: https://github.com/dimagi/commcare-odk/pull/535


ctsims:
this looks good. Can we make sure the new method is used across both Tests/Util (the mock session wrapper) and j2me?

amstone326:
Um yea do you mean like set something up to enforce that now, or what?

ctsims:
Sorry, looks like it's set to be used in the SessionWrapper, and the Android session wrapper, but not the commcare j2me app?

amstone326:
It's actually only set to be used in the SessionWrapper right now. 

I didn't add a hook to AndroidSessionWrapper because it wasn't clear when that would ever be used. There aren't any similar methods in AndroidSessionWrapper at all, like setCommand() or setDatum(), the AndroidSessionWrapper just calls those methods of its session object when it needs to.

And what would it mean to add to j2me?

ctsims:
@amstone326 J2ME code that does the same thing is here:
https://github.com/dimagi/commcare/blob/master/application/src/org/commcare/util/CommCareSessionController.java#L271

It pre-dates the use of DATUM_COMPUTED, so it's checking for a null nodeset instead.

amstone326:
@ctsims Updated with j2me changes

PR #141, opened 2015-08-18
Fix for the parser letting you index your own case

ctsims:


phillipm:
Nice! I vaguely remember this being brought up on a chat or in a FogBugz ticket. Would be nice to link to that here if that is the case.

PR #142, opened 2015-08-20
fix build xml

wpride:


wpride:
this is where the javarosa libs go to now with the gradle build

phillipm:
Was a particular build breaking due to this path being incorrect? Looking at how dir.jrlib is used/set by the main commcare-mobile ant build task, this value shouldn't matter; though I might be misunderstanding the build setup.

wpride:
I think we're actually not using this now, but I'm still using this in a script so this is only helpful to me. we should really just delete this and I'll change scripts I think but i'll figure it out tomw

PR #145, opened 2015-08-20
Detail Screens (Including Tabs) in CLI

ctsims:
Added compound screen mechanism to provide navigation between internal states, and implemented both single (no-tab) and compound (multiple tab) entity detail screens

### DIFF
--- a/util/src/org/commcare/util/cli/ApplicationHost.java
+++ b/util/src/org/commcare/util/cli/ApplicationHost.java
@@ -15,14 +15,11 @@ import org.javarosa.core.services.PropertyManager;
 import org.javarosa.core.services.storage.IStorageIterator;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
 import org.javarosa.engine.XFormPlayer;
-import org.javarosa.xml.util.InvalidStructureException;
-import org.javarosa.xml.util.UnfullfilledRequirementsException;
 import org.javarosa.xpath.XPathException;
 import org.javarosa.xpath.XPathParseTool;
 import org.javarosa.xpath.expr.XPathExpression;
 import org.javarosa.xpath.expr.XPathFuncExpr;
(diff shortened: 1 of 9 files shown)


util/src/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -106,14 +103,22 @@ private void processAppUpdate() {
...
+                }
+
                 System.out.println("\n\n\n\n\n\n");
+                System.out.println(s.getWrappedDisplaytitle(mSandbox, mPlatform));
+
+                System.out.println("====================");
                 s.prompt(System.out);
                 System.out.print("> ");
 
+                screenIsRedrawing = false;
```
orangejenny: this line doesn't look necessary

util/src/org/commcare/util/cli/EntityListSubscreen.java
```diff
@@ -0,0 +1,137 @@
...
+        for (DetailField field : fields) {
+            Object o = field.getTemplate().evaluate(context);
+            String s;
+            if (!(o instanceof String)) {
+                s = "";
+            } else {
+                s = (String)o;
+            }
+
+            int widthHint = SCREEN_WIDTH / fields.length;
```
orangejenny: Non-blocking (and I realize you're just moving this, I just haven't looked at much of the CLI code yet) but I'm surprised you're bothering with template width hints at all, doesn't seem like they'd ever translate well to an ASCII interface.

ctsims: They actually look surprisingly good! we have 100 columns for characters,
which actually might be more than on a lot of phones :p.

orangejenny: oh, that actually makes a ton of sense - the percentages make sense, then.

PR #146, opened 2015-08-21
Fixed issue with session data getting double loaded from the stack

ctsims:
From 
http://manage.dimagi.com/default.asp?178635

If the session has two datums on it that conflict on the definition of a value, the session was getting loaded with both.

### DIFF
--- a/backend/src/org/commcare/util/CommCareSession.java
+++ b/backend/src/org/commcare/util/CommCareSession.java
@@ -405,9 +405,16 @@ public class CommCareSession {
 
         for (StackFrameStep step : frame.getSteps()) {
             if (step.getType() == SessionFrame.STATE_DATUM_VAL) {
-                TreeElement datum = new TreeElement(step.getId());
-                datum.setValue(new UncastData(step.getValue()));
-                sessionData.addChild(datum);
+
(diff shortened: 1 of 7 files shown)


ctsims:
Also adds dbl management integration tests

backend/src/org/commcare/util/CommCareSession.java
```diff
@@ -405,9 +405,16 @@ public FormInstance getSessionInstance(String deviceId, String appversion, Strin
...
         for (StackFrameStep step : frame.getSteps()) {
             if (step.getType() == SessionFrame.STATE_DATUM_VAL) {
-                TreeElement datum = new TreeElement(step.getId());
-                datum.setValue(new UncastData(step.getValue()));
-                sessionData.addChild(datum);
+
+                Vector<TreeElement> matchingElements = sessionData.getChildrenWithName(step.getId());
+
+                if(matchingElements.size() > 0) {
+                    matchingElements.elementAt(0).setValue(new UncastData(step.getValue()));
```
wpride: So we're just defaulting to the older of the two values? Is that always correct behavior?

ctsims: Defaulting to the new value. The attached tests set the value to init and then update it to 'new'

PR #149, opened 2015-08-25
Remove StorageFullException!

phillipm:
@ctsims says ```StorageFullException``` is completely useless, so let's take it out

Cross-requested with:
https://github.com/dimagi/javarosa/pull/149
https://github.com/dimagi/commcare-odk/pull/550

### DIFF
--- a/backend/src/org/commcare/xml/SuiteParser.java
+++ b/backend/src/org/commcare/xml/SuiteParser.java
@@ -11,7 +11,6 @@ import org.commcare.suite.model.Menu;
 import org.commcare.suite.model.Suite;
 import org.javarosa.core.model.instance.FormInstance;
 import org.javarosa.core.services.storage.IStorageUtilityIndexed;
-import org.javarosa.core.services.storage.StorageFullException;
 import org.javarosa.core.services.storage.StorageManager;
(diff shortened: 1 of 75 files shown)


backend/src/org/commcare/xml/SuiteParser.java
```diff
@@ -118,11 +117,6 @@ public Suite parse() throws InvalidStructureException, IOException, XmlPullParse
             // TODO Auto-generated catch block
             e.printStackTrace();
             throw new InvalidStructureException("Pull Parse Exception, malformed XML.", parser);
-        } catch (StorageFullException e) {
-            e.printStackTrace();
-            //BUT not really! This should maybe be added to the high level declaration
-            //instead? Or maybe there should be a more general Resource Management Exception?
-            throw new InvalidStructureException("Problem storing parser suite XML", parser);
```
amstone326: It seems like the StorageFullException was being used here to indicate some other actual problem. We may not want to lose that?

amstone326:
Left a couple of comments on the various parts of this (one here and one in javarosa). Good to merge after those are addressed, unless @ctsims needs to take a look to make sure we're not losing anything we shouldn't be.

PR #150, opened 2015-08-25
Subentity detail

orangejenny:
Parsing and model changes to support https://github.com/dimagi/commcare-odk/pull/554

Did a semi-implementation in the CLI that piggybacks heavily on the entity select screen. This was for debugging purposes, no guarantee it's worth anything.


ctsims:
If we are on J2ME right now what would happen if we tried to run this? Would be good for it to fail in some expected way.

util/src/org/commcare/util/cli/EntityScreen.java
```diff
@@ -113,10 +114,17 @@ public boolean setCurrentScreenToDetail() {
...
 
-    public void setCurrentScreenToDetail(int index) {
+    public void setCurrentScreenToDetail(int index) throws CommCareSessionException {
         EvaluationContext subContext = new EvaluationContext(mSession.getEvaluationContext(), this.mCurrentSelection);
 
-        this.mCurrentScreen = new EntityDetailSubscreen(index, this.mLongDetailList[index], subContext, getDetailListTitles(subContext));
+        TreeReference detailNodeset = this.mLongDetailList[index].getNodeset();
+        if (detailNodeset != null) {
+            TreeReference contextualizedNodeset = detailNodeset.contextualize(this.mCurrentSelection);
+            this.mCurrentScreen = new EntityListSubscreen(this.mLongDetailList[index], subContext.expandReference(contextualizedNodeset), subContext);
```
ctsims: This screws up the terminal outputs from this, right? IE: after choosing the list detail things go badly, and it shows an odd list of items that have 11) before and such?

Not sure how insistent to be on making the CLI functional for new features, but depending on LOE can you let me know whether you're planning on making it work more similarly to the existing details?

ctsims:
Oh, one other thing: We won't have this regressable on HQ for a while, so would be super helpful to have at least a walkthrough in a MockApp harness for loading and making sure everything is working.

orangejenny:
J2ME: I suspect it'll fail with a bad XPath expression, since the nodeset never gets applied to the detail's main node. Thoughts on how robust it should be? I'm guessing it wouldn't be terrible to implement on J2ME (beyond the basic terrible nature of J2ME).

util/src/org/commcare/util/cli/EntityScreen.java
orangejenny: Yeah, wanting this to actually work in the CLI is reasonable. Don't want it to block this PR, but happy to make myself a ticket and do it once I'm through my near-term TDH tasks.

ctsims:
Hey,

I made some changes to striping that should be fairly easy to pick up, but will need to be adapted here (it makes the White/Grey v. Blue/Blue striping work).

orangejenny:
@ctsims Updated so that if you try to access case details for a detail with a nodeset, J2ME will throw a RuntimeException saying `Entity subnodes not supported: /your/nodeset`  Does that work?

PR #152, opened 2015-08-25
Case parent queries should fail silently

phillipm:
Working on fixing case queries that include parent http://manage.dimagi.com/default.asp?179998

Bug was introduced by this PR https://github.com/dimagi/commcare/pull/127


phillipm:
Moving this to https://github.com/dimagi/commcare/pull/155

PR #155, opened 2015-08-26
Use a hardcoded template for casedb instances

phillipm:
Create CaseDataInstance, which extends ExternalDataInstance, but provides a custom hasTemplatePath implementation that respects the static guarantee's that the [casedb](https://github.com/dimagi/commcare/wiki/casedb) spec gives us regarding path structure.

Fixes http://manage.dimagi.com/default.asp?179998

Bug was introduced by this PR https://github.com/dimagi/commcare/pull/127, which has been reverted in master (https://github.com/dimagi/commcare/pull/151). This PR re-implements the changes.

__I didn't test the changes I made to files in ```application/```, since those are used by J2ME phones, which I don't have access to.__

Cross-requested with:
https://github.com/dimagi/commcare-odk/pull/560
https://github.com/dimagi/javarosa/pull/155


cases/resources/casedb_instance_structure.xml
```diff
@@ -0,0 +1,28 @@
+<instance>
```
phillipm: I need a top level node that isn't ```casedb```. Is ```instance``` a good choice for a name or is there something better?

application/build.xml
```diff
@@ -129,6 +129,7 @@
 
                 <!-- Resources that should be used in the Polish build (images, the polish.css file, etc) -->
                 <resources defaultexcludes="no" excludes="readme.txt,.project">
+                    <root dir="${dir.commcare.cases.resources}" />
```
phillipm: Does this look like the right place to ensure that [cases/resources/casedb_instance_structure.xml](https://github.com/dimagi/commcare/blob/bugs/case-data-instance/cases/resources/casedb_instance_structure.xml) makes it into the j2me build?

cases/src/org/commcare/cases/instance/CaseDataInstance.java
```diff
@@ -0,0 +1,84 @@
...
+     * Does the reference follow the statically defined CaseDB spec?
+     */
+    public boolean hasTemplatePath(TreeReference ref) {
+        if (caseDbSpecTemplate != null) {
+            return followsTemplateSpec(ref, caseDbSpecTemplate, 0);
+        } else {
+            // failed to load casedb spec template, default to super implementation
+            Logger.log("CaseDb Warning",
+                    "Using default hasTemplatePath implementation: reference resolution will break!");
+            return super.hasTemplatePath(ref);
```
phillipm: This code path should never be reached. Do you think it is smarter to throw a runtime exception here or default to the super's implementation?

application/build.xml
ctsims: Hm, I believe so, yeah. Did we test on a j2me?

cases/resources/casedb_instance_structure.xml
ctsims: I would call it "wrapper", honestly, just because I think that's the clearest way to illustrate that it's not there for a functional purpose

cases/src/org/commcare/cases/instance/CaseDataInstance.java
ctsims: Honestly I'm thinking whether it would make more sense to literally inline the template text as static content here, although that would make the permgen size of this file really large. 

It makes me really uncomfortable that the code path for loading this is also static, as it introduces _huge_ unpredictable hits across the codebase for when a complex parse happens (And will/can mean it happens at weird and bad times), can we make the static init here happen lazily instead? (Is fine for the template parse itself to be static, I think), as long as it is stateless and threadsafe)

phillipm: I agree that the use of static instances should be limited. Do you recommend a non-static pattern to use?

That said, I don't understand what is unpredictable about this. Static initialization seems to have defined semantics (https://stackoverflow.com/questions/3499214/java-static-class-initialization) and in this case will occur when we register the CaseDataInstance using ```registerExternalDataInstanceBuilder``` (unless we begin to make use of CaseDataInstance definitions earlier).

Static initialization seems to be thread safe by default (http://stackoverflow.com/a/462733)

ctsims: We will end up touching this class (loading into permgen space) when/if we do any reflected walks through the code (like we do in CCODK when we build the prototype factory.

Is the inherent advantage here (v. lazy init) that it occurs in a threadsafe way? I think in general it's ~always safer to lazy init this kind of thing whenever possible, since the semantics are much clearer and if there's a failure it happens explicitly when you requested an operation that used the class.

cases/src/org/commcare/cases/ledger/instance/LedgerInstanceTreeElement.java
```diff
@@ -94,10 +94,7 @@ public void attachStringCache(Interner<String> stringCache) {
     public LedgerChildElement getChild(String name, int multiplicity) {
         if ((multiplicity == TreeReference.INDEX_TEMPLATE) &&
                 "ledger".equals(name)) {
-            return null;
-            // TODO PLM: commenting out the following line until http://manage.dimagi.com/default.asp?179998 is fixed
-            // this will break the ledger query silent fail fix for http://manage.dimagi.com/default.asp?177247
-            // return LedgerChildElement.TemplateElement(this);
+            return LedgerChildElement.TemplateElement(this);
```
ctsims: why did we not need to make one of these for ledgers as well as cases?

util/src/org/commcare/util/mocks/CommCareInstanceInitializer.java
```diff
@@ -13,7 +15,7 @@
 /**
  * @author ctsims
  */
-public class CommCareInstanceInitializer extends InstanceInitializationFactory {
+public class CommCareInstanceInitializer implements InstanceInitializationFactory {
```
ctsims: Hm, do we know where the J2ME one of these lives? It should also get changed, right?

cases/src/org/commcare/cases/ledger/instance/LedgerInstanceTreeElement.java
phillipm: This is an artifact of the spec's structure for ledgers. Since everything after "ledger" is a wildcard, we don't run into the issue where traversing the template passes over TreeElements. This was the behavior during template checking of Ledger/CaseChildElement.

phillipm: I could change it to follow a similar pattern, but things work currently, so I figured I wouldn't bother

util/src/org/commcare/util/mocks/CommCareInstanceInitializer.java
phillipm: Good call, forgot to check that! I've made InstanceIntializerFactory a class again with default behavior

cases/src/org/commcare/cases/instance/CaseDataInstance.java
phillipm: okay, I think I've implemented lazy loading. Please check that it is what you had in mind.

application/build.xml
phillipm: No, should I do that or wait for QA?

cases/src/org/commcare/cases/instance/CaseDataInstance.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private static void loadTemplateSpecLazily() {
+        if (caseDbSpecTemplate == null) {
+            try {
+                caseDbSpecTemplate =
+                        FormLoadingUtils.xmlToTreeElement("/casedb_instance_structure.xml");
+            } catch (InvalidStructureException e) {
+                throw new RuntimeException("Failed to load casedb template spec");
+            } catch (IOException e) {
+                throw new RuntimeException("Failed to load casedb template spec");
```
phillipm: I think the correct thing to do if the template fails to load is error out.

cases/src/org/commcare/cases/instance/CaseDataInstance.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public CaseDataInstance(ExternalDataInstance instance) {
+        super(instance);
+    }
+
+    /**
+     * Does the reference follow the statically defined CaseDB spec?
+     */
+    public boolean hasTemplatePath(TreeReference ref) {
+        synchronized (lock) {
```
phillipm: Is it overkill to synchronize the whole template path check method? We need to synchronize both reads & writes, not just writes, correct?

application/build.xml
ctsims: I'd do if after we pull at the earliest

cases/src/org/commcare/cases/instance/CaseDataInstance.java
ctsims: you only need to lock around the load method, that'll be enough to block anything you need.

cases/src/org/commcare/cases/instance/CaseDataInstance.java
ctsims: Definitely agree. I might make this error very, very explanatory since I'm not sure what people would expect to happen here.

cases/src/org/commcare/cases/ledger/instance/LedgerInstanceTreeElement.java
ctsims: cool, agreed

PR #161, opened 2015-09-04
J2ME tabs

orangejenny:
http://manage.dimagi.com/default.asp?154957

Make tabs "work" on J2ME: just display all fields from all tabs on one screen.

So a case detail configuration like this:
screen shot 2015-09-04 at 2 19 11 pm

Will look like this:
img_20150904_131650


orangejenny:
Given that this has never worked, getting it into 2.23 isn't a priority, so I'm just comparing with master.

application/src/org/commcare/entity/CommCareEntity.java
```diff
@@ -165,8 +186,42 @@ protected void loadEntity(TreeReference element) {
...
+        for (int i = 0; i < details.length; i++) {
+            count += details[i].getFields().length;
+        }
+        return count;
+    }
+
     private void loadVars(EvaluationContext ec, Detail detail) {
-        Hashtable<String, XPathExpression> decs = detail.getVariableDeclarations();
+        Detail[] details = getFlattenedDetails(detail);
+        Hashtable<String, XPathExpression> decs = new Hashtable<String, XPathExpression>();
```
wpride: better name for "decs" would be nice

wpride:
The 'j' 'k' 'i' iteration stuff is messy but was already there/unavoidable due to no decent iterator methods in j2me 

I'm gtm after annoying nitpick addressed

application/src/org/commcare/entity/CommCareEntity.java
```diff
@@ -165,8 +186,42 @@ protected void loadEntity(TreeReference element) {
         loadTexts(ec);
     }
 
+    /**
+     * Given a detail, return an array of details that will contain either
+     * - all child details
+     * - a single-element array containing the given detail, if it has no children
+     * @param d The detail to flatten.
+     */
+    private Detail[] getFlattenedDetails(Detail d) {
```
ctsims: Can we move this to the detail class or somewhere that will make it possible for us to unify the behavior on platforms which don't support tabs?

orangejenny: ha, I had the exact opposite thought, "This logic should live in J2ME because surely any other platform would support tabs." will update.

orangejenny:
feedback addressed, should be good to go

PR #162, opened 2015-09-10
throw storage full exception properly

wpride:


wpride:
cross https://github.com/dimagi/commcare-odk/pull/620

backend/src/org/commcare/data/xml/DataModelPullParser.java
```diff
@@ -62,7 +63,7 @@ public DataModelPullParser(InputStream is, TransactionParserFactory factory, boo
         this.rListener = rListener;
     }
 
-    public Boolean parse() throws InvalidStructureException, IOException, XmlPullParserException, UnfullfilledRequirementsException {
+    public Boolean parse() throws InvalidStructureException, IOException, XmlPullParserException, UnfullfilledRequirementsException, StorageFullException {
```
ctsims: As a runtime exception we don't need to add this signature.

backend/src/org/commcare/data/xml/DataModelPullParser.java
```diff
@@ -135,6 +136,9 @@ private void parseBlock(String root) throws InvalidStructureException, IOExcepti
                 if (!failfast) {
                     try {
                         transaction.parse();
+                    } catch(StorageFullException e){
```
ctsims: By "Try forever" you just mean that it fails silently, right?

phillipm:
@wpride can this be closed now that you've updated https://github.com/dimagi/commcare-odk/pull/620 ?

PR #163, opened 2015-09-10
Add EntitySubnodeDetailSubscreen

orangejenny:
Opening for feedback.

@ctsims This no longer crashes, but after you confirm on the detail screen, it brings up an empty menu screen instead of going back home - the command should be null/root, but it's still m0. I'm having a pretty difficult time debugging the session changes, am curious if you see anything obviously wrong here. If not, I'll dig more into sessions.


orangejenny:
Debugging this is low priority, given that ICDS is the only project using this feature. Closing for the time being.

PR #164, opened 2015-09-21
Case autoselect

amstone326:
-Moved session navigation logic that used to be executed in CommCareHomeActivity out into its own Android-independent class
-Added case list auto-select logic: If the autoselect flag is enabled for a module and there is exactly 1 case in the nodeset, that case will be selected for the user automatically, rather than showing an entity select screen with only 1 item in it. CommCare now automatically proceeds either to the detail confirm screen for that case, if it is defined, or to form entry if it is not.
-The new test file contains tests for both for the session navigation refactor and the case list auto-select functionality

cross: https://github.com/dimagi/commcare-odk/pull/642


backend/src/org/commcare/suite/model/SessionDatum.java
```diff
@@ -172,4 +174,19 @@ public TreeReference getEntityFromID(EvaluationContext ec, String elementId) {
             return null;
         }
     }
+
+    public static String getCaseIdFromReference(TreeReference contextRef,
```
amstone326: Not sure if this is the best place for this method -- I pulled this functionality out of a larger method in EntitySelectActivity and made it static, and needed to move it into commcare so it would be accessible to all of the session navigation code I moved into here. But if someone doesn't like it in this file / has a better idea, let me know.

amstone326:
Release Note: If the autoselect flag is enabled for a module and there is exactly 1 case in the nodeset, that case will be selected for the user automatically, rather than showing an entity select screen with only 1 item in it.

backend/src/org/commcare/util/SessionNavigationResponder.java
```diff
@@ -0,0 +1,22 @@
+package org.commcare.util;
```
phillipm: `util` is sort of a strange place to put this. Is there a session related folder that this, and related files, can be put in? If not, let's create one.

amstone326: Agreed, but I put these files here because this is where CommCareSession.java is also. Happy to move them all into a session-related package.

tests/test/org/commcare/test/utilities/MockSessionNavigationResponder.java
```diff
@@ -0,0 +1,41 @@
...
+import org.javarosa.core.model.condition.EvaluationContext;
+
+/**
+ * A mock implementer of the SessionNavigationResponder interface, for testing purposes
+ *
+ * @author amstone
+ */
+public class MockSessionNavigationResponder implements SessionNavigationResponder {
+
+    private SessionWrapper sessionWrapper;
```
phillipm: can be final

tests/test/org/commcare/backend/util/test/SessionNavigatorTests.java
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * Tests for SessionNavigator.java
+ *
+ * @author amstone
+ */
+public class SessionNavigatorTests {
+
+    MockApp mApp;
+    MockSessionNavigationResponder mSessionNavigationResponder;
+    SessionNavigator sessionNavigator;
```
phillipm: These can be made private

tests/test/org/commcare/backend/util/test/SessionNavigatorTests.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Test
+    public void testNavWithoutAutoSelect() {
+        SessionWrapper session = mApp.getSession();
+
+        // Before anything has been done in the session, the sessionNavigatorResponder should be
+        // directed to get a command
+        sessionNavigator.startNextSessionStep();
+        Assert.assertEquals(SessionNavigator.GET_COMMAND,
+                mSessionNavigationResponder.getLastResultCode());
```
phillipm: This code block is repeated several times, would be good to lift into a helper or put once in the `setUp` method

phillipm:
Nice tests and refactor. Ready to merge after fixing the minor lint warnings & code duplication that I pointed out.

tests/test/org/commcare/backend/util/test/SessionNavigatorTests.java
amstone326: good call

amstone326:
@phillipm Thanks! Addressed your comments. @ctsims, do you want to do a pass before it's merged?

PR #169, opened 2015-09-30
Make session stack robust to datums added out of order

amstone326:
Fix for http://manage.dimagi.com/default.asp?183023


tests/test/org/commcare/backend/session/test/SessionStackTests.java
```diff
@@ -20,24 +20,21 @@
 public class SessionStackTests {
     MockApp mApp;
 
-    @Before
-    public void init() throws Exception{
-        mApp = new MockApp("/complex_stack/");
-    }
-
     @Test
```
amstone326: All of the pre-existing tests in this file had the 'expected' and 'actual' arguments to assertEquals() swapped, so fixed that.

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -248,15 +253,30 @@ public SessionDatum getNeededDatum() {
      * @return A session datum definition if one is pending. Null otherwise.
      */
     public SessionDatum getNeededDatum(Entry entry) {
-        int nextVal = getData().size();
-        //If we've already retrieved all data needed, return null.
-        if (nextVal >= entry.getSessionDataReqs().size()) {
+        Vector<SessionDatum> allDatumsNeeded = entry.getSessionDataReqs();
+        OrderedHashtable datumsSoFar = getCollectedDatums();
+
+        if (datumsSoFar.size() >= allDatumsNeeded.size()) {
```
ctsims: This actually isn't a sufficient check, because the frame can contain additional data that is irrelevant (is kind of a long story).

either way, would prefer to stick with doing the full check.

phillipm:
:+1: Looks good!

tests/resources/session-tests-template/suite.xml
```diff
@@ -70,6 +70,21 @@
       <datum id="case_id" nodeset="instance('casedb')/casedb/case[@case_type='pregnancy'][@status='open']" value="./@case_id" detail-select="m0_case_short" detail-confirm="m0_case_long" autoselect="true"/>
     </session>
   </entry>
+    <entry>
+        <form>http://openrosa.org/formdesigner/1A0EDB0F-D139-47F2-B06C-FF1203DC5AC4</form>
```
ctsims: This isn't actually pointing at the placeholder, as a heads up

tests/resources/session-tests-template/suite.xml
```diff
@@ -70,6 +70,21 @@
...
+        <command id="m0-f3">
+            <text>
+                <locale id="forms.m0f3"/>
```
ctsims: Can we replace this with a non-locale'd text? (we don't use locale stuff in mockapps generally, for clarify)

tests/test/org/commcare/backend/session/test/SessionStackTests.java
```diff
@@ -20,24 +20,21 @@
...
+        mApp = new MockApp("/complex_stack/");
         SessionWrapper session = mApp.getSession();
-        Assert.assertEquals(session.getNeededData(), SessionFrame.STATE_COMMAND_ID);
+
+        Assert.assertEquals(SessionFrame.STATE_COMMAND_ID, session.getNeededData());
 
         session.setCommand("m0");
 
-        Assert.assertEquals(session.getNeededData(), SessionFrame.STATE_DATUM_COMPUTED);
+        Assert.assertEquals(SessionFrame.STATE_DATUM_COMPUTED, session.getNeededData());
```
ctsims: good catch on this, thanks!

tests/test/org/commcare/backend/session/test/SessionStackTests.java
```diff
@@ -85,21 +84,47 @@ public void testViewNav() throws Exception {
...
+
+        // Set 2 of the 3 needed datums, but not in order (1st and 3rd)
+        session.setDatum("case_id", "case_id_value");
+        session.setDatum("usercase_id", "usercase_id_value");
+
+        // Session should now need the case_id_new_visit_0, which is a computed datum
+        Assert.assertEquals(SessionFrame.STATE_DATUM_COMPUTED, session.getNeededData());
+
+        // The key of the needed datum should be "case_id_new_visit_0"
+        Assert.assertEquals("case_id_new_visit_0", session.getNeededDatum().getDataId());
```
ctsims: I think we should add this datum to the frame and confirm that the session also successfully moves on (just to ensure the end-to-end is also still working)

PR #171, opened 2015-09-30
Session stack fix part 2

amstone326:
Fixed getNeededData() to do a proper check for what the next datum needed by an entry is -- makes the stack robust to more complicated versions of the datum entries on the stack being out of order, and to having bogus data on the stack.

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -72,7 +72,7 @@ public class CommCareSession {
         this.platform = platform;
         collectedDatums = new OrderedHashtable();
         this.frame = new SessionFrame();
-        this.frameStack = new Stack<>();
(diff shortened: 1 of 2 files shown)


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -72,13 +72,19 @@ public CommCareSession(CommCarePlatform platform) {
         this.platform = platform;
         collectedDatums = new OrderedHashtable();
         this.frame = new SessionFrame();
-        this.frameStack = new Stack<SessionFrame>();
+        this.frameStack = new Stack<>();
```
ctsims: this syntax doesn't work on j2me :/

amstone326: Ugh sorry I'm the worst ever at remembering that

PR #177, opened 2015-10-15
Enabled suite files to request that menus use the grid view layout

ctsims:


backend/src/org/commcare/util/CommCarePlatform.java
```diff
@@ -437,4 +438,22 @@ public String getModuleNameForEntry(Entry entry) {
...
     }
+
+    public String getMenuDisplayStyle(String menuId) {
+        Vector<Suite> installed = getInstalledSuites();
+        String commonDisplayStyle = null;
+        for(Suite s : installed) {
+            for(Menu m : s.getMenus()) {
+                if(menuId.equals(m.getId())) {
+                    if(m.getStyle() != null) {
+                        if(commonDisplayStyle != null && !m.getStyle().equals(commonDisplayStyle)){
```
phillipm: Can you explain why this conditional is needed?

tests/test/org/commcare/backend/suite/model/test/AppStructureTests.java
```diff
@@ -0,0 +1,46 @@
...
+                "grid",
+                mApp.getSession().getPlatform().getMenuDisplayStyle("root"));
+
+        Assert.assertEquals("Common Menu Style",
+                "list",
+                mApp.getSession().getPlatform().getMenuDisplayStyle("m1"));
+
+        Assert.assertEquals("Disperate Menu Style",
+                null,
+                mApp.getSession().getPlatform().getMenuDisplayStyle("m2"));
```
phillipm: Can you write a test that tries to look-up the style for `m0`, which doesn't have a style?

backend/src/org/commcare/util/CommCarePlatform.java
ctsims: This is what is covered in the mixed-menu styles on the test below. In commcare you can have more than one <menu> element for an ID to enable suite files to inject things into menus as extensions. In the event that two (or more) menus for an ID define different @style attributes for the same menu Id, it'll fall back to using neither.

tests/test/org/commcare/backend/suite/model/test/AppStructureTests.java
ctsims: Done. Also added a test to clarify the ambiguous case where one menu defines a style and the other doesn't

PR #178, opened 2015-10-20
Update our Extension Case strategy to match the new concept

ctsims:
New notion is documented here:
https://gist.github.com/dannyroberts/f184daad468fb7debf10

Adds and removes a bit of functionality from extensions that makes them more clear and meets our programatic needs better.


cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -57,14 +61,13 @@ public CasePurgeFilter(IStorageUtilityIndexed<Case> caseStorage) {
...
 
-    private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<String> owners) {
-        //Create a DAG. The Index will be the case GUID. The Nodes will be a string array containing
-        //[CASE_STATUS, string(storageid)]
-        //CASE_STATUS is enumerated as one of STATUS_LIVE, STATUS_DEAD, or STATUS_ABANDONED
-        DAG<String, String[], String> g = new DAG<String, String[], String>();
+    private void setIdsToRemoveWithNewExtensions(IStorageUtilityIndexed<Case> caseStorage, Vector<String> owners) {
+        //Create a DAG. The Index will be the case GUID. The Nodes will be a int array containing
+        //[STATUS_FLAGS, storageid]
+        DAG<String, int[], String> g = new DAG<String, int[], String>();
```
phillipm: `g` is a bad variable name. What is stored in the edge data of the graph?

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -102,29 +105,20 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
+                nodeStatus |= STATUS_OWNED;
             }
 
+            if(!c.isClosed()) {
+                nodeStatus |= STATUS_OPEN;
+            }
+
+            if(owned && !c.isClosed()) {
+                nodeStatus |= STATUS_RELEVANT;
+            }
```
phillipm: perfect place for a `private static int calculateCaseStatus(boolean isOwned, boolean isOpen)` helper method

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
-        for (Enumeration iterator = g.getNodes(); iterator.hasMoreElements(); ) {
-            String[] node = (String[])iterator.nextElement();
-            if (!node[0].equals(STATUS_LIVE)) {
-                idsToRemove.addElement(Integer.valueOf(node[1]));
+    private boolean is(int[] node, int flag) {
+        return (node[0] & flag) == flag;
+    }
+
+
+    private boolean hasExtension(Vector<CaseIndex> indexHolder) {
```
phillipm: Unused

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
         }
+    }
 
-        //Ok, so now just go through all nodes and signal that we need to remove anything
-        //that isn't live!
-        for (Enumeration iterator = g.getNodes(); iterator.hasMoreElements(); ) {
-            String[] node = (String[])iterator.nextElement();
-            if (!node[0].equals(STATUS_LIVE)) {
-                idsToRemove.addElement(Integer.valueOf(node[1]));
+    private boolean is(int[] node, int flag) {
```
phillipm: consider a more descriptive name: `caseStatusIs`, `nodeStatusHasFlag`, `statusHasMask`

phillipm: consider passing in the case status instead of the int array whose first element is the case's status

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
             //current node
             String index = toProcess.pop();
-            String[] node = g.getNode(index);
-
-            //Walk all indexing nodes.
-            for (Edge<String, String> edge : g.getParents(index)) {
-                if (g.getNode(edge.i)[0].equals(STATUS_ABANDONED) &&
-                        edge.e.equals(CaseIndex.RELATIONSHIP_EXTENSION) &&
-                        node[0].equals(STATUS_LIVE)) {
+            int[] node = g.getNode(index);
```
phillipm: Since we don't care about the case ID stored in the node, I would change this to `int caseStatus = g.getNode(index)[0];`

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
+    }
 
-        //Pass 2: Start at Sinks (nodes which only receive edges) and walk down the
-        //graphs, marking necessary nodes as alive
-        toProcess = g.getSinks();
+    private void walk(DAG<String, int[], String> g, boolean direction, int mask, int mark) {
+        walk(g, direction, mask, mark, null);
+    }
+
+    private void walk(DAG<String, int[], String> g, boolean direction, int mask, int mark, String relationship) {
```
phillipm: Consider renaming to a more descriptive name, something like `private void applyMaskToDAG(DAG<String, int[], String> dag, boolean startAtSources, int predicateMask, int maskToApply,  String caseRelationship)`, and adding method level comments.

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
-
-            //Walk all indexing nodes.
-            for (Edge<String, String> edge : g.getParents(index)) {
-                if (g.getNode(edge.i)[0].equals(STATUS_ABANDONED) &&
-                        edge.e.equals(CaseIndex.RELATIONSHIP_EXTENSION) &&
-                        node[0].equals(STATUS_LIVE)) {
+            int[] node = g.getNode(index);
 
-                    g.getNode(edge.i)[0] = STATUS_LIVE;
+            for (Edge<String, String> edge : (direction ? g.getChildren(index) : g.getParents(index))) {
```
phillipm: For readability it would be good to unpack ternary operation to outside of the loop

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
+                if (is(node, mask) && (relationship == null || edge.e.equals(relationship))) {
+                    g.getNode(edge.i)[0] |= mark;
```
phillipm: Tricky tricky array entry dereferencing... I would really prefer to see a solution that doesn't store heterogeneous data in arrays.

Like `CasePurgeDag extends Dag<String, CasePurgeData, String> ...`
and 
```class CasePurgeData { 
private int caseStatus;
private final int storageId;
public boolean hasFlag();
public void addStatusFlag();
public int getStatus();
...
}
```

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
+            if(is(node, STATUS_OWNED | STATUS_RELEVANT | STATUS_AVAILABLE)) {
+                node[0] |= STATUS_ALIVE;
+            }
+        }
+
+        walk(g, true, STATUS_ALIVE, STATUS_ALIVE);
+        walk(g, false, STATUS_ALIVE, STATUS_ALIVE, CaseIndex.RELATIONSHIP_EXTENSION);
+    }
+
+    private boolean hasOutgoingExtension(DAG<String, int[], String> g, String index) {
```
phillipm: Move this to directly below where it is used

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
         }
 
-        //Pass 2: Start at Sources (nodes which only produce edges) and walk up
-        //the tree, marking relevant nodes as alive
-        Stack<String> toProcess = g.getSources();
-        while (!toProcess.isEmpty()) {
-            //current node
-            String index = toProcess.pop();
-            String[] node = g.getNode(index);
+        markRelevant(g);
```
phillipm: I would rename this to `propagateRelevancy`

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
-                    g.getNode(edge.i)[0] = STATUS_LIVE;
+            for (Edge<String, String> edge : (direction ? g.getChildren(index) : g.getParents(index))) {
+                if (is(node, mask) && (relationship == null || edge.e.equals(relationship))) {
+                    g.getNode(edge.i)[0] |= mark;
                 }
-
-                //add this to the process stack
-                //TODO: We're going to walk stuff redundantly this
-                //way. Can we optimize?
                 toProcess.addElement(edge.i);
```
phillipm: It doesn't seem like this DAG traversal algorithm has any measures that prevent the same node from being processed more than once. Won't a graph like the following hit `C` twice?
```
A -> B, C
B -> C,D
C -> D
```

tests/test/org/commcare/cases/util/test/CasePurgeFilterTests.java
```diff
@@ -60,73 +70,9 @@ public void setUp() throws Exception {
...
-        try {
-            storage.write(a);
-            storage.write(b);
-            storage.write(c);
-            storage.write(d);
-
-            int[] present = new int[] {a.getID(), b.getID(), d.getID()};
-            int[] toRemove = new int[] { c.getID()};
+    @RunWithResource("/case_relationship_tests.json")
+    public void runTests() {
```
phillipm: This method should be renamed to delineate the relationship with `CasePurgeTest`

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -132,61 +126,91 @@ private void setIdsToRemoveNew(IStorageUtilityIndexed<Case> caseStorage, Vector<
...
-            //current node
-            String index = toProcess.pop();
-            String[] node = g.getNode(index);
+        markRelevant(g);
+        markAvailable(g);
+        markLive(g);
+    
+        //Ok, so now just go through all nodes and signal that we need to remove anything
+        //that isn't live!
+        for (Enumeration<int[]> iterator = g.getNodes(); iterator.hasMoreElements(); ) {
```
phillipm: This and all other instances of `for (Enumeration<SomeType> ...` will cause J2ME builds to fail, right?

phillipm:
Went a little overboard with this review. Would really appreciate if you addressed the code readability comments, but would definitely look into the DAG-traversal related comments.

cases/src/org/commcare/cases/util/CasePurgeFilter.java
ctsims: We need to retain the node[] array itself for mutation below, so we have to have the actual object itself

cases/src/org/commcare/cases/util/CasePurgeFilter.java
ctsims: It doesn't, really. In practice I've avoided prematurely optimizing because in large-N situations the reality is that fetching all of the case data from storage is the bottleneck by a pretty large margin

cases/src/org/commcare/cases/util/CasePurgeFilter.java
ctsims: Hm, we'll play it your way, I think, although I'm not sure I agree with this in general. This refactor turns

is(node, STATUS_ALIVE)

into

caseStatusIs(node.getStatus(), STATUS_ALIVE), which I don't think is more readable. 

I suppose a further refactor would just be 

node.statusMatches(STATUS_ALIVE) which would be good

ctsims:
@phillipm Addressed all of the code readability points. I didn't refactor out the independent datatypes for now (CaseRelationshipGraph, etc) but clarified and unpacked where possible

cases/src/org/commcare/cases/util/CasePurgeFilter.java
phillipm: I guess I was concerned about the behavior being incorrect, but in this case we are alright because the operation applied to the nodes is idempotent

PR #181, opened 2015-10-26
Override core behavior to return CaseDataInstance

wpride:


wpride:
Possible fix for http://manage.dimagi.com/default.asp?183837

Basically the initialization factory was not overridden to use CaseDataInstance object in J2ME meaning that we never had access to the template when we tried to walk to parent reference.

Don't pull til I can test tomorrow

application/src/org/commcare/util/CommCareInstanceInitializer.java
```diff
@@ -45,6 +46,14 @@ public CommCareInstanceInitializer(Interner<String> stringCache, CommCareSession
         this.session = session;
     }
```
ctsims: I'd add an @override here. We can use those in /commcare now.

phillipm:
Did someone confirm this fixed http://manage.dimagi.com/default.asp?183837 ?

PR #191, opened 2015-11-09
Restore last case list query

phillipm:
cross-request: https://github.com/dimagi/commcare-odk/pull/795


tests/test/org/commcare/backend/session/test/SessionNavigatorTests.java
```diff
@@ -144,5 +144,39 @@ public void testAutoSelectEnabledWithOneCaseAndNoDetail() {
         triggerSessionStepAndCheckResultCode(SessionNavigator.START_FORM_ENTRY);
     }
 
+    @Test
+    public void testSettingAndGettingSessionExtras() {
```
phillipm: This is a pretty standard session traversal. Let me know if there are corner cases that I should add to the test.

PR #193, opened 2015-11-10
Merge HQ case purge tests

proteusvacuum:
- Added a few tests from HQ
- snake_case the names so they are greppable when tests fail (HQ uses the name for the name of the test run by the test runner)
- Pretty print the JSON


proteusvacuum:
@ctsims

PR #195, opened 2015-11-12
Enable restoration of serialized commcare session

phillipm:
cross-requested with https://github.com/dimagi/commcare-odk/pull/811

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -16,12 +16,16 @@ import org.javarosa.core.model.instance.InstanceInitializationFactory;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.services.locale.Localizer;
 import org.javarosa.core.util.OrderedHashtable;
+import org.javarosa.core.util.externalizable.DeserializationException;
+import org.javarosa.core.util.externalizable.ExtUtil;
 import org.javarosa.xpath.XPathException;
 import org.javarosa.xpath.XPathParseTool;
 import org.javarosa.xpath.expr.XPathExpression;
 import org.javarosa.xpath.expr.XPathFuncExpr;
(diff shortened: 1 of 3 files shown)


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -52,7 +54,7 @@
     /**
      * A table of all datums (id --> value) that are currently on the session stack
      */
-    private final OrderedHashtable collectedDatums;
+    private final OrderedHashtable<String, String> collectedDatums;
```
amstone326: Probably not important but what is this change for?

backend/src/org/commcare/session/SessionFrame.java
```diff
@@ -95,12 +98,10 @@ public String getFrameId() {
      * This snapshot can be referenced later to compare the eventual state
      * of the frame to an earlier point
      */
-    public void captureSnapshot() {
-        synchronized (steps) {
```
amstone326: Not sure what this lock was doing but why was it necessary to remove it / is it definitely ok to do so?

backend/src/org/commcare/session/CommCareSession.java
phillipm: Adding type annotations like this help prevent runtime errors by doing type checking during compile time.
And they prevent annoying lint warnings :)

backend/src/org/commcare/session/SessionFrame.java
phillipm: I didn't really remove it, I moved it to the method signature.
The reason I did that is because all the usages to `synchronized (steps)` wrapped the entire method, which means it is equivalent to making the whole method `synchronized`

phillipm: when you synchronize at the method level it uses a "hidden" class instance lock.

PR #196, opened 2015-11-19
Save custom data to session

wpride:


wpride:
cross https://github.com/dimagi/commcare-odk/pull/834

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -65,17 +66,24 @@
...
         this.frame = new SessionFrame();
         this.frameStack = new Stack<SessionFrame>();
+        this.customSessionData = new Hashtable<>();
     }
 
     public Vector<Entry> getEntriesForCommand(String commandId) {
         return this.getEntriesForCommand(commandId, new OrderedHashtable());
     }
 
+    public void putSessionDatum(String key, String value){
```
amstone326: Call this anything but 'datum', there are already too many datas and datums in a session. Try like putCustomSessionValue() or something like that.

phillipm:
Curious, would be possible to make use of the infrastructure from https://github.com/dimagi/commcare/pull/191 that enables storing extra data in the session's frames?

ctsims:
@wpride Can we talk about the intent and purpose of this? I don't think I'm aligned that this is a good building block

PR #197, opened 2015-11-22
Graphing x name

orangejenny:
Support for https://github.com/dimagi/commcare-odk/pull/837 which adds tooltips. This allows the user to configure the label for x-values in the tooltip.


phillipm:
this PR should target the `graphing` branch, right?

orangejenny:
it doesn't matter for this one, the old AChartEngine-based code will just ignore it

PR #198, opened 2015-11-24
Use new fixture key for J2me storage

wpride:
Fix for http://manage.dimagi.com/default.asp?183837

We updated the key in the storage refactor and this wasn't reflected here


wpride:
@ctsims

PR #202, opened 2015-11-24
Support for bar-color

orangejenny:
TDH wants to support bar-specific colors so you can highlight the current user's data in mobile UCR. Since graphing doesn't have the concept of point-specific configuration, this is implemented by introducing a new series-level parameter, `bar-color` which accepts a function. During evaluation, the function is applied to each node in the series, building up an array, and the array is then encoded in JSON and used to overwrite the original function.

http://manage.dimagi.com/default.asp?187761

### DIFF
--- a/backend/src/org/commcare/suite/model/graph/Graph.java
+++ b/backend/src/org/commcare/suite/model/graph/Graph.java
@@ -37,9 +37,9 @@ public class Graph implements Externalizable, DetailTemplate, Configurable {
     private Vector<Annotation> mAnnotations;
 
     public Graph() {
-        mSeries = new Vector<>();
-        mConfiguration = new Hashtable<>();
-        mAnnotations = new Vector<>();
+        mSeries = new Vector<XYSeries>();
+        mConfiguration = new Hashtable<String, Text>();
+        mAnnotations = new Vector<Annotation>();
     }
 
     public String getType() {
@@ -146,9 +146,9 @@ public class Graph implements Externalizable, DetailTemplate, Configurable {
     private void evaluateSeries(GraphData graphData, EvaluationContext context) {
         try {
             for (XYSeries s : mSeries) {
-                Hashtable<String, Text> expandableConfiguration = new Hashtable<>();
-                for (Enumeration<String> e = s.getExpandableConfigurationKeys(); e.hasMoreElements();) {
-                    String key = e.nextElement();
+                Hashtable<String, Text> expandableConfiguration = new Hashtable<String, Text>();
+                for (Enumeration e = s.getExpandableConfigurationKeys(); e.hasMoreElements();) {
+                    String key = (String) e.nextElement();
                     Text value = s.getConfiguration(key);
                     if (value != null) {
                         expandableConfiguration.put(key, value);
@@ -162,8 +162,8 @@ public class Graph implements Externalizable, DetailTemplate, Configurable {
                 Hashtable<String, Vector<String>> expandedConfiguration = new Hashtable();
                 for (TreeReference ref : refList) {
                     EvaluationContext refContext = new EvaluationContext(seriesContext, ref);
-                    for (Enumeration<String> e = expandableConfiguration.keys(); e.hasMoreElements();) {
-                        String key = e.nextElement();
+                    for (Enumeration e = expandableConfiguration.keys(); e.hasMoreElements();) {
+                        String key = (String) e.nextElement();
                         if (!expandedConfiguration.containsKey(key)) {
                             expandedConfiguration.put(key, new Vector<String>());
                         }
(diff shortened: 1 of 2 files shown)


backend/src/org/commcare/suite/model/graph/Graph.java
```diff
@@ -173,6 +185,27 @@ private void evaluateSeries(GraphData graphData, EvaluationContext context) {
...
+                    String json = "";
+                    for (String pointValue : expandedConfiguration.get(key)) {
+                        json += "," + "'" + pointValue.replaceAll("'", "&apos;") + "'";
+                    }
+                    json = "[" + json.substring(1) + "]";
+                    Text value = Text.PlainText(json);
+                    s.setConfiguration(key, value);
+                }
+
+                // Handle configuration after data, since data processing may update configuration
```
orangejenny: Alternatively, the expanded configuration could be stored in a new configuration key, in case the Graph gets evaluated multiple times. Seemed slightly cleaner to have fewer keys and to have the XPath function gone by the time cc-odk gets the Graph.

backend/src/org/commcare/suite/model/graph/Graph.java
```diff
@@ -173,6 +185,27 @@ private void evaluateSeries(GraphData graphData, EvaluationContext context) {
                     }
                 }
                 graphData.addSeries(seriesData);
+
+                for (Enumeration<String> e = expandedConfiguration.keys(); e.hasMoreElements();) {
+                    String key = e.nextElement();
+                    String json = "";
```
orangejenny: Java 1.3 predates JSON... :sob:

backend/src/org/commcare/suite/model/graph/Graph.java
```diff
@@ -173,6 +185,27 @@ private void evaluateSeries(GraphData graphData, EvaluationContext context) {
...
+                    for (String pointValue : expandedConfiguration.get(key)) {
+                        json += "," + "'" + pointValue.replaceAll("'", "&apos;") + "'";
```
orangejenny: This could be more robust, but given that this is only for colors, there shouldn't be any special characters in the first place.

backend/src/org/commcare/suite/model/graph/Graph.java
```diff
@@ -37,9 +37,9 @@
     private Vector<Annotation> mAnnotations;
 
     public Graph() {
-        mSeries = new Vector<XYSeries>();
-        mConfiguration = new Hashtable<String, Text>();
-        mAnnotations = new Vector<Annotation>();
+        mSeries = new Vector<>();
+        mConfiguration = new Hashtable<>();
+        mAnnotations = new Vector<>();
```
phillipm: Leaving out type annotations at the commcare level will break J2ME builds :(

backend/src/org/commcare/suite/model/graph/Graph.java
```diff
@@ -141,26 +141,35 @@ private void evaluateConfiguration(Configurable template, ConfigurableData data,
...
 
     /*
-     * Helper for evaluate. Looks at a single series.
+     * Helper for evaluate. Looks at all series.
      */
     private void evaluateSeries(GraphData graphData, EvaluationContext context) {
         try {
             for (XYSeries s : mSeries) {
+                Hashtable<String, Text> expandableConfiguration = new Hashtable<>();
+                for (Enumeration<String> e = s.getExpandableConfigurationKeys(); e.hasMoreElements();) {
```
phillipm: I think that type annotations inside of for-each loops break J2ME as well.
you'll need `for(Enumberation e = ...)`

backend/src/org/commcare/suite/model/graph/Graph.java
```diff
@@ -141,26 +141,35 @@ private void evaluateConfiguration(Configurable template, ConfigurableData data,
...
+                Hashtable<String, Vector<String>> expandedConfiguration = new Hashtable();
                 for (TreeReference ref : refList) {
                     EvaluationContext refContext = new EvaluationContext(seriesContext, ref);
+                    for (Enumeration<String> e = expandableConfiguration.keys(); e.hasMoreElements();) {
+                        String key = e.nextElement();
+                        if (!expandedConfiguration.containsKey(key)) {
+                            expandedConfiguration.put(key, new Vector<String>());
+                        }
+                        String value = expandableConfiguration.get(key).evaluate(refContext);
+                        expandedConfiguration.get(key).addElement(value);
```
phillipm: Seeing `expandableConfiguration` and expandedConfiguration` right next to each other it is hard to differentiate them. What about something like `rawConfig` and `expandedConfig`?

backend/src/org/commcare/suite/model/graph/Graph.java
phillipm: You should use a `StringBuilder` since String are immutable and operating over them will always create new objects.

orangejenny:
@phillipm should be good now

PR #214, opened 2015-12-04
Move LivePrototypeFactory to javarosa

phillipm:
Move LivePrototypeFactory to javarosa repo so we can also use it in tests there


phillipm:
Re-opening to target 2.25

PR #218, opened 2015-12-08
Parent index fix

wpride:


wpride:
this fixes http://manage.dimagi.com/default.asp?188774 in CloudCare with the new backend

wpride:
also contains some other code for manipulating the sqlite DB files that should only impact Tf. I'm not sure whether we absolutely need this in 2.25 or if its OK just to get on HQ

wpride:
cross https://github.com/dimagi/javarosa/pull/205/files

api/org/commcare/api/persistence/SqlHelper.java
```diff
@@ -79,13 +79,14 @@ public static PreparedStatement prepareTableSelectStatement(Connection c,
                                                                 String storageKey,
                                                                 String[] fields,
                                                                 String[] values,
-                                                                Persistable p) {
+                                                                Persistable p) throws IllegalArgumentException{
```
phillipm: `IllegalArgumentException` is a runtime exception, so it shouldn't be part of method signatures.

api/org/commcare/api/persistence/SqliteIndexedStorageUtility.java
```diff
@@ -37,12 +37,23 @@
     private final Class<T> prototype;
     private final String tableName;
     private final String sandboxId;
+    private File databaseFolder;
```
phillipm: can be `final`

modern/org/commcare/modern/database/DatabaseHelper.java
```diff
@@ -76,6 +76,10 @@
 
             set = true;
         }
+        // we couldn't match any of the fields to our columns
+        if(!set){
+            throw new IllegalArgumentException();
```
phillipm: `throw new IllegalArgumentException()` -> `throw new IllegalArgumentException("Unable to match provided fields with columns.")` or something along those lines

phillipm:
Can you leave a PR comment around what parts fix the non-main instance bug you linked to, and which parts of the PR address DB file manipulation? I'm having a hard time seeing what part fixes the bug.

api/org/commcare/api/persistence/SqlHelper.java
ctsims: I'm not sure I agree with that, actually. I think it's fine to throw typed exceptions and document in the method signature when they are thrown so people can plan around using them semantically.

In general non-runtime exceptions are a better fit for that, but if we are using the exceptions semantically in any way, I think documenting them in the signature is better than not.

modern/org/commcare/modern/database/DatabaseHelper.java
ctsims: Agreed.

api/org/commcare/api/persistence/SqlSandboxUtils.java
```diff
@@ -13,13 +13,28 @@
...
+
     public static UserSqlSandbox getStaticStorage(String username) {
-        return new UserSqlSandbox(username);
+        return new UserSqlSandbox(username, null);
     }
 
     public static void deleteDatabaseFolder(){
-        File databaseFolder = new File(SqliteIndexedStorageUtility.DATABASE_FOLDER);
-        if(databaseFolder.exists()){
+        deleteDatabaseFolder(null);
```
ctsims: It makes way more sense to me to set the argument here to be SqliteIndexedStorageUtility.DATABASE_FOLDER rather than swithing on Null below.

api/org/commcare/api/persistence/SqlSandboxUtils.java
```diff
@@ -13,13 +13,28 @@
  * @author wspride
  */
 public class SqlSandboxUtils {
+
+    public static UserSqlSandbox getStaticStorage(String username, String path) {
+        System.out.println("Get STatic storage path: " + path);
```
ctsims: This shoudln't be here

api/org/commcare/api/persistence/SqlSandboxUtils.java
```diff
@@ -13,13 +13,28 @@
  * @author wspride
  */
 public class SqlSandboxUtils {
+
+    public static UserSqlSandbox getStaticStorage(String username, String path) {
```
ctsims: Did you include this to write a new test to fix the bug? If so is that test included anywhere? There's new code here, but no new code paths that use it

modern/org/commcare/modern/database/DatabaseHelper.java
```diff
@@ -76,6 +76,10 @@
 
             set = true;
         }
+        // we couldn't match any of the fields to our columns
+        if(!set){
+            throw new IllegalArgumentException("Unable to match provided fields with columns.");
```
wpride: @phillipm this is the part that fixes the bug

api/org/commcare/api/persistence/SqlHelper.java
phillipm: I think a good middle-ground is to write a `@throws SomeRuntimeException` javadoc entry for runtime exceptions, instead of adding them to the method signature. Adding a `throws RuntimeException` to a method signature has no semantic effect on how the method or how it is invoked, thus I would argue that it should be avoided and the corresponding information should be moved to comments, where things that don't have semantic effects should live.

phillipm:
Looks good to me, @ctsims?

api/org/commcare/api/persistence/SqlHelper.java
ctsims: Adding to javadoc is definitely the primary utility. @wpride can you make sure to add the circumstances under which this gets thrown as a field in the method level javadoc comment? I'm fine with following phillip's pattern of not adding to the signature if there's a doc comment

api/org/commcare/api/persistence/SqlSandboxUtils.java
```diff
@@ -13,13 +13,25 @@
...
+    /**
+     * Used by touchforms
+     */
+    public static UserSqlSandbox getStaticStorage(String username, String path) {
+        return new UserSqlSandbox(username, path);
+    }
+
     public static UserSqlSandbox getStaticStorage(String username) {
-        return new UserSqlSandbox(username);
+        return new UserSqlSandbox(username, null);
```
ctsims: There's a UserSqlSandbox constructor that doens't require an argument for the second field, we should use that instead of assuming what the default should be for clarity

api/org/commcare/api/persistence/SqliteIndexedStorageUtility.java
```diff
@@ -37,12 +37,23 @@
...
+        this(prototype, sandboxId, tableName, null);
+    }
+
+    public SqliteIndexedStorageUtility(Class<T> prototype, String sandboxId, String tableName, String databasePath) {
         this.tableName = tableName;
         this.sandboxId = sandboxId;
         this.prototype = prototype;
 
+        if(databasePath == null){
+            databaseFolder = new File(DATABASE_FOLDER);
```
ctsims: Is it possible to pass this in from outside such that there is never a dependency between two different classes "Agreeing" on the default? It sketches me out in so many places we check for null and set to the default if so. Feels like it would be more consistent and reliable if all of these classes required the path location, and there was just one central place that set what it was if one was not provided.

ctsims: I'm going to double down on that, actually, because it looks like we're basically doing this anyway, which means some of these code paths are never triggered.

ctsims:
@wpride @phillipm Two final comments.

The second one boils down to having everything that _could_ rely on a path rely on a path (not optionally) and having one concrete place in the code where we chose between a default or a provided path. I don't see a value to having an implicit dependence spread out on it being optional.

PR #221, opened 2015-12-09
Store fixtures in files instead of DB

phillipm:
Make non-J2ME fixtures be stored in files instead of the db 

cross-requested with https://github.com/dimagi/commcare-odk/pull/906 and https://github.com/dimagi/javarosa/pull/209


ctsims:
Looks good

PR #223, opened 2015-12-09
Validate cases DAG before case purge

amstone326:
Fix for: http://manage.dimagi.com/default.asp?188630

When cases on the phone have an index to a case that no longer exists on the phone for some reason, an edge is created in the case DAG that points to a non-existent node, causing a NPE. This PR prevents that crash by removing any such invalid edges from the DAG. It also goes a step further and, for any cases that are made invalid by the non-existence of that parent node, removes that case node from the DAG and purges the case from the phone.

Cross: https://github.com/dimagi/javarosa/pull/212, https://github.com/dimagi/commcare-odk/pull/910


amstone326:
@ctsims This is ready now. I've been in the weeds a bit on this for a while in terms of fixing edge cases in tests, so definitely worth putting fresh eyes on the general approach to make sure everything seems legit.

Also, not sure why github is registering this as not finding any tests. The tests are definitely all running and passing on my machine locally. Any ideas?

amstone326:
@ctsims Also forgot to say-- this is obviously for master. If you decide after looking at it that we definitely want it in 2.25, I'll duplicate it over.

ctsims:
Ah, good to know. If it's complex that seems like it makes sense to me.

Did you update the wiki docs with the info about what should happen if
there are missing cases in the process? We can go over how to phrase those
updates if you want.

-Clayton

amstone326:
@ctsims No, I didn'

amstone326:
@ctsims No I didn't, was going to wait till this was actually merged. But it would be very helpful to get a bit of help with wording that, let's do that once you've reviewed?

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -43,7 +45,10 @@
 
     private static final int STATUS_OPEN = 16;
 
-    Vector<Integer> idsToRemove = new Vector<Integer>();
+    Vector<Integer> idsToRemove = new Vector<>();
```
phillipm: `private final`

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -65,17 +70,16 @@ public CasePurgeFilter(IStorageUtilityIndexed<Case> caseStorage, Vector<String>
...
 
     private void setIdsToRemoveWithNewExtensions(IStorageUtilityIndexed<Case> caseStorage, Vector<String> owners) {
-        //Create a DAG. The Index will be the case GUID. The Nodes will be a int array containing
-        //[STATUS_FLAGS, storageid], Edges are a string representing the relationship between the
-        //nodes, which is one of the Case Index relationships (IE: parent, extension)
-        DAG<String, int[], String> graph = new DAG<String, int[], String>();
+        // Create a DAG. The Index will be the case GUID. The Nodes will be a int array containing
+        // [STATUS_FLAGS, storageid], Edges are a string representing the relationship between the
+        // nodes, which is one of the Case Index relationships (IE: parent, extension)
+        internalCaseDAG = new DAG<>();
```
phillipm: You can't remove the type info here or j2me will crash. so it needs to be:
`internalCaseDAG` = new DAG<String, int[], String<>();`

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -222,7 +227,96 @@ private void propogateMarkToDAG(DAG<String, int[], String> dag, boolean walkFrom
...
+        // parents because they are still valid)
+        Vector<Edge<String, String>> parentCases = internalCaseDAG.getChildren(indexOfRemovedNode);
+        for (Edge<String, String> parent : parentCases) {
+            accumulatedEdgesToRemove.add(new String[]{indexOfRemovedNode, parent.i});
+        }
+
+        // Once all edges to/from this node have been removed, delete the node itself from the
+        // DAG, and add it to the list of cases to be purged
+        int storageIdOfRemovedNode = internalCaseDAG.getNode(indexOfRemovedNode)[1];
+        internalCaseDAG.removeNode(indexOfRemovedNode);
```
phillipm: If you make `removeNode` have the same type signature as `Hashtable.remove` then you don't need to call `getNode` in the line above but rather:
```
int storageIdOfRemovedNode = internalCaseDAG.removeNode(indexOfRemovedNode)[1];
```

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -222,7 +227,96 @@ private void propogateMarkToDAG(DAG<String, int[], String> dag, boolean walkFrom
...
+     * Traverse the graph to remove any edges to empty nodes (which are created when a child
+     * makes a placeholder index to a parent, but then the parent does not actually exist on the
+     * phone for some reason). Then remove any nodes that are made invalid by that parent node
+     * not existing
+     *
+     * @return Whether or not this method invocation removed any invalid edges from the DAG
+     */
+    private boolean removeInvalidEdges() {
+        Hashtable<String, Vector<Edge<String, String>>> allEdges = internalCaseDAG.getEdges();
+        Set<String> childOfNonexistentParent = new HashSet<>();
```
phillipm: `Set` and `HashSet` aren't available in J2ME java :(

tests/test/org/commcare/cases/util/test/CasePurgeRegressions.java
```diff
@@ -35,4 +51,230 @@ public void testSimpleExtensions() throws Exception {
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static Set<String[]> getSimpleFormEdges(
+            Hashtable<String, Vector<DAG.Edge<String, String>>> edges) {
+        Set<String[]> simpleFormEdges = new HashSet<>();
+        for (String sourceIndex : edges.keySet()) {
```
phillipm: use `keys` instead of `keySet` for J2ME compat

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -222,7 +227,96 @@ private void propogateMarkToDAG(DAG<String, int[], String> dag, boolean walkFrom
...
+        Set<String[]> edgesToRemove = new HashSet<>();
+        boolean invalidEdgesRemoved = false;
+        for (String originIndex : allEdges.keySet()) {
```
phillipm: Use 'keys' instead of 'keySet' for J2ME compat

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -222,7 +227,96 @@ private void propogateMarkToDAG(DAG<String, int[], String> dag, boolean walkFrom
...
+                                        Set<String[]> accumulatedEdgesToRemove) {
+        // Wording is confusing here -- because all edges in this graph are from a child case to
+        // a parent case, calling getParents() for a node returns all of its child/extension cases
+        Vector<Edge<String, String>> childCases = internalCaseDAG.getParents(indexOfRemovedNode);
+        for (Edge<String, String> child : childCases) {
+            // Want to remove the edge from child case to parent case
+            accumulatedEdgesToRemove.add(new String[]{child.i, indexOfRemovedNode});
+
+            // Recurse on child case
+            accumulatedEdgesToRemove = removeNodeAndPropagate(child.i, accumulatedEdgesToRemove);
```
phillipm: This assignment shouldn't be needed, right?
just `removeNodeAndPropagate(child.i, accumulatedEdgesToRemove);`

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -119,21 +123,22 @@ private void setIdsToRemoveWithNewExtensions(IStorageUtilityIndexed<Case> caseSt
...
-                graph.setEdge(c.getCaseId(), index.getTarget(), index.getRelationship());
+                internalCaseDAG.setEdge(c.getCaseId(), index.getTarget(), index.getRelationship());
             }
             indexHolder.removeAllElements();
         }
 
-        propogateRelevance(graph);
-        propogateAvailabile(graph);
-        propogateLive(graph);
+        invalidEdgesWereRemoved = removeInvalidEdges();
```
ctsims: The fact that this directly modifies the DAG doesn't really feel right to me. Is it super hard to feed it down the line and  have it return back a list of the cases to be removed? you could pass in the "toRemove" vector as an object and have the method populate it, which would let you retain the scalar boolean return code.

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -222,7 +227,96 @@ private void propogateMarkToDAG(DAG<String, int[], String> dag, boolean walkFrom
...
+                    edgesToRemove.add(new String[]{originIndex, targetIndex});
+                    childOfNonexistentParent.add(originIndex);
+                }
+            }
+        }
+
+        // Any case node with a nonexistent parent should be removed from the graph, as should
+        // all of its descendants, and any edges to or from it
+        for (String index : childOfNonexistentParent) {
+            edgesToRemove = removeNodeAndPropagate(index, edgesToRemove);
```
ctsims: I'm a big confused, this both takes in edgesToRemove as an assignment but also passes it in as an argument. It's an accumulator, right? If so we shouldn't need to be passing it back.

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -222,7 +227,96 @@ private void propogateMarkToDAG(DAG<String, int[], String> dag, boolean walkFrom
...
+        }
+
+        // Do actual edge removal after we finish all iterations through the edge list, to prevent
+        // ConcurrentModificationException
+        for (String[] edge : edgesToRemove) {
+            internalCaseDAG.removeEdge(edge[0], edge[1]);
```
ctsims: one thing to note here: the edge set isn't necessary unique, since String[] is compared by reference, not value, so if someone added

new String[] {"A","B"}
new String[] {"A","B"}

it would be in there twice, which might cause problems here? Not sure if the DAG type is happy to receive a remove request for an edge that doesn't exist.

cases/src/org/commcare/cases/util/CasePurgeFilter.java
amstone326: @phillipm Thanks, I think for some reason at the time I was thinking this is only an issue in javarosa.

cases/src/org/commcare/cases/util/CasePurgeFilter.java
amstone326: This is fine, the DAG.removeEdge() handles a call for a non-existent edge gracefully.

cases/src/org/commcare/cases/util/CasePurgeFilter.java
amstone326: True, fixed.

cases/src/org/commcare/cases/util/CasePurgeFilter.java
amstone326: @ctsims Not sure I understand exactly what you're looking for here, but I think I addressed this by renaming removeInvalidEdges() to getInvalidEdges(), and having it return the list of edges to remove, and then having the caller method loop through and do the actual removal of edges from the DAG.

amstone326:
@ctsims, @phillipm Addressed comments, should be ready to merge.

amstone326:
This should be ready to merge, except that I'm still not sure why github is showing no tests. Don't know if we need to be concerned about that.

phillipm:
Tests don't work yet with PRs that have cross-requested dependencies.

tests/test/org/commcare/cases/util/test/CasePurgeRegressions.java
```diff
@@ -35,4 +51,230 @@ public void testSimpleExtensions() throws Exception {
...
+        nodesExpectedToBeLeft.add("case_one");
+        nodesExpectedToBeLeft.add("case_two");
+
+        Set<String[]> edgesExpectedToBeLeft = new HashSet<>();
+        edgesExpectedToBeLeft.add(new String[]{"case_two", "case_one"});
+
+        // Check that the edges and nodes still present in the graph are as expected
+        DAG<String, int[], String> internalCaseGraph = filter.getInternalCaseGraph();
+        checkProperNodesPresent(nodesExpectedToBeLeft, internalCaseGraph);
+        //checkProperEdgesPresent(edgesExpectedToBeLeft, internalCaseGraph);
```
phillipm: commented out line should be removed, and also `edgesExpectedToBeLeft` is never read from (because of commented out code)

tests/test/org/commcare/cases/util/test/CasePurgeRegressions.java
```diff
@@ -1,15 +1,30 @@
 package org.commcare.cases.util.test;
 
+import org.commcare.cases.model.Case;
+import org.javarosa.core.services.storage.IStorageIterator;
+import org.javarosa.core.services.storage.IStorageUtilityIndexed;
+import org.javarosa.core.services.storage.util.DummyIndexedStorageUtility;
+import org.javarosa.core.util.DataUtil;
+import org.junit.Assert;
```
phillipm: Some unused imports

tests/test/org/commcare/cases/util/test/CasePurgeRegressions.java
amstone326: Those weren't supposed to be commented out, thanks for catching.

amstone326:
@phillipm Addressed comments

PR #229, opened 2015-12-23
Add test for cousins

proteusvacuum:
This test was failing on HQ. I don't imagine it will fail here (was an issue with my implementation), but just wanted to keep our json file in sync.
@phillipm @ctsims


proteusvacuum:
wow, this test suite runs fast. nice!

PR #233, opened 2016-01-05
Allow the suite parser to skip adding resources

ctsims:
X-Requested (but can be pulled on its own) with CommCare Android

### DIFF
--- a/backend/src/org/commcare/xml/SuiteParser.java
+++ b/backend/src/org/commcare/xml/SuiteParser.java
@@ -1,6 +1,3 @@
-/**
- *
- */
 package org.commcare.xml;
 
 import org.commcare.resources.model.Resource;
@@ -25,6 +22,13 @@ import java.util.Hashtable;
 import java.util.Vector;
 
 /**
+ * Parses a suite file resource and creates the associated object 
+ * containing the menu, detail, entry, etc definitions. This parser
+ * will also create models for any resource installers that are defined
+ * by the suite file and add them to the resource table provided
+ * with the suite resource as the parent, that behavior can be skipped
+ * by setting a flag if the resources have already been promised.
+ * 
  * @author ctsims
  */
 public class SuiteParser extends ElementParser<Suite> {
@@ -151,7 +155,6 @@ public class SuiteParser extends ElementParser<Suite> {
     /**
      * If set to true, the parser won't process adding incoming resources to the resource table.
      * This is helpful if the suite is being processed during a non-install phase
-     * @param skipResources
      */
     public void setSkipResources(boolean skipResources) {
         this.skipResources = skipResources;
--- a/backend/src/org/commcare/xml/SuiteParser.java
+++ b/backend/src/org/commcare/xml/SuiteParser.java
@@ -33,6 +33,8 @@ public class SuiteParser extends ElementParser<Suite> {
     ResourceTable table;
     String resourceGuid;
 
+    private boolean skipResources = false;
+
     public SuiteParser(InputStream suiteStream, ResourceTable table, String resourceGuid) throws IOException {
         super(ElementParser.instantiateParser(suiteStream));
         this.table = table;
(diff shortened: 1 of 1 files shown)


backend/src/org/commcare/xml/SuiteParser.java
```diff
@@ -75,19 +77,25 @@ public Suite parse() throws InvalidStructureException, IOException, XmlPullParse
...
                             Resource r = new ResourceParser(parser, maximumResourceAuthority).parse();
-                            table.addResource(r, table.getInstallers().getMediaInstaller(path), resourceGuid);
+                            if(!skipResources) {
+                                table.addResource(r, table.getInstallers().getMediaInstaller(path), resourceGuid);
+                            }
                         }
                     } else if (parser.getName().toLowerCase().equals("xform")) {
                         //skip xform stuff for now
                         parser.nextTag();
                         Resource r = new ResourceParser(parser, maximumResourceAuthority).parse();
```
wpride: Can those line go inside the if() block?

backend/src/org/commcare/xml/SuiteParser.java
```diff
@@ -75,19 +77,25 @@ public Suite parse() throws InvalidStructureException, IOException, XmlPullParse
...
-                        table.addResource(r, table.getInstallers().getLocaleFileInstaller(localeKey), resourceGuid);
+                        if(!skipResources) {
+                            table.addResource(r, table.getInstallers().getLocaleFileInstaller(localeKey), resourceGuid);
+                        }
                     } else if (parser.getName().toLowerCase().equals("media")) {
                         String path = parser.getAttributeValue(null, "path");
                         //Can be an arbitrary number of resources inside of a media block.
                         while (this.nextTagInBlock("media")) {
                             Resource r = new ResourceParser(parser, maximumResourceAuthority).parse();
```
wpride: and this line

backend/src/org/commcare/xml/SuiteParser.java
```diff
@@ -140,4 +148,12 @@ protected boolean inValidationMode() {
         return false;
     }
 
+    /**
+     * If set to true, the parser won't process adding incoming resources to the resource table.
+     * This is helpful if the suite is being processed during a non-install phase
+     * @param skipResources
```
phillipm: Remove unused javadoc param entry

backend/src/org/commcare/xml/SuiteParser.java
```diff
@@ -33,6 +33,8 @@
     ResourceTable table;
     String resourceGuid;
 
+    private boolean skipResources = false;
```
phillipm: Could fixture processing also be skipped on app init?

backend/src/org/commcare/xml/SuiteParser.java
```diff
@@ -33,6 +33,8 @@
...
+
     public SuiteParser(InputStream suiteStream, ResourceTable table, String resourceGuid) throws IOException {
```
phillipm: Will you add a class level doc? Something saying that this is used to used to build menus, entry actions, display details, get the app version, and install resources. I found myself wondering why we even run the suite parser on app init, and having such documentation would have answered it for me.

ctsims: Sure, will add

backend/src/org/commcare/xml/SuiteParser.java
ctsims: Hm, plausible. The app fixtures are tiny right now, so I'm inclined to keep doing it until I have a chance to consider it more deeply

backend/src/org/commcare/xml/SuiteParser.java
ctsims: We're stream processing, so we can't in this case, we need to consume the full document in order to move on, and can't skip stuff.

ctsims:
@phillipm Addressed your comments. @wpride Unfortunately we can't quite just skip due to the way we're streaming rather than processing the full DOM. We might be able to skip the elements, but I'm not sure it's actually much faster.

PR #235, opened 2016-01-08
Refactor Detail related code

phillipm:
Remove background field from `DetailFragment` and code surrounding it. Also other random refactors.

review commit-by-commit

cross-requested with https://github.com/dimagi/commcare-odk/pull/967


backend/src/org/commcare/xml/DetailParser.java
```diff
@@ -223,10 +217,10 @@ public Detail parse() throws InvalidStructureException, IOException, XmlPullPars
                         }
                     } else if (name.equals("background")) {
                         if (nextTagInBlock("background")) {
+                            // parse but don't use
+                            // TODO PLM: is there a way to just skip over entire tag?
                             checkNode("text");
-                            //Get it if so
-                            background = new TextParser(parser).parse();
-                            builder.setBackground(background);
+                            new TextParser(parser).parse();
```
phillipm: Any tips on addressing this TODO?

ctsims: Hey, there is actually a method in our internal parser to skip a tag body. Would be good to add a detail parse including such a tag to a regression to make sure the jump method actually works, though, since jumping over tags in stream parsing can get dicey. Grab me if you want to do so before we pull this, otherwise, I'm good to pull

amstone326:
@phillipm Good to merge but not sure if you still want to make a change based on Clayton's comment?

phillipm:
@amstone326 @ctsims Added a small regression for skipping over "background" tags found in a detail's field. Should be good to go now

tests/resources/complex_stack/suite.xml
```diff
@@ -51,6 +51,9 @@
                     <xpath function="case_name"/>
                 </text>
             </template>
+            <background>
+                <text>Regression for deprecated background tag</text>
+            </background>
```
phillipm: Added the regression to the stack test suite file. Doesn't really fit in, but didn't want to bother spinning up a new test and resource files. I can do that if you think it is worth it.

PR #241, opened 2016-01-13
Refactor CommCarePlatform

phillipm:


backend/src/org/commcare/util/CommCarePlatform.java
```diff
@@ -32,25 +30,23 @@
 public class CommCarePlatform implements CommCareInstance {
     // TODO: We should make this unique using the parser to invalidate this ID or something
     public static final String APP_PROFILE_RESOURCE_ID = "commcare-application-profile";
-
-    private Vector<Integer> suites;
```
phillipm: Pretty strange that  this isn't used... The ODK level `AndroidCommCarePlatform` class has an equivalent that is used, but I'm guess the J2ME code somehow doesn't use this...

backend/src/org/commcare/util/CommCarePlatform.java
```diff
@@ -145,26 +140,6 @@ public String getModuleNameForEntry(Entry entry) {
         return null;
     }
 
-    public static Vector<Resource> getResourceListFromProfile(ResourceTable master) {
```
phillipm: Defined in `ResourceManager`

PR #242, opened 2016-01-13
Distinguish between Entry and View at the class level

phillipm:
In prep for case migration feature, which requires extending suites to include something very similar to entries and views.

This PR breaks Entry up into `FormEntry` and `ViewEntry`, which share a lot of code, placed in `Entry`

cross-requested with https://github.com/dimagi/commcare-odk/pull/975


backend/src/org/commcare/suite/model/Entry.java
```diff
@@ -19,23 +19,15 @@
...
- * begin, and what the User Interface should
- * present to the user regarding these actions</p>
+ * Describes a user-initiated form entry action. Includes information that
+ * needs to be collected before that action can begin and what the UI should
+ * present to the user regarding this action.
  *
  * @author ctsims
  */
-public class Entry implements Externalizable, MenuDisplayable {
+public class Entry extends EntryBase {
```
amstone326: could consider renaming this class to FormEntry, and then you could call EntryBase just Entry

backend/src/org/commcare/suite/model/View.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Hashtable;
+import java.util.Vector;
+
+/**
+ * An entry action used for viewing data; main difference from an Entry is that
+ * it doesn't end in launching form entry.
+ *
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+public class View extends EntryBase {
```
amstone326: Can this have a more descriptive name? Feels weird to name something "View" when it is something much more specific than anything like the notion of Android's View class

backend/src/org/commcare/xml/EntryParser.java
```diff
@@ -18,22 +20,25 @@
...
+        return new EntryParser(parser, false);
+    }
+
+    public static EntryParser buildEntryParser(KXmlParser parser) {
+        return new EntryParser(parser, true);
+    }
+
+    @Override
+    public EntryBase parse() throws InvalidStructureException, IOException, XmlPullParserException {
         String block = isEntry ? "entry" : "view";
```
amstone326: Purely for my own curiosity/confusion, but how/why was this code already here if you just added the notion of a View?

phillipm: Views have existed for a while, I just made them explicit classes instead of a distinction derived from the internal state of an `Entry` object (I think when the namespace is `null`)

amstone326: Ah, cool. What was the usage of views previously?

phillipm: It is used when you want to view a entity detail without having it lead to form entry.

backend/src/org/commcare/suite/model/Entry.java
```diff
@@ -41,20 +37,25 @@
...
         this.display = display;
         this.data = data;
-        xFormNamespace = formNamespace;
         this.instances = instances;
         this.stackOperations = stackOperations;
         this.assertions = assertions;
     }
 
+    public boolean isView() {
+        return this instanceof ViewEntry;
```
ctsims: Wouldn't it make more sense to do this through the polymorphism of the objects? Seems convoluted to couple at this level.

PR #245, opened 2016-01-26
Report missing and removed cases during sync to device logs

amstone326:
cross: https://github.com/dimagi/commcare-odk/pull/1017

### DIFF
--- a/cases/src/org/commcare/cases/util/CasePurgeFilter.java
+++ b/cases/src/org/commcare/cases/util/CasePurgeFilter.java
@@ -45,8 +45,18 @@ public class CasePurgeFilter extends EntityFilter<Case> {
 
     private final Vector<Integer> idsToRemove = new Vector<Integer>();
 
+    // The index is a string containing the case GUID. The Nodes will be a int array containing
+    // [STATUS_FLAGS, storageid]. Edges are a string representing the relationship between the
+    // nodes, which is one of the Case Index relationships (IE: parent, extension)
     private DAG<String, int[], String> internalCaseDAG;
+
+    // Flag that gets checked by DataPullTask, in order to report to device logs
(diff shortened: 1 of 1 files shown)


cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -315,6 +324,23 @@ public boolean invalidEdgesWereRemoved() {
...
+    public String getRemovedCasesString() {
+        return flattenVectorOfStrings(this.casesRemovedDueToMissingCases);
+    }
+
+    private static String flattenVectorOfStrings(Vector<String> v) {
+        String s = "";
+        for (String caseId : v) {
+            s = s + caseId + ", ";
+        }
+        s = s.substring(0, s.length()-2);
```
ctsims: This will throw an exception if there are no strings in the vector. I would honestly just space separate them and not worry about the hanging space at the end.

cases/src/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -315,6 +324,23 @@ public boolean invalidEdgesWereRemoved() {
...
+    public String getMissingCasesString() {
+        return flattenVectorOfStrings(this.missingCases);
+    }
+
+    public String getRemovedCasesString() {
+        return flattenVectorOfStrings(this.casesRemovedDueToMissingCases);
+    }
+
+    private static String flattenVectorOfStrings(Vector<String> v) {
+        String s = "";
```
ctsims: For what it's worth, is good to get into the pattern of using a StringBuilder for this kind of thing, which is easier on the heap.

amstone326:
@ctsims Comments addressed.

PR #248, opened 2016-01-29
Enable multiple actions in the case list

phillipm:
cross-requested with https://github.com/dimagi/commcare-odk/pull/1033


util/src/org/commcare/util/cli/EntityListSubscreen.java
```diff
@@ -132,6 +133,7 @@ public void prompt(PrintStream out) {
     @Override
     public boolean handleInputAndUpdateHost(String input, EntityScreen host) throws CommCareSessionException {
         if("action".equals(input) && mAction != null) {
+            // TODO PLM: handle multiple actions
             host.setPendingAction(mAction);
```
phillipm: @ctsims Not sure how to handle multiple actions in a detail for this specific context.

PR #251, opened 2016-02-02
Allow graph apps to be installed by commcare CLI

phillipm:
Currently, if you run the commcare cli on apps that have graphs in them it crashes on startup. Now it only crashes if you to load a view with a graph in it.

### DIFF
--- a/backend/src/org/commcare/xml/DetailParser.java
+++ b/backend/src/org/commcare/xml/DetailParser.java
@@ -232,6 +232,6 @@ public class DetailParser extends CommCareElementParser<Detail> {
     }
 
     protected GraphParser getGraphParser() {
-        return new GraphParser(parser);
+        return new DummyGraphParser(parser);
     }
 }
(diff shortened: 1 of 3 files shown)


backend/src/org/commcare/xml/DetailParser.java
```diff
@@ -148,10 +148,6 @@ public Detail parse() throws InvalidStructureException, IOException, XmlPullPars
                     parser.nextTag();
                     DetailTemplate template;
                     if (form.equals("graph")) {
-                        GraphParser gp = getGraphParser();
-                        if (gp == null) {
-                            throw new InvalidStructureException("No graph parser available", parser);
```
phillipm: @orangejenny Means that if you try to install an app w/ graphing on J2ME it won't fail immediately, just if you try to open something that uses the graphing. Hope that is a good failure mode...

orangejenny:
nice, looks good

PR #253, opened 2016-02-05
Synchronous server requests

phillipm:
depended on by https://github.com/dimagi/commcare-odk/pull/1059
cross-request: https://github.com/dimagi/javarosa/pull/289

building on work in https://github.com/dimagi/commcare/pull/250


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -60,6 +62,12 @@
     private String currentXmlns;
 
     /**
+     * Provides access to data instances in the evaluation context built from the session.
+     * Maps frame step IDs to data instances; only containing entries in the current frame.
+     */
+    private final Hashtable<String, ExternalDataInstance> instances;
```
ctsims: I'm confused about the key here. Is it the ID of the frame step, or is it the ID of the instance itself? Forcing the two to be the same could lead to a lot of confusion/ambiguity v. just recording which index the step is in the current frame.

phillipm: It should be the ID of the frame step (as noted)

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -370,6 +394,26 @@ public void setDatum(String keyId, String value) {
...
+     * Set a (xml) data instance as the result to a session query datum.
+     * The instance is available in session's evaluation context until the corresponding query frame is removed
+     */
+    public void setQueryDatum(ExternalDataInstance queryResultInstance) {
+        SessionDatum datum = getNeededDatum();
+        if (datum instanceof RemoteQueryDatum) {
+            StackFrameStep step =
+                    new StackFrameStep(SessionFrame.STATE_QUERY_REQUEST,
+                            datum.getDataId(), datum.getValue());
+            frame.pushStep(step);
```
ctsims: I'm curious about one thing here: why are we not injecting the instance straight into the frame? Seems odd to keep the two in sync this way.

phillipm: The alternative would be to traverse the stack frame in `getEvaluationContext` and pull out any stored instances. Not exactly sure why I took this approach, maybe to avoid needing to add `ExternalDataInstance` externalization logic to `StackFrameStep`.
Do you think storing associated instances in the frame step is better? I guess it does seem conceptually cleaner

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -773,6 +843,11 @@ public boolean isViewCommand(String command) {
         return entries.size() == 1 && entries.elementAt(0).isView();
     }
 
+    public boolean isSyncCommand(String command) {
+        Vector<Entry> entries = this.getEntriesForCommand(command);
+        return entries.size() == 1 && entries.elementAt(0) instanceof SyncEntry;
```
ctsims: Is there a reason we're doing this with a type check rather than a polymorphic check(IE: "entries.elementAt(0).isSync() or (0).getCommandType().equals(TYPE_SYNC))? The latter would be more "Best practice"y.

backend/src/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -0,0 +1,81 @@
...
+        this.queryDatum = queryDatum;
+        this.evaluationContext = evaluationContext;
+    }
+
+    public Hashtable<String, DisplayUnit> getNeededUserInputDisplays() {
+        return queryDatum.getUserQueryPrompts();
+    }
+
+    public Hashtable<String, String> getUserAnswers() {
+        return userAnswers;
```
ctsims: Is it possible for the user answers to be pre-populated in the same manner as the query params, but in a way that can be overriden, or is it a binary thing (some params are functions, some are user input)?

phillipm: No, but we could add it. What would be the use-case?

backend/src/org/commcare/session/CommCareSession.java
phillipm: Added `isSync` so that we don't have to externalize subclass tags.

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -192,6 +212,9 @@ public String getNeededData() {
```
amstone326: It won't let me comment nearer to the comment, but it would be good to update this method's javadoc to either say "1 of 6" or better yet just remove the numeric portion entirely since it's bound to get stale

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -192,6 +212,9 @@ public String getNeededData() {
 
         if (needDatum != null) {
             return needDatum;
+        } else if (entries.size() == 1 && entries.elementAt(0) instanceof SyncEntry) {
+            // one sync entry, with no needed data, matches the command; fire that request
```
amstone326: seems like an unnecessary comment that could potentially confuse more than it helps

backend/src/org/commcare/session/CommCareSession.java
amstone326: @phillipm I agree with this; it seems a lot more well-contained to make the instance a part of the `StackFrameStep` itself, instead of relying on keeping this random `instances` member variable in sync with the stack.

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -370,6 +393,26 @@ public void setDatum(String keyId, String value) {
...
+     * The instance is available in session's evaluation context until the corresponding query frame is removed
+     */
+    public void setQueryDatum(ExternalDataInstance queryResultInstance) {
+        SessionDatum datum = getNeededDatum();
+        if (datum instanceof RemoteQueryDatum) {
+            StackFrameStep step =
+                    new StackFrameStep(SessionFrame.STATE_QUERY_REQUEST,
+                            datum.getDataId(), datum.getValue());
+            frame.pushStep(step);
+            if (queryResultInstance != null) {
```
amstone326: would it ever be ok for this to be null?

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -403,19 +446,37 @@ private void syncState() {
...
             }
         }
+
+        for (String staleInstanceKey : staleInstanceKeys) {
+            instances.remove(staleInstanceKey);
+        }
+    }
+
+    private static Vector<String> getKeys(Hashtable<String, ?> table) {
+        // <3 <3 <3  J2ME  <3 <3 <3
```
amstone326: lol

backend/src/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -0,0 +1,90 @@
...
+            params.put(key, evaluatedExpr);
+        }
+        for (Enumeration e = userAnswers.keys(); e.hasMoreElements(); ) {
+            String key = (String)e.nextElement();
+            params.put(key, userAnswers.get(key));
+        }
+        return params;
+    }
+
+    public static String calculateHidden(XPathExpression expr,
```
amstone326: could have a better name

backend/src/org/commcare/suite/model/EntityDatum.java
```diff
@@ -119,6 +119,9 @@ public TreeReference getEntityFromID(EvaluationContext ec, String elementId) {
         //Get root nodeset
         TreeReference nodesetRef = this.getNodeset().clone();
         Vector<XPathExpression> predicates = nodesetRef.getPredicate(nodesetRef.size() - 1);
+        if (predicates == null) {
```
amstone326: was this just a bug previously that you're fixing here?

tests/test/org/commcare/backend/session/test/BasicSessionNavigationTests.java
```diff
@@ -129,4 +129,45 @@ public void testStepBackWithComputedDatum() {
...
 
+    @Test
+    public void testStepToSyncRequest() {
+        SessionWrapper session = mApp.getSession();
+
+        session.setCommand("patient-search");
+        Assert.assertEquals(session.getNeededData(), SessionFrame.STATE_QUERY_REQUEST);
+        session.setQueryDatum(null);
+
+        /*
```
amstone326: why is all of this commented out?

backend/src/org/commcare/suite/model/EntityDatum.java
phillipm: Not sure, but it looks like the way that `TreeReferenceLevel.getPredicates()` is implemented, it can return `null`, so this handles that case.

tests/test/org/commcare/backend/session/test/BasicSessionNavigationTests.java
phillipm: oops, it is hold over from before I introduced the `RemoteQuerySessionManager`. I'll remove it

backend/src/org/commcare/session/CommCareSession.java
phillipm: Sounds good, I updated the PR to store xml instances in the `StepFrameStep`

PR #258, opened 2016-02-19
Prevent empty case_name in case create block

phillipm:
Blindly implementing constraint to prevent `case_name` from being empty in a case `<create>` block. Please confirm that this looks like the correct approach.

Addresses this backlog ticket: http://manage.dimagi.com/default.asp?34579


wpride:
:+1: 

@phillipm fixed a failing test

PR #264, opened 2016-02-26
Parse empty fixtures correctly: don't swallow subsequent fixtures

phillipm:
Whenever empty fixtures were encountered they would swallow the rest of the restore file into them.

http://manage.dimagi.com/default.asp?158398

cross-request: https://github.com/dimagi/javarosa/pull/257

### DIFF
--- a/backend/src/org/commcare/xml/FixtureXmlParser.java
+++ b/backend/src/org/commcare/xml/FixtureXmlParser.java
@@ -50,11 +50,14 @@ public class FixtureXmlParser extends TransactionParser<FormInstance> {
 
         String userId = parser.getAttributeValue(null, "user_id");
 
-        //Get to the data root
-        parser.nextTag();
+        TreeElement root;
+        if (nextTagInBlock("fixture")) {
+            //TODO: We need to overwrite any matching records here.
+            root = new TreeElementParser(parser, 0, fixtureId).parse();
+        } else {
+            root = new TreeElement();
+        }
 
-        //TODO: We need to overwrite any matching records here.
-        TreeElement root = new TreeElementParser(parser, 0, fixtureId).parse();
         FormInstance instance = new FormInstance(root, fixtureId);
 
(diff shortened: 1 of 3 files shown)


backend/src/org/commcare/xml/FixtureXmlParser.java
```diff
@@ -50,11 +50,14 @@ public FormInstance parse() throws InvalidStructureException, IOException,
...
         String userId = parser.getAttributeValue(null, "user_id");
 
-        //Get to the data root
-        parser.nextTag();
+        TreeElement root;
+        if (nextTagInBlock("fixture")) {
+            //TODO: We need to overwrite any matching records here.
+            root = new TreeElementParser(parser, 0, fixtureId).parse();
+        } else {
+            root = new TreeElement();
```
phillipm: Does this seem like a good value for the root of an empty fixture?

cases/src/org/commcare/cases/instance/CaseChildElement.java
```diff
@@ -54,7 +54,6 @@ private CaseChildElement(CaseInstanceTreeElement parent) {
         this.mult = TreeReference.INDEX_TEMPLATE;
         this.caseId = null;
 
-        empty = new TreeElement();
```
phillipm: Unneeded, see following line

backend/src/org/commcare/xml/FixtureXmlParser.java
ctsims: Oh, do we need one? If we have an empty fixture, I don't think we need to add a record to our DB. I think an empty <fixture/> tag should either delete an existing fixture or be ignored entirely. We can't really make a meaningful "Base" element like this

phillipm: Cool, updated such that `parse()` returns `null` for empty fixtures. Do you forsee this breaking anything?

PR #265, opened 2016-02-26
Detect case index invalidation on commit

ctsims:
Adds functionality to the Case XML transaction parser to detect when and what caseID's are "disrupted" by a transaction. This can be used to detect intelligently when the environment should be triggering a case purge to reconcile the set of cases on the device.


tests/test/org/commcare/cases/util/test/CaseParseReindexTests.java
```diff
@@ -0,0 +1,104 @@
...
+ * Quick test to be able to restore a set of user data
+ * and ensure users and groups are properly being included
+ * in case purges.
+ *
+ * @author ctsims
+ */
+public class CaseParseReindexTests {
+
+    private MockUserDataSandbox sandbox;
+    private Vector<String> owners;
```
phillipm: Unused

cases/src/org/commcare/cases/model/Case.java
```diff
@@ -290,6 +299,8 @@ public void removeIndex(String indexName) {
 
         if (toRemove != null) {
             indices.removeElement(toRemove);
+            return true;
         }
+        return false;
```
phillipm: This line isn't covered by the test suite; pretty trivial line, but might be worth adding in

PR #267, opened 2016-03-09
J2ME Build Fix and Cert Redirect

ctsims:
Fixes the J2ME build (Serialization issues again)

Adds a platform redirect on J2ME where if an https request has certificate failures, it redirects the user to an optional URL that allows the user to create an exception on the device 

cross-request: https://github.com/dimagi/javarosa/pull/265

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ local.properties
 backend/bin/
 application/bin/
 application/build
+core/bin/
+core/build/
 util/bin
 application/dist
 application/lib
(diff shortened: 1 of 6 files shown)


wpride:
Curious: 
Why the core/* files needed to be added now?
Why we can remove the "Action" class from prototyping?

ctsims:
@wpride 

Core/* files fix the eclipse build.
Action class doesn't need to have its full name retained anymore because Aliza made it abstract.

PR #268, opened 2016-03-10
Filter case list by non-searchable callout results 

phillipm:
Allows for case list intent callouts to have a detail field associated with the response data returned from the callout. This allows data, which can be used for displaying or filtering, to be attached to cases in the case list.

- The `header` of the detail field is the column header in the case list. If the `width` of the `header` is 0, then the column will not be shown, but the result data will still be mapped onto cases, for use in case filter.
- The `template` of the detail field designates the mapping key used to map a callout result data point to a case. This will usually just be the `case_id`, but one could imagine a want for customization. 

```
    <lookup type="text/plain" action="com.simprints.id.IDENTIFY" image="jr://file/commcare/image/module1_case_list_lookup.png" name="Scan fingerprint">
      <extra key="deviceId" value="123"/>
      <response key="fake"/>
      <field>
        <header>
          <text>
            <locale id="m1.case_short.callout.header"/>
          </text>
        </header>
        <template>
          <text>
            <xpath function="@case_id"/>
          </text>
        </template>
      </field>
    </lookup>
```


backend/src/org/commcare/suite/model/Callout.java
```diff
@@ -98,19 +121,15 @@ public String getDisplayName() {
...
-
     public Hashtable<String, String> getExtras() {
         return extras;
     }
 
     public Vector<String> getResponses() {
         return responses;
     }
+
+    public DetailField getResponseDetail() {
```
amstone326: rename to `getResponseDetailField()`

backend/src/org/commcare/suite/model/StackFrameStep.java
```diff
@@ -86,19 +87,17 @@ public boolean getValueIsXPath() {
         return valueIsXpath;
     }
 
-    public void addExtra(String key, String value) {
+    /**
+     * @param value Must extend Externalizable or be a basic data type (String, Vector, etc)
```
amstone326: Is there no way to enforce this in the method signature?

amstone326: `public <T extends Externalizable> void addExtra(String key, T value)` would take care of the Externalizable portion, but I don't know of an encompassing class for "basic data types"

tests/test/org/commcare/backend/suite/model/test/StackFrameStepTests.java
```diff
@@ -45,6 +51,15 @@ public void setUp() {
 
         datumComputedStepV1 = new StackFrameStep(SessionFrame.STATE_DATUM_COMPUTED, "datum_val_id", "datum_val1");
         datumComputedStepV2 = new StackFrameStep(SessionFrame.STATE_DATUM_COMPUTED, "datum_val_id", "datum_val2");
+
+        // frame steps can store externalizable data, such as ints, Strings,
+        // or anything that implements Externalizable
+        stepWithExtras = new StackFrameStep(SessionFrame.STATE_DATUM_COMPUTED, "datum_val_id", "datum_val2");
+        stepWithExtras.addExtra("key", 123);
+
+        // Demonstrate how frame steps can't non-externalizable data
```
amstone326: typo in comment

backend/src/org/commcare/suite/model/StackFrameStep.java
phillipm: Yeah, I wanted to take this route, but `<T extends Externalizable>` doesn't work with basic data types

PR #270, opened 2016-03-11
Formplayer Changes

wpride:
OK so this PR became sort of a monster. Can be roughly broken down into three parts:

api/json/* classes - new classes for turning Javarosa and CommCare objects into JSON. Many of these functions are Java versions of what we were doing in Python before

~~SqlHelper fix where the update function wasn't correctly updating all the fields~~

~~Moving some classes around that FormPlayer needed to access (SessionWrapper, Roots/References) and some renaming~~

Ideally this would be all separate but it'd be hard to do that retroactively. Tests for the new classes currently live in the Formplayer repo. ~~Adding "Do not merge" tag for the moment since I want to polish a bit more but open for review now.~~

Two of the things above merged separately

cross-request: https://github.com/dimagi/javarosa/pull/282


wpride:
Removing don't pull now that build.gradle flavor is fixed

wpride:
this also contains https://github.com/dimagi/commcare/pull/271

util/src/org/commcare/util/cli/OptionsScreen.java
```diff
@@ -0,0 +1,8 @@
+package org.commcare.util.cli;
+
+/**
+ * Created by willpride on 2/26/16.
+ */
+public interface OptionsScreen {
+    public abstract String[] getOptions();
```
phillipm: remove `public abstract` those are implicitly set via the context

util/src/org/commcare/util/cli/OptionsScreen.java
```diff
@@ -0,0 +1,8 @@
+package org.commcare.util.cli;
+
+/**
+ * Created by willpride on 2/26/16.
+ */
+public interface OptionsScreen {
```
phillipm: Can you add a comment describing what options are?

modern/org/commcare/modern/session/SessionWrapper.java
```diff
@@ -12,10 +16,10 @@
  * @author ctsims
  */
 public class SessionWrapper extends CommCareSession {
-    
-    final UserSandbox mSandbox;
-    final CommCarePlatform mPlatform;
-    CLIInstanceInitializer initializer;
+
+    protected UserSandbox mSandbox;
+    protected CommCarePlatform mPlatform;
```
phillipm: can you add back the `final`s for these two fields?

modern/org/commcare/modern/session/SessionWrapper.java
```diff
@@ -59,4 +63,10 @@ public void clearVolitiles() {
     public void setComputedDatum() {
         setComputedDatum(getEvaluationContext());
     }
+
+    public FormDef loadFormByXmlns(String xmlns) {
```
phillipm: can you add a `@SuppressWarning("unused")` annotation so that we know this is used somewhere.
Also, this exact method is in `CommCareConfigEngine` and can be made static, so you should consider just calling that method or moving it here.

api/org/commcare/api/json/Walker.java
```diff
@@ -0,0 +1,95 @@
...
+
+            if(!relevant){
+                currentIndex = step(currentIndex, false);
+                break;
+            }
+
+            JSONObject obj = new JSONObject();
+            PromptToJson.parseQuestionType(fem, obj);
+
+            obj.put("relevant", relevant ? 1 : 0);
```
phillipm: `relevant` is always `true`

api/org/commcare/api/json/Walker.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Created by willpride on 12/8/15.
+ */
+public class Walker {
+
+    JSONArray compiler;
+    FormIndex parentIndex;
+    FormEntryController fec;
+    FormEntryModel fem;
```
phillipm: pretty sure all these fields can be `private final`

api/org/commcare/api/json/Walker.java
```diff
@@ -0,0 +1,95 @@
+package org.commcare.api.json;
+
+import org.javarosa.core.model.FormIndex;
+import org.javarosa.form.api.FormEntryController;
+import org.javarosa.form.api.FormEntryModel;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+/**
+ * Created by willpride on 12/8/15.
```
phillipm: What does a `Walker` walk over / do?
I would mention somewhere that it actually builds a `JSONArray` of a form (?) (not even sure what that means)

api/org/commcare/api/json/Walker.java
```diff
@@ -0,0 +1,95 @@
...
+        this.fem = fem;
+    }
+
+    public FormIndex step(FormIndex formIndex, boolean descend){
+        FormIndex nextIndex = fec.getAdjacentIndex(formIndex, true, descend);
+        fem.setQuestionIndex(nextIndex);
+        return nextIndex;
+    }
+
+    public boolean indexInScope(FormIndex formIndex){
```
phillipm: this and `step` should be private

phillipm:
Can you run the code indenter on all the new files you added in this PR?

api/org/commcare/api/json/PromptToJson.java
```diff
@@ -0,0 +1,227 @@
...
+                obj.put("answer", (double)answerValue.getValue());
+                return;
+            case Constants.DATATYPE_DATE:
+                obj.put("answer", (DateUtils.formatDate((Date)answerValue.getValue(), DateUtils.FORMAT_ISO8601)));
+                return;
+            case Constants.DATATYPE_TIME:
+                obj.put("answer", answerValue.getDisplayText());
+                return;
+            case Constants.DATATYPE_DATE_TIME:
+                obj.put("answer", ((Date)answerValue.getValue()).getTime()/1000);
```
phillipm: Is there a reason you are doing the `/ 1000`?

api/org/commcare/api/json/PromptToJson.java
```diff
@@ -0,0 +1,227 @@
...
+                return;
+            case Constants.DATATYPE_CHOICE:
+                Selection singleSelection = ((Selection)answerValue.getValue());
+                singleSelection.attachChoice(prompt.getQuestion());
+                obj.put("answer", ((Selection)answerValue.getValue()).index + 1);
```
phillipm: Can you internalize the `((Selection)answerValue.getValue()).index + 1` into the `Selection` class so that if we change how indexing works we don't have to change the `+ 1` here?
Something like adding `Selection.getAnswerIndex()`

api/org/commcare/api/json/PromptToJson.java
```diff
@@ -0,0 +1,227 @@
...
+                for(Selection selection: selections){
+                    selection.attachChoice(prompt.getQuestion());
+                    int ordinal = selection.index + 1;
+                    acc.put(ordinal);
+                }
+                obj.put("answer", acc);
+                return;
+            case Constants.DATATYPE_GEOPOINT:
+                double[] coords = ((GeoPointData)prompt.getAnswerValue()).getValue();
+                obj.put("answer", Arrays.copyOfRange(coords, 0, 2));
```
phillipm: Might be more robust to access this data via `GeoPointData.getLatitude()` and `GeoPointData.getLongitude()`, given that if the internal representation changes we don't need to update this code.

phillipm:
Is the `tests/resources/basic-test.ccz` file that was added used by the form player repo's tests? If so, would it be possible to move `basic-test.ccz` there?

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+            Vector<Selection> ret = new Vector<>();
+            for (String s: split){
+                int index = Integer.parseInt(s);
+                Selection ans = formEntryPrompt.getSelectChoices().get(index -1).selection();
+                ret.add(ans);
+            }
+            return new SelectMultiData(ret);
+        } else if (formEntryPrompt.getDataType() == Constants.DATATYPE_GEOPOINT){
+            return AnswerDataFactory.template(formEntryPrompt.getControlType(), formEntryPrompt.getDataType()).cast(
+                    new UncastData(data.replace(",", " ").replace("[","").replace("]", "")));
```
phillipm: `data.replace(",", " ").replace("[","").replace("]", "")` can you somehow lift this out to an intermediate var or helper method so that you can provide a description as to why this string manipulation is needed?

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        return data.equals("") ? null : AnswerDataFactory.template(formEntryPrompt.getControlType(), formEntryPrompt.getDataType()).cast(new UncastData(data));
```
phillipm: We should lift the `AnswerDataFactory.template(formEntryPrompt.getControlType(), formEntryPrompt.getDataType()).cast(new UncastData(data))` pattern into a static helper; I see this incantation is used all over the code base. I'm also guilty of copying this pattern w/o refactoring it: Let's keep it in mind for a a future PR

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    /**
+     * OK, this function is kind of a monster. Given a FormDef and a String representation of the form index,
+     * return a full fledged FormIndex object.
```
phillipm: :dragon: 
Do you have any tests around this?

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+        } else if(stringIndex.equals(">")){
+            return FormIndex.createEndOfFormIndex();
+        }
+
+        List<Pair<Integer, Integer>> list = stepToList(stringIndex);
+
+        FormIndex ret = reduceFormIndex(list, null);
+        ret.assignRefs(form);
+        return ret;
+    }
```
phillipm: Great code org / helper method ordering on this method :+1:

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        return data.equals("") ? null : AnswerDataFactory.template(formEntryPrompt.getControlType(), formEntryPrompt.getDataType()).cast(new UncastData(data));
+    }
+
+    /**
+     * OK, this function is kind of a monster. Given a FormDef and a String representation of the form index,
+     * return a full fledged FormIndex object.
+     */
+    private static FormIndex indexFromString(String stringIndex, FormDef form){
```
phillipm: This is pretty epic! I considered doing something like this for form entry session replaying feature and got scared of the prospects of doing this string to FormIndex conversion :)

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+        return ret;
+    }
+
+    /**
+     * Given the string representation of one "Step" in a form index, return an Integer pair of <index, multiplicity>
+     */
+    private static Pair<Integer, Integer> stepFromString(String step){
+        // honestly not sure. thanks obama/drew
+        if(step.endsWith("J")){
+            return new Pair<>(Integer.parseInt("" + step.substring(0, step.length()-1)), -10);
```
phillipm: Guessing that `-10` should really be `TreeReference.INDEX_REPEAT_JUNCTURE`?

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+        }
+        return ret;
+    }
+
+    /**
+     * Given the string representation of one "Step" in a form index, return an Integer pair of <index, multiplicity>
+     */
+    private static Pair<Integer, Integer> stepFromString(String step){
+        // honestly not sure. thanks obama/drew
+        if(step.endsWith("J")){
```
phillipm: lol

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+        // we want to deal with both '.' and '_' as separators for the time being for TF legacy reasons
+        String[] split = step.split("[._:]");
+        // the form index is the first part, the multiplicity is the second
+        int i = Integer.parseInt(split[0].trim());
+        int mult = -1;
+        try{
+            mult = Integer.parseInt(split[1].trim());
+        } catch(IndexOutOfBoundsException | NullPointerException e){
+            // do nothing, leave mult as -1
+        }
```
phillipm: I prefer an `if (split.length > 1 && split[1] != null) {...` instead of a `try` / `catch`

phillipm:
Done w/ a pass that focused mostly on smaller details. @ctsims might be better for a higher-level design pass since I assume you've synced a bit w/ him about the form player architecture. Also would be happy to dive into that but would want to have some tests I could spin up to play around with.

wpride:
tests will fail until https://github.com/dimagi/javarosa/pull/282 is merged

phillipm:
Hey Will, if you add
```
cross-request: https://github.com/dimagi/javarosa/pull/282
```
to this PRs description, then the tests will take that branch cross-request into account

wpride:
@phillipm very sexy, thanks

api/org/commcare/api/json/JsonActionUtils.java
wpride: added

wpride:
thanks for the review @phillipm. @ctsims think we're just blocking on a pass from you now

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,270 @@
...
+    public static JSONArray walkToJSON(FormEntryModel fem, FormEntryController fec) {
+        try {
+            JSONArray ret = new JSONArray();
+            FormIndex formIndex = FormIndex.createBeginningOfFormIndex();
+            Walker walker = new Walker(ret, formIndex, fec, fem);
+            walker.walk();
+            return ret;
+        } catch (Exception e) {
+            e.printStackTrace();
+            return null;
```
ctsims: There are tons of things that could have gone wrong in this call, and returning null from here will both crash or break other code, and has no clear semantic meaning. You need to figure out what can be failing here and make sure that it is handled

api/org/commcare/api/json/PromptToJson.java
```diff
@@ -0,0 +1,228 @@
...
+        parseCaption(prompt, questionJson);
+        questionJson.put("help", jsonNullIfNull(prompt.getHelpText()));
+        questionJson.put("binding", jsonNullIfNull(prompt.getQuestion().getBind().getReference().toString()));
+        questionJson.put("style", jsonNullIfNull(parseStyle(prompt)));
+        questionJson.put("datatype", jsonNullIfNull(parseControlType(prompt)));
+        questionJson.put("required", prompt.isRequired() ? 1 : 0);
+        try {
+            parseQuestionAnswer(questionJson, prompt);
+        } catch (Exception e) {
+            e.printStackTrace();
```
ctsims: What are the failures here and why is it fine to swallow them

ctsims:
@wpride Please clean up error handling, will review afterwards.

api/org/commcare/api/json/JsonActionUtils.java
wpride: @ctsims thanks, think I added this for testing and forgot to remove. Exceptions should just bubble, thanks

wpride:
@ctsims error handling fixed, thanks.

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+     * @param model           the FormEntryModel under consideration
+     * @param formIndexString the form index of the repeat group to be deleted
+     * @return The JSON representation of the updated form tree
+     */
+    public static JSONObject deleteRepeatToJson(FormEntryController controller,
+                                                FormEntryModel model, String formIndexString) {
+        JSONObject ret = new JSONObject();
+        FormIndex formIndex = indexFromString(formIndexString, model.getForm());
+        controller.deleteRepeat(formIndex);
+        ret.put(ApiConstants.QUESTION_TREE_KEY, walkToJSON(model, controller));
```
phillipm: Could return result of `getCurrentJson` to avoid repeating this line (same for `descendRepeatToJson`)

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+     * Return the IAnswerData version of the string data input
+     *
+     * @param formEntryPrompt the FormEntryPrompt for this question
+     * @param data            the String answer
+     * @return the IAnswerData version of @data above
+     */
+    public static IAnswerData getAnswerData(FormEntryPrompt formEntryPrompt, String data) {
+
+        if (formEntryPrompt.getDataType() == Constants.DATATYPE_CHOICE) {
+            int index = Integer.parseInt(data);
```
phillipm: Throws a `NumberFormatException` runtime exception that might be worth catching

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Return the IAnswerData version of the string data input
+     *
+     * @param formEntryPrompt the FormEntryPrompt for this question
+     * @param data            the String answer
+     * @return the IAnswerData version of @data above
+     */
+    public static IAnswerData getAnswerData(FormEntryPrompt formEntryPrompt, String data) {
+
+        if (formEntryPrompt.getDataType() == Constants.DATATYPE_CHOICE) {
```
phillipm: This series of if/else's might look a bit clearer as a switch statement

api/org/commcare/api/json/JsonActionUtils.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    /**
+     * Return the IAnswerData version of the string data input
+     *
+     * @param formEntryPrompt the FormEntryPrompt for this question
+     * @param data            the String answer
+     * @return the IAnswerData version of @data above
+     */
+    public static IAnswerData getAnswerData(FormEntryPrompt formEntryPrompt, String data) {
```
phillipm: This does some specific conversions from Touchforms format to xforms formats. Do you have tests that make sure these conversions are handled correctly?

api/org/commcare/api/json/PromptToJson.java
```diff
@@ -0,0 +1,224 @@
...
+        return ret;
+    }
+
+
+    private static String parseControlType(FormEntryPrompt prompt) {
+        if (prompt.getControlType() == Constants.CONTROL_TRIGGER) {
+            return "info";
+        }
+        switch (prompt.getDataType()) {
+            case Constants.DATATYPE_NULL:
```
phillipm: Can you add a comment describing why this switch case is falling through to the next? This is a common gotcha, so it is worth explicitly asserting this is what we want.

api/org/commcare/api/json/JsonActionUtils.java
wpride: :mind blown:

api/org/commcare/api/json/JsonActionUtils.java
wpride: Will be handled here https://github.com/dimagi/commcare/blob/c6cb68176405a95fbcc1100a84b461fbb7486812/api/org/commcare/api/json/JsonActionUtils.java#L88-L88

api/org/commcare/api/json/JsonActionUtils.java
wpride: The code is covered here:

https://github.com/dimagi/formplayer/blob/6ef454ccfe40def7227ec2828bb32d22a4927f7a/src/test/java/tests/FormEntryTest.java#L57-L57

Though I'd accept an argument that this is too sparse

api/org/commcare/api/json/PromptToJson.java
wpride: (added)

wpride:
@ctsims would be good to get this reviewed before 2.28

wpride:
@phillipm Merged. Would be good to sanity check the build.gradle file since that had the most conflicts

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -64,7 +64,7 @@
     /**
      * The current session frame data
      */
-    private SessionFrame frame;
+    public SessionFrame frame;
```
phillipm: Can you revert this change and access `frame` via the existing `getFrame` method?

PR #275, opened 2016-03-23
Notify user of self-indexing case in restore payload

phillipm:
cross-request: https://github.com/dimagi/javarosa/pull/278

### DIFF
--- a/backend/src/org/commcare/xml/CaseXmlParser.java
+++ b/backend/src/org/commcare/xml/CaseXmlParser.java
@@ -184,8 +184,8 @@ public class CaseXmlParser extends TransactionParser<Case> {
 
                     String value = parser.nextText().trim();
 
-                    if(value.equals(caseId)) {
-                        throw new InvalidStructureException("Invalid index operation! Case " + caseId + " cannot index itself!", parser);
+                    if (value.equals(caseId)) {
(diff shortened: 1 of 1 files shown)


backend/src/org/commcare/xml/CaseXmlParser.java
```diff
@@ -184,8 +186,8 @@ public Case parse() throws InvalidStructureException, IOException, XmlPullParser
 
                     String value = parser.nextText().trim();
 
-                    if(value.equals(caseId)) {
-                        throw new InvalidStructureException("Invalid index operation! Case " + caseId + " cannot index itself!", parser);
+                    if (value.equals(caseId)) {
+                        throw new ActionableInvalidStructureException("case.error.self.index", new String[]{caseId}, "Case " + caseId + " cannot index itself");
```
ctsims: It looks like this mesage is only localized in the android repo. We should add back to the JR repos as well, right? Or were you thinking we would just use the fallback text for that? (Also fine)

phillipm: The localization is performed at the javarosa level. The key only exists at the android level. I've updated the javarosa localization to return the default message if the localization key isn't found.

Where would we add the localization key to the J2ME code-base?

PR #279, opened 2016-04-05
Add copy constructor to session classes

phillipm:
for use in https://github.com/dimagi/commcare-odk/pull/1187

cross-request: https://github.com/dimagi/commcare-odk/pull/1187


wpride:
@phillipm So we don't mind that these are shallow copies? (Though agree that this is an improvement on previous either way)

phillipm:
I don't think these are shallow copy constructors, are they? Other than immutable objects like strings.

PR #280, opened 2016-04-05
Refactor: getNeededData

phillipm:
Added comments, helper methods, and did some variable renaming to try to make getNeededData a little clearer.

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -160,7 +160,6 @@ public class CommCareSession {
         String needDatum = null;
         String nextKey = null;
         for (Entry e : possibleEntries) {
-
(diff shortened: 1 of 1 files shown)


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -144,59 +144,60 @@ public CommCarePlatform getPlatform() {
...
+        }
+    }
 
+    /**
+     * Checks that all entries have the same id for their first required data,
+     * and if so, returns the data's associated session state. Otherwise,
+     * returns null.
+     */
+    private String getDataNeededByAllEntries(Vector<Entry> entries) {
         String needDatum = null;
```
amstone326: I think renaming these to something like `datumNeededByAllEntriesSoFar` (and `idOfDatumNeededByAllEntriesSoFar`) would increase the readability of this by a lot

phillipm:
@amstone326 : updated, but left the id var name the same because it seemed too long

PR #282, opened 2016-04-06
Represent datum data using OO

phillipm:
cross-request: https://github.com/dimagi/commcare-odk/pull/1189


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -160,15 +186,14 @@ public String getNeededData() {
...
             SessionDatum datumNeededForThisEntry = getFirstMissingDatum(this.getData(), e.getSessionDataReqs());
             if (datumNeededForThisEntry != null) {
                 String needed = datumNeededForThisEntry.getDataId();
                 if (nextKey == null) {
                     nextKey = needed;
-                    if (datumNeededForThisEntry.getNodeset() != null) {
+                    if (datumNeededForThisEntry instanceof EntityDatum) {
                         needDatum = SessionFrame.STATE_DATUM_VAL;
-                    } else {
+                    } else if (datumNeededForThisEntry instanceof ComputedDatum) {
```
phillipm: Depending on how `FormIdDatum`s are used, adding this `if` to the `else` might be invalid

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -343,10 +368,10 @@ public void setComputedDatum(EvaluationContext ec) throws XPathException {
             e.printStackTrace();
             throw new RuntimeException(e.getMessage());
         }
-        if (datum.getType() == SessionDatum.DATUM_TYPE_FORM) {
+        if (datum instanceof FormIdDatum) {
             setXmlns(XPathFuncExpr.toString(form.eval(ec)));
             setDatum("", "awful");
-        } else {
+        } else if (datum instanceof ComputedDatum) {
```
phillipm: Same as above regarding `FormIdDatum`

backend/src/org/commcare/session/SessionNavigator.java
```diff
@@ -95,36 +96,38 @@ private void handleGetDatum() {
...
-            // No confirm detail defined for this entity select, so just set the case id right away
-            // and proceed
-            String autoSelectedCaseId = SessionDatum.getCaseIdFromReference(
-                    currentAutoSelectedCase, selectDatum, ec);
-            currentSession.setDatum(selectDatum.getDataId(), autoSelectedCaseId);
-            startNextSessionStep();
-        } else {
-            sendResponse(LAUNCH_CONFIRM_DETAIL);
+        if (selectDatum instanceof EntityDatum) {
+            EntityDatum entityDatum = (EntityDatum)selectDatum;
```
phillipm: The check and cast to `EntityDatum` are unfortunate, but at least it is nice that we are being explicit that this behavior is only supported for case selections and not other datums.

phillipm:
@ctsims wanted to double check that changing the serialization scheme of `SessionDatum` won't require any sort of data migration. Sound right?

backend/src/org/commcare/suite/model/ComputedDatum.java
```diff
@@ -0,0 +1,20 @@
+package org.commcare.suite.model;
+
+/**
+ * Piece of required session data that is acquired via an xpath computation
+ *
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public class ComputedDatum extends SessionDatum {
+    public ComputedDatum() {
```
wpride: Is this needed for serialization?

phillipm: Yeah, I'll add a comment stating so

backend/src/org/commcare/suite/model/EntityDatum.java
```diff
@@ -0,0 +1,149 @@
...
+public class EntityDatum extends SessionDatum {
+    private TreeReference nodeset;
+    private String shortDetail;
+    private String longDetail;
+    private String inlineDetail;
+    private String persistentDetail;
+    private boolean autoSelectEnabled;
+
+    public EntityDatum() {
+
```
wpride: comment if needed for serialization?

wpride:
@phillipm has conflicts

backend/src/org/commcare/session/SessionNavigator.java
wpride: What's the failure behavior here? IE when EntityDatum is not a selectDatum

wpride: nvm I see this is appropriate from the comment on getAutoSlelectedCase

backend/src/org/commcare/session/SessionNavigator.java
```diff
@@ -95,36 +96,38 @@ private void handleGetDatum() {
...
         }
     }
 
     /**
-     *
      * Returns the auto-selected case for the next needed datum, if there should be one.
      * Returns null if auto selection is not enabled, or if there are multiple available cases
      * for the datum (and therefore auto-selection should not be used).
      */
     private TreeReference getAutoSelectedCase() {
```
wpride: Not your code, but the whole handleGetDatum() autoSelect workflow seems janky. We get the autoSelectedCase, then  put the result in a class variable. then call handleAutoSelect and do the instanceof/cast with selectDatum both times. Maybe belongs in a separate refactor but could move everything besides the if (autoSelection == null) case inside an if (selectDatum instanceof EntityDatum) cast passing through the casted EntityDatum and TreeReference as needed

phillipm: Yeah, I agree that the autoselect code makes less sense after my refactor.  I'll make a note to follow up with a future refactor PR

PR #284, opened 2016-04-07
Break profile parser up into reasonable methods

amstone326:


backend/src/org/commcare/xml/ProfileParser.java
```diff
@@ -109,113 +149,78 @@ public Profile parse() throws InvalidStructureException, IOException, XmlPullPar
...
-                    // we shouldn't ever get this I don't believe, maybe on the
-                    // last node?
-                } else if (eventType == KXmlParser.TEXT) {
-                    // Shouldn't ever get this (Delete the if, if so).
                 }
-                eventType = parser.next();
-            } while (eventType != KXmlParser.END_DOCUMENT);
-
-            return profile;
+            } else if (tag.equals("sense")) {
```
wpride: Can we remove this?

amstone326: I refrained from doing so in case that was in there to remind someone to implement something at some point.. very happy for it to be gone if that's not the case.

PR #293, opened 2016-04-22
Return true on isView for some FormEntry instances

phillipm:
Fix the assumption that `<entry>` is never a view, which I made in https://github.com/dimagi/commcare/pull/242

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -1,8 +1,8 @@
 package org.commcare.session;
 
 import org.commcare.suite.model.Detail;
-import org.commcare.suite.model.FormEntry;
 import org.commcare.suite.model.Entry;
+import org.commcare.suite.model.FormEntry;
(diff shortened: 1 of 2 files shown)


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -784,7 +784,10 @@ private void markCurrentFrameForDeath() {
      */
     public boolean isViewCommand(String command) {
         Vector<Entry> entries = this.getEntriesForCommand(command);
-        return entries.elementAt(0).isView();
+        Entry entry = entries.elementAt(0);
+        return entries.size() == 1
+                && entry.isView()
+                && entry.getPostEntrySessionOperations().size() == 0;
```
phillipm: Add back `entry.getPostEntrySessionOperations().size() == 0` which I should have never taken out at https://github.com/dimagi/commcare/pull/242/files#diff-256562e5e1473541dfae8ff8b3f5bfcbL788

PR #296, opened 2016-04-28
Consumer apps config changes

amstone326:
cross-request: https://github.com/dimagi/commcare-odk/pull/1230


backend/src/org/commcare/resources/ResourceManager.java
```diff
@@ -136,8 +136,17 @@ protected void loadProfileIntoTable(ResourceTable table,
             throws UnfullfilledRequirementsException,
```
phillipm: Instead of relying on name overloading, can you change this method's name to include the fact that a remote authority is being?

PR #297, opened 2016-05-03
Handle an XPathException in SessionNavigator properly

amstone326:
cross-request: https://github.com/dimagi/commcare-odk/pull/1240 (see description there)


backend/src/org/commcare/session/SessionNavigator.java
```diff
@@ -135,7 +135,8 @@ private void handleCompute() {
             currentSession.setComputedDatum(ec);
         } catch (XPathException e) {
             this.thrownException = e;
-            sendResponse(EXCEPTION_THROWN);
+            sendResponse(XPATH_EXCEPTION_THROWN);
+            return;
```
amstone326: The absence of this return statement is what caused the `XPathException` that was being thrown to manifest itself as an infinite loop and eventual force close.

PR #302, opened 2016-06-01
Add relevancy condition to post request

phillipm:
Add `relevant="some_xpath_expr"` to the `post` tag of a sync request. The relevancy condition is checked before making the post request. This allows mobile to check if the case is already owned by the user before claiming it on HQ. Doing this check on the server side is much more difficult.

cross-request: https://github.com/dimagi/commcare-odk/pull/1288

example:
```
 <sync-request>
    <post url="https://www.fake.com/claim_patient/"
          relevant="count(instance('casedb')/casedb/case[@case_id=instance('session')/session/data/case_id]) = 0">
      <data key="selected_case_id" ref="instance('session')/session/data/case_id"/>
    </post>
    <command id="patient-search">
      <display>
        <text>Global search for person</text>
      </display>
    </command>
    <instance id="session" src="jr://instance/session"/>
    <session>
      <query url="https://www.fake.com/patient_search/" storage-instance="patients">
        <data key="device_id" ref="instance('session')/session/context/deviceid"/>
        <prompt key="name">
          <display>
            <text>
              <locale id="query.name"/>
            </text>
          </display>
        </prompt>
        <prompt key="patient_id">
          <display>
            <text>
              <locale id="query.id"/>
            </text>
          </display>
        </prompt>
      </query>
      <datum id="case_id" nodeset="instance('patients')/patients/patient" value="./id" detail-select="patient_short" detail-confirm="patient_long"/>
    </session>
    <stack>
      <create>
        <command value="'m1'"/>
        <datum id="case_id" value="instance('session')/session/data/case_id"/>
      </create>
    </stack>
  </sync-request>
```

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -463,7 +463,7 @@ public class CommCareSession {
         }
 
         Entry e = platform.getMenuMap().get(command);
-        if (e.isView()) {
+        if (e.isView() || e.isSync()) {
             return null;
         } else {
             return ((FormEntry)e).getXFormNamespace();
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -81,7 +81,7 @@ public class CommCareSession {
 
     public CommCareSession(CommCarePlatform platform) {
         this.platform = platform;
-        collectedDatums = new OrderedHashtable<String, String>();
+        this.collectedDatums = new OrderedHashtable<String, String>();
         this.frame = new SessionFrame();
         this.frameStack = new Stack<SessionFrame>();
     }
@@ -189,7 +189,7 @@ public class CommCareSession {
      * @return One of the session SessionFrame.STATE_* strings, or null if
      * the session does not need anything else to proceed
      */
-    public String getNeededData() {
+    public String getNeededData(EvaluationContext evalContext) {
         if (currentCmd == null) {
             return SessionFrame.STATE_COMMAND_ID;
         }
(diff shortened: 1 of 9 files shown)


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -199,7 +199,9 @@ public String getNeededData() {
 
         if (needDatum != null) {
             return needDatum;
-        } else if (entries.size() == 1 && entries.elementAt(0) instanceof SyncEntry) {
+        } else if (entries.size() == 1
+                && entries.elementAt(0) instanceof SyncEntry
+                && ((SyncEntry)entries.elementAt(0)).getSyncPost().isRelevant(evalContext)) {
```
wpride: any chance `getSyncPost()` will return null?

phillipm: Good call, added https://github.com/dimagi/commcare/pull/302/commits/28b607158cb72ae8ff480cebcfe35a568b732fe4 to error out if the `sync-request` suite xml entry doesn't have a `<post>` element

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -356,14 +358,14 @@ public Suite getCurrentSuite() {
         return null;
     }
 
-    public void stepBack() {
```
wpride: Probably a separate PR/idea but all this evalContext passing around the CommCareSession feels a little silly, is there any reason not to just pair the evalContext with the Session at this point? (as seems to be the case in the SessionWrappers)? Does the evalContext ever "diverge" from the CommCareSession such that commCareSession having a getter would return the wrong thing?

backend/src/org/commcare/xml/EntryParser.java
```diff
@@ -149,6 +153,18 @@ private void parseInstance(Hashtable<String, DataInstance> instances) {
 
     private SyncPost parsePost() throws InvalidStructureException, IOException, XmlPullParserException {
         String url = parser.getAttributeValue(null, "url");
+
```
wpride: not in this PR, but line 136 above is the only one still using `this.parser` instead of `parser`

backend/src/org/commcare/session/CommCareSession.java
phillipm: Yeah, I hope to revisit this and clean it up. I tried to go that route but they are pretty deeply interdependent.

PR #306, opened 2016-06-08
Installation and update performance improvements

phillipm:
**Release Notes**: Install and update performance enhancements 

Speed-ups:
- From 12.5 minutes to 2.5 minutes for offline install of ICDS AWW app 
- From 8 mins to 1.5 for downloading an update
- From 12 mins to 2 for applying the downloaded update

Additional improvement: the install/update progress bar updates much more frequently 

Changes:
- Don't use intermediate vector when iterating over resources read from storage
- Don't recalculate table update/install progress every time a new resource is added. Only do it when a 'compound' resource is added. 'Compound' meaning a resource that points to other resources, such as a profile, suite, or media suite resource. 
- Store compound resources in a cache instead of loading them from storage. There should only ever be 3 of these, but they are accessed from storage pretty frequently right now.
- Lift storage look-ups out of loop bodies. Instead of looking up resources by ID from storage, load them all into an intermediate table that is passed around. I don't expect a table with 2k `Resource` objects to be an issue on older devices, but we should test that assumption.
- Unify `flagForDeletion` and `uninstall`, which both iterate over the same resource set, happen one right after the other, and can be lifted into the same loop, avoiding a pass of storage rights.

cross-request: https://github.com/dimagi/commcare-odk/pull/1309

### DIFF
--- a/backend/src/org/commcare/resources/ResourceManager.java
+++ b/backend/src/org/commcare/resources/ResourceManager.java
@@ -233,8 +233,7 @@ public class ResourceManager {
                 upgradeTable.destroy();
 
                 Logger.log("Resource", "Clearing out old resources");
-                tempTable.flagForDeletions(masterTable);
-                tempTable.completeUninstall();
+                tempTable.uninstall(masterTable);
             } finally {
                 if (!upgradeSuccess) {
                     repair();
--- a/backend/src/org/commcare/resources/model/ResourceTable.java
+++ b/backend/src/org/commcare/resources/model/ResourceTable.java
@@ -608,7 +608,9 @@ public class ResourceTable {
         // Upgrade elements should result in their counterpart in this table
         // being unstaged (which can be reverted).
 
-        for (Resource r : incoming.getResources()) {
+
+        for (IStorageIterator it = incoming.storage.iterate(); it.hasMore(); ) {
+            Resource r = (Resource)it.nextRecord();
             Resource peer = getResourceWithId(r.getResourceId());
             if (peer == null) {
                 // no corresponding resource in this table; use incoming
@@ -667,7 +669,8 @@ public class ResourceTable {
      */
     public void uninstall(ResourceTable replacement) {
         cleanup();
-        for (Resource r : getResources()) {
+        for (IStorageIterator it = storage.iterate(); it.hasMore(); ) {
+            Resource r = (Resource)it.nextRecord();
             if (replacement.getResourceWithId(r.getResourceId()) == null ||
                     r.getStatus() == Resource.RESOURCE_STATUS_UNSTAGED) {
                 // No entry in 'replacement' so it's no longer relevant
@@ -740,7 +743,8 @@ public class ResourceTable {
         }
 
         // Copy over all of our resources to the new table
-        for (Resource r : this.getResources()) {
+        for (IStorageIterator it = storage.iterate(); it.hasMore(); ) {
+            Resource r = (Resource)it.nextRecord();
             r.setID(-1);
             newTable.commit(r);
         }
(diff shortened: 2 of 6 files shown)


backend/src/org/commcare/resources/model/ResourceTable.java
```diff
@@ -617,27 +695,47 @@ public boolean upgradeTable(ResourceTable incoming) throws UnresolvedResourceExc
...
+                            r.getRecordGuid() + ". " + e.getMessage());
+                }
+            } else if (r.getStatus() == Resource.RESOURCE_STATUS_DELETE) {
+                // NOTE: Shouldn't be a way for this condition to occur, but check anyways...
+                try {
+                    r.getInstaller().uninstall(r);
+                } catch (Exception e) {
+                    Logger.log("Resource", "Error uninstalling resource " +
+                            r.getRecordGuid() + ". " + e.getMessage());
+                }
```
phillipm: Thoughts on this conditional / `NOTE` I've made?

backend/src/org/commcare/resources/model/ResourceTable.java
```diff
@@ -443,33 +501,50 @@ public void prepareResourcesUpTo(ResourceTable master,
...
                 if (!r.isNewer(peer)) {
                     // This resource doesn't need to be updated, copy
                     // the existing resource into this table
                     peer.mimick(r);
                     commit(peer, Resource.RESOURCE_STATUS_INSTALLED);
+
+                    if (stateListener != null) {
+                        // copying a resource over shouldn't add anymore
+                        // resources to be processed
+                        stateListener.simpleResourceAdded();
```
phillipm: Does my comment sound like a valid assumption?

backend/src/org/commcare/resources/ResourceManager.java
```diff
@@ -233,8 +233,7 @@ public void upgrade()
                 upgradeTable.destroy();
 
                 Logger.log("Resource", "Clearing out old resources");
-                tempTable.flagForDeletions(masterTable);
-                tempTable.completeUninstall();
```
phillipm: This unification should be double checked.

backend/src/org/commcare/resources/model/ResourceTable.java
phillipm: Thoughts on this conditional / `NOTE` I've made?
In this change I've removed the only place where a resource is ever marked `RESOURCE_STATUS_DELETE`. This is now the only place where `RESOURCE_STATUS_DELETE` is referenced; and it probably isn't even needed now. 

The `upgrade()` logic is written with fault tolerance in mind, so I'm trying to figure out how `RESOURCE_STATUS_DELETE` played into that. Currently not seeing how it does. Specifically, `RESOURCE_STATUS_DELETE` is only used after the upgrade is considered complete, at which point `repair()` won't be called, which is the only other place that `tempTable` is used. Hence we don't need to be particularly careful with `tempTable` getting cleared correctly because it will get `destroy()`ed before `upgrade()` is entered again.
Long, rambly way of saying this change is fine.

phillipm:
@ctsims Can you review this please?

backend/src/org/commcare/resources/model/ResourceTable.java
ctsims: Yes, that looks right.

backend/src/org/commcare/resources/model/ResourceTable.java
ctsims: Were there any places where the temp table "Clearing" could be "picked up" again in a state where the 

tempTable.completeUninstall();

Was called, but the flag for delete step was not? If not, I think you're right that we've already accepted the upgrade, so I can't find a way in my mind where this would need to be a two-step process

PR #315, opened 2016-06-27
Don't use PrefixTree in Localizer

phillipm:
Remove use of `PrefixTree` in language localization code.
Replace J2ME compatible `replaceAll` and `lastIndexOf` with stock `String` versions.
Use `BufferedReader` instead of rolling our own line reader.

Speeds up full ICDS app init by ~16 seconds from 38 to 22 seconds.

cross-request: https://github.com/dimagi/commcare-android/pull/1335


javarosa/src/main/java/org/javarosa/core/services/locale/Localizer.java
```diff
@@ -730,4 +584,54 @@ public void writeExternal(DataOutputStream dos) throws IOException {
...
+    }
+
+    /**
+     * For Testing: Get default form fallback mode
+     *
+     * @return default form fallback mode
+     */
+    public boolean getFallbackForm() {
+        return fallbackDefaultForm;
+    }
```
phillipm: This diff block is just copying code, no changes were made

ctsims:
Out of curiosity, is it clear whether removing the prefix tree, or using the string optimizations was the performance gain? 

Additionally for these refactors, it would be useful to see their impact on memory footprint as well as run-time. Most of them weren't implemented to speed things up, and if we're reducing the size of forms we can run that doesn't seem like a win.

phillipm:
Looking at the memory monitor at the login screen and after changing a language (after running a GC), the footprints are nearly identical, with the new `Hashtable` implementation being slightly smaller.

The speed-ups come mostly from removing the `PrefixTree` from the localize logic, with slight improvements from using a stock file reader.

Interestingly, I was a little sloppy with my changes yesterday: the custom J2ME functions that @wspride implemented are faster than the stock ones I tried using. The revert sped things up by an additional few seconds

PR #316, opened 2016-06-27
Remove PrefixTree

phillipm:
Removing `PrefixTree` from use in `PrototypeFactory` shaves ~.75 seconds of start-up time for a small CC app on a Nexus 5x. Taking the time from 2.7 seconds to 1.9 seconds (I'm doing super rough timing estimates BTW). I expect that would scale nicely to older phones. Memory footprint seems to be the same.

cross-request: https://github.com/dimagi/commcare-android/pull/1336


javarosa/src/main/java/org/javarosa/core/services/locale/Localizer.java
```diff
@@ -47,17 +44,14 @@ public Localizer() {
...
-     * Full constructor.
-     *
      * @param fallbackDefaultLocale If true, search the default locale when no translation for a particular text handle
      *                              is found in the current locale.
      * @param fallbackDefaultForm   If true, search the default text form when no translation is available for the
      *                              specified text form ('long', 'short', etc.). Note: form is specified by appending ';[form]' onto the text ID.
      */
     public Localizer(boolean fallbackDefaultLocale, boolean fallbackDefaultForm) {
-        stringTree = new PrefixTree(10);
-        localeResources = new OrderedHashtable<>();
```
ctsims: Is this (Ordered -> Non Ordered) vital for performance? I don't remember why these are ordered, but it seems like an odd detail to exist and may have relevance.

phillipm: Checking how `localeResources` is used, it should only improve performance. That is, we never iterate over `localeResources` elements, so there is no reason to order it.

PR #317, opened 2016-06-27
Remove 'Map' a custom hashtable implementation

phillipm:
Our serialization code used a custom implementation of hashtables in certain places where it might save memory. I'm removing that under the assumption that removing it will make code maintenance easier and potentially make things faster, since hand-rolled implementations are rarely faster than core libs. 

I'm not going to profile the performance of these changes because the `Map` implementation is only used in ~4 places that seem to not be critical code paths.


orangejenny:
what a glorious day

PR #318, opened 2016-06-27
Remove unused interning infrastructure

phillipm:
There is a lot of interning infrastructure that is currently unused.
Instead of investing to see if interning makes things on Android more efficient, I'm electing to remove the code for now. We can revisit this when we establish it is needed.

Interning removed from:
- creation of `String` tree element name keys in `CaseInstanceTreeElement` and `LedgerInstanceTreeElement`
- creation of `TreeReferenceLevel` instances
- creation of `String` instances in `ExtUtil` serialization code
- creation of `String` instances in `XFormParser`
- creation of `XPathStep` instances in `XPathPathExpr`


phillipm:
@ctsims Are you fine with the decision of removing unused interning code?

ctsims:
@phillipm What's the thinking behind it being unused? Interning provides a significant decrease in our memory footprint if nothing else, along with significantly faster equality checks

phillipm:
@ctsims All the code paths that register interning caches come from J2ME. In Android interning is not currently utilized.

amstone326:
@ctsims Are you ok with this based on what Phillip said?

ctsims:
@amstone326 @phillipm I'm comfortable removing it although I think we may be better of by investing in doing the interning for the strings at least.  A _huge_ percentage of our runtime comes from comparing strings while doing equality checks. Interning common strings makes those kinds of checks _significantly_ faster.

phillipm:
Since it is dead code that hasn't been enabled for at least a year or more, my vote is to remove it. When we decide we want to resource more performance work we can do a comprehensive analysis of all the places where interning yield benefits.

ctsims:
@phillipm Up to you. I'm comfortable with both, just wanted to confirm that we hadn't tried, for instance, turning it on and seeing if the performance was worth investing in, given that the infrastructure is written and would likely be quick to enable.

Let me know if you are signing off on killing it, and I'll pull (or Aliza can if she reads first).

amstone326:
If we think it's worth looking into while the code is still here, I'd be up for taking a pass at it; my load is kind of light for the next couple of days till the sprint. If we think it could have a lot of benefit, it kind of seems like a waste to delete all of the existing infrastructure and then have to reimplement it.

phillipm:
Closing this because leaving the interning code, which is pretty contained, in until we get a chance to check its efficacy is fine.

PR #321, opened 2016-07-01
Union is intersection

phillipm:
Method named `union` performs set intersection

`commcare-core` version of https://github.com/dimagi/commcare-j2me/pull/313 with unit tests and proper `union` -> `intersction` renaming

Also fixes CLI jar build script

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,9 @@ sourceSets {
   cli {
       java {
           srcDir 'util/src'
-          compileClasspath += main.output + harness.output
+          srcDir 'javarosa/util/schema-gen/src'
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -52,7 +52,9 @@ sourceSets {
   cli {
       java {
           srcDir 'util/src'
-          compileClasspath += main.output + harness.output
+          srcDir 'javarosa/util/schema-gen/src'
```
amstone326: what's up here?

phillipm: Gradle isn't being my friend these days and I can't figure out how to reference other sourcesets in a sourceset def. So I just copied them. This is needed to get the CLI building again

PR #322, opened 2016-07-01
actually serialize grid coordinates

wpride:
Don't know why this didn't come up before, but manifested in setting up case tiles on Formplayer


amstone326:
should this be in 2.28?

wpride:
don't think it needs to be, hasn't been seen in the wild

PR #325, opened 2016-07-06
Fix for J2ME Calculated Session Data

ctsims:
Fixes a bug where the new polymorphic session datum types weren't working right on j2me.

Fix for:
http://manage.dimagi.com/default.asp?230701

### DIFF
--- a/application/src/org/commcare/util/CommCareSessionController.java
+++ b/application/src/org/commcare/util/CommCareSessionController.java
@@ -270,7 +270,7 @@ public class CommCareSessionController {
         //TODO: Stuff can be in more than one suite!!!
         Suite suite = session.getCurrentSuite();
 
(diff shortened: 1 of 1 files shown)


ctsims:
@phillipm What is the right process to follow for this to get into new-j2me as well?

phillipm:
To get javarosa j2me changes into 2.28 and any j2me related changes into
the new j2me repo you will need to create a PR in the commcare-j2me repo

ctsims:
@phillipm Cool. This one is a commcare change for 2.28, so I think it still goes here. I'll make a duplicate change in commcare-j2me for future releases on j2me.

ctsims:
@phillipm Great, replicated here: https://github.com/dimagi/commcare-j2me/pull/315

PR #326, opened 2016-07-07
Move case selection predicate to 1st position when loading persistent case tiles

phillipm:
Final fix for http://manage.dimagi.com/default.asp?221062
Reduces 26 second form load to 3.5 seconds

The idea is to put the case id selection predicate first, so it doesn't load all the cases in the db and run all the other predicates over those cases first. Case IDs are indexed in the db, so doing that first will be super fast.

### DIFF
--- a/backend/src/org/commcare/suite/model/EntityDatum.java
+++ b/backend/src/org/commcare/suite/model/EntityDatum.java
@@ -122,7 +122,11 @@ public class EntityDatum extends SessionDatum {
         if (predicates == null) {
             predicates = new Vector<XPathExpression>();
         }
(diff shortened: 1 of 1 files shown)


amstone326:
wow, this is awesome

ctsims:
@phillipm If this is going into 2.28 can we make sure to double check against another app that uses case tiles?

ctsims:
Also, awesome job! this is huge, sorry for the oversight on it in the original design

phillipm:
@ctsims what other apps use case tiles?

ctsims:
ICDS, TDH, and I believe UATBC

PR #330, opened 2016-07-11
Some linting: type and auto-boxing related warning

phillipm:


javarosa/src/main/java/org/javarosa/core/model/data/IntegerData.java
```diff
@@ -50,49 +34,42 @@ public IntegerData(Integer n) {
...
         return DataUtil.integer(n);
     }
 
+    @Override
     public void setValue(Object o) {
-        if (o == null) {
-            throw new NullPointerException("Attempt to set an IAnswerData class to null.");
-        }
-        n = ((Integer)o).intValue();
+        n = (Integer)o;
```
amstone326: will attempting to cast a null `Object` to an `Integer` throw a `NullPointerException` like the old behavior did?

amstone326: ha nevermind, just saw you reverted this change in the next commit

phillipm: I actually reverted this later in the PR because I think it was an incorrect change

phillipm: ah, you saw it

PR #331, opened 2016-07-11
Remove unused locale change and element state change alerts 

phillipm:
cross-request: https://github.com/dimagi/commcare-android/pull/1379


amstone326:
@phillipm Thinking I'm missing the hook that indicated to you this was all unused, would you mind clarifying?

phillipm:
I read [a comment](https://github.com/dimagi/commcare-core/pull/331/files#diff-ab59a1ac6da2e874115efe1bb4a8bc21L210) that said `localeChanged` is `DEPRECATED` and traced through all the method calls. They went in big circles so I deleted everything. It all still seems to work just fine.

`FormElementStateListener` isn't referenced on FormPlayer/CloudCare, but @wpride you will need to remove [these lines](https://github.com/dimagi/formplayer/blob/master/src/main/java/hq/RestoreUtils.java#L74-L77) from the FormPlayer code.

PR #332, opened 2016-07-11
Remove dead code that denotes sorting of triggerables

phillipm:
In celebration of Drew's visit this week figured we'd remove some of his dead code that has been around for 7 years :)

### DIFF
--- a/javarosa/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/javarosa/src/main/java/org/javarosa/core/model/FormDef.java
@@ -90,11 +90,6 @@ public class FormDef implements IFormElement, Persistable, IMetaData,
     // depend on any result from evaluating tB
     private Vector<Triggerable> triggerables;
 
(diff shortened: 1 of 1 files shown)


javarosa/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -90,11 +90,6 @@
     // depend on any result from evaluating tB
     private Vector<Triggerable> triggerables;
 
-    // true if triggerables has been ordered topologically (DON'T DELETE ME
-    // EVEN THOUGH I'M UNUSED)
```
amstone326: I guess we're ignoring this comment? Any idea why he said don't delete?

phillipm: Haha, let's ask him when he visits this week

phillipm: Or we can just assume that if it hasn't been used in the last 7 years it won't be used in the next few

PR #337, opened 2016-07-13
Move case claim code into commcare-core

phillipm:
cross-request: https://github.com/dimagi/commcare-android/pull/1377

### DIFF
--- a/backend/src/org/commcare/suite/model/RemoteQueryDatum.java
+++ b/backend/src/org/commcare/suite/model/RemoteQueryDatum.java
@@ -10,6 +10,8 @@ import org.javarosa.xpath.expr.XPathExpression;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.Hashtable;
 
 /**
@@ -26,10 +28,10 @@ public class RemoteQueryDatum extends SessionDatum {
     public RemoteQueryDatum() {
     }
 
-    public RemoteQueryDatum(String url, String storageInstance,
+    public RemoteQueryDatum(URL url, String storageInstance,
                             Hashtable<String, XPathExpression> hiddenQueryValues,
                             Hashtable<String, DisplayUnit> userQueryPrompts) {
-        super(storageInstance, url);
+        super(storageInstance, url.toString());
         this.hiddenQueryValues = hiddenQueryValues;
         this.userQueryPrompts = userQueryPrompts;
     }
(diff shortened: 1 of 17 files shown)


core/org/commcare/core/encryption/CryptUtil.java
```diff
@@ -0,0 +1,163 @@
+package org.commcare.core.encryption;
```
wpride: Was any of this not a straight copy?

javarosa/src/main/java/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -34,52 +33,21 @@ public static void writeFromInputToOutput(InputStream in, OutputStream out, long
     }
 
     public static void writeFromInputToOutputSpecific(InputStream in, OutputStream out) throws InputIOException, OutputIOException {
-        writeFromInputToOutput(in, out, null);
+        writeFromInputToOutputInner(in, out);
     }
 
     public static void writeFromInputToOutput(InputStream in, OutputStream out) throws IOException {
         try {
-            writeFromInputToOutput(in, out, null);
```
wpride: lol @ tally

javarosa/src/main/java/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -110,4 +78,71 @@ public OutputIOException(IOException internal) {
...
+     */
+    public static void writeFromInputToOutputNew(InputStream is, OutputStream os) throws IOException {
+        writeFromInputToOutputNew(is, os, null);
+    }
+
+    /**
+     * Write is to os and close both
+     */
+    public static void writeFromInputToOutputNew(InputStream is, OutputStream os, StreamReadObserver observer) throws IOException {
+        byte[] buffer = new byte[8192];
```
wpride: What's with 8192?

backend/src/org/commcare/suite/model/RemoteQueryDatum.java
```diff
@@ -26,10 +28,10 @@
     public RemoteQueryDatum() {
     }
 
-    public RemoteQueryDatum(String url, String storageInstance,
+    public RemoteQueryDatum(URL url, String storageInstance,
                             Hashtable<String, XPathExpression> hiddenQueryValues,
                             Hashtable<String, DisplayUnit> userQueryPrompts) {
-        super(storageInstance, url);
+        super(storageInstance, url.toString());
```
phillipm: This is pretty messy, but I didn't really see a better way given constraints. 

Here is my sorta hand wavy justification:
I'm utilizing the `value` field of the `SessionDatum` super class to store the url. But since we have a custom serialization framework, we can't just serialize the URL, we need to turn it into a String first.

Given that the `url` field is externally used as a `URL`, I figured we should represent it as one coming in as well, so that the only place that it isn't a `URL` is inside the class abstraction.

core/org/commcare/core/encryption/CryptUtil.java
phillipm: It was a straight copy

javarosa/src/main/java/org/javarosa/core/io/StreamsUtil.java
phillipm: I think it is what the default buffer size is for similar library implementations of code like this.

PR #338, opened 2016-07-13
Decrement GPS auto-capture timeout / Stop polling after form completes

phillipm:
Pass event along in xform action processing to allow actions to dispatch on event type.

cross-request: https://github.com/dimagi/commcare-android/pull/1380


phillipm:
Found an implementation that doesn't require form completion event triggers

PR #341, opened 2016-07-15
Allow fixtures in suite.xml to be updated on app update

phillipm:
Fixtures in the suite.xml file of an app don't get updated when the app is updated. This causes issues for the scheduler feature used by ICDS.
 
cross-request: https://github.com/dimagi/commcare-android/pull/1391
Fix for http://manage.dimagi.com/default.asp?178966


backend/src/org/commcare/xml/SuiteParser.java
```diff
@@ -118,9 +130,10 @@ public Suite parse() throws InvalidStructureException, IOException,
                         Menu m = new MenuParser(parser).parse();
                         menus.addElement(m);
                     } else if (parser.getName().toLowerCase().equals("fixture")) {
-                        //this one automatically commits the fixture to the global memory
-                        if (!inValidationMode()) {
-                            new FixtureXmlParser(parser, false, fixtureStorage).parse();
+                        if (!isValidationPass) {
+                            // commit fixture to the memory, overwriting existing
+                            // fixture only during install, not init or validation
+                            new FixtureXmlParser(parser, !skipResources, fixtureStorage).parse();
```
phillipm: The fix is to pass `!skipResources` in as the `overwrite` argument to the fixture parser. `!skipResources` is `true` only when suite file is installed. It is `false` on suite initialization and validation passes.

proteusvacuum:
:balloon:

backend/src/org/commcare/resources/ResourceManager.java
```diff
@@ -232,7 +232,7 @@ public void upgrade()
                 //everything here? Locale files are registered elsewhere, and we
                 //can't guarantee we're the only thing in there, so we can't
                 //straight up clear it...
-                platform.initialize(masterTable);
+                platform.initialize(masterTable, true);
```
ctsims: This step isn't really guaranteed to happen. IE: If the upgrade is successful and the app crashes in platform.clearAppState(), we won't process the new fixture.

phillipm: Ah, good call. I added a comment in the code warning for this and we can follow up in another PR to fix it.

PR #342, opened 2016-07-15
Move the format-calendar functions into commcare-core

wpride:
Move the format calendar function from commcare-android CustomFunctionHandler into the core code. Also unified some of the date util logic. 

fixes http://manage.dimagi.com/default.asp?232182
cross-request https://github.com/dimagi/commcare-android/pull/1392

~~Open Question (hold off on merge):
The Ethiopian and Nepali month names are currently hard coded in. Previously we could get these from Android and they would be localized. Need to figure out a new localization strategy.~~


phillipm:
This awesome!

PR #343, opened 2016-07-18
Show case name in saved form for registration cases

phillipm:
Registration forms don't show case names.
For instance the `[Unavailable]` below:
image

This addresses that and moves a lot of session descriptor logic into `commcare-core` so it can be tested without Robolectric, which is damn slow.

Fix for http://manage.dimagi.com/default.asp?223384

cross-request: https://github.com/dimagi/commcare-android/pull/1398


backend/src/org/commcare/util/FormDataUtil.java
```diff
@@ -22,13 +23,54 @@ public static String getTitleFromSession(UserSandbox userSandbox,
...
+            return loadTitleFromEntity(entityDatum, value, evalContext, sessionCopy, userSandbox);
+        }
+    }
+
+    private static Pair<EntityDatum, String> getDatumAndCaseId(CommCareSession session,
+                                                               EvaluationContext evaluationContext) {
+        String datumValue = null;
+        while (session.getFrame().getSteps().size() > 0) {
+            SessionDatum datum = session.getNeededDatum();
+            if (isCaseIdComputedDatum(datum, datumValue, session.getPoppedStep())) {
```
amstone326: `datumValue` should always be null when you pass it in here, so the check for `currentDatumValue == null` in the body of `isCaseIdComputedDatum` seems unnecessary

phillipm: No, there are times when it isn't. Step through it while running the `FormDataUtilTests`

PR #346, opened 2016-07-21
Migrate FormRecord last_modified date

phillipm:
When storing `Date` metadata fields in SQLite DB, use `.getTime()` to store it as a long instead of a `String` via `toString()`. This prevents awkward date parsing issues where ambiguous timezones are used.

Includes migration of `FormRecord` table and updates the date parsing logic to account for seconds since epoch.

cross-request: https://github.com/dimagi/commcare-android/pull/1412


cases/src/org/commcare/cases/instance/CaseInstanceTreeElement.java
```diff
@@ -69,7 +69,6 @@ public CaseInstanceTreeElement(AbstractTreeElement instanceRoot, IStorageUtility
         this.instanceRoot = instanceRoot;
         this.storage = storage;
         this.reportMode = reportMode;
-        storage.setReadOnly();
```
phillipm: All implementations were empty

modern/org/commcare/modern/database/DatabaseHelper.java
```diff
@@ -106,11 +107,18 @@ private static void addDataToValues(HashMap<String, Object> values,
...
-                if(o == null ) { continue;}
-                String value = o.toString();
-                values.put(TableBuilder.scrubName(key), value);
+                if (o == null) {
+                    continue;
+                }
+                String scrubbedKey = TableBuilder.scrubName(key);
+                if (o instanceof Date) {
+                    // store date as seconds since epoch
+                    values.put(scrubbedKey, ((Date)o).getTime());
```
phillipm: Any metadata field that is a `Date` will now be stored as seconds from epoch. This fits nicely with SQLite's Date model, or lack of one. I've verified that the only metadata date field in `commcare-core` and `commcare-android` is in `FormRecord` (and `AndroidLogEntry`, which never reads the metadata value from the DB). @wpride do you use date metadata fields CloudCare-exclusive models?

PR #349, opened 2016-07-28
Invalid data

sjain28:


javarosa/src/main/java/org/javarosa/core/model/data/InvalidData.java
```diff
@@ -0,0 +1,61 @@
+package org.javarosa.core.model.data;
+
+import org.javarosa.core.util.externalizable.DeserializationException;
+import org.javarosa.core.util.externalizable.PrototypeFactory;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+
+/**
+ * Created by Saumya on 6/27/2016.
```
wpride: Class level comment good

javarosa/src/main/java/org/javarosa/core/model/data/InvalidDateData.java
```diff
@@ -0,0 +1,30 @@
+package org.javarosa.core.model.data;
+
+/**
+ * Created by Saumya on 7/25/2016.
```
wpride: ""

PR #350, opened 2016-07-29
add locale data sources and interface

wpride:
Implement abstraction to allow overriding where to get array values from.

cross-request https://github.com/dimagi/commcare-android/pull/1426

### DIFF
--- a/javarosa/src/main/java/org/javarosa/xform/util/CalendarUtils.java
+++ b/javarosa/src/main/java/org/javarosa/xform/util/CalendarUtils.java
@@ -1,6 +1,7 @@
 package org.javarosa.xform.util;
 
-import org.javarosa.core.services.locale.Localization;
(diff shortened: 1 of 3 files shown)


javarosa/src/main/java/org/javarosa/xform/util/CalendarUtils.java
```diff
@@ -366,4 +369,12 @@ public static long toMillisFromJavaEpoch(int year, int month, int day, long mill
         return millisFromMinDay + MIN_MILLIS_FROM_JAVA_EPOCH + millisOffset;
     }
 
+    public static String[] getMonthsArray(String key){
+        System.out.println("Getting months array with key: " + key + ", source: " + arrayDataSource);
```
ctsims: I'd eliminate this. could get quite verbose if there are bulk calculations.

PR #352, opened 2016-08-01
Remove JavaResourceReference which is a dup of ResourceReference

phillipm:
Was seeing an intermittent test failure which I think is due to synchronicity issues with `ReferenceManager`. Instead of fixing that, I decided to remove one of the two reference factory that can be registered to derive `jr://resource/` which fixes the failure at https://github.com/dimagi/commcare-core/blob/master/tests/test/org/commcare/util/reference/test/JavaResourceReferenceTest.java#L26


javarosa/src/test/java/org/javarosa/core/model/test/QuestionDefTest.java
```diff
@@ -168,12 +168,4 @@ public QuestionDef getQ() {
...
         this.q = q;
     }
-
-    public FormEntryPrompt getFep() {
-        return fep;
-    }
-
-    public void setFep(FormEntryPrompt fep) {
-        this.fep = fep;
-    }
```
phillipm: unused

PR #354, opened 2016-08-01
remove DATUM's before popping COMMANDs

wpride:
I want to discuss this with someone so added "don't pull" for now, but this is a fix for http://manage.dimagi.com/default.asp?233398

The advanced module was pushing two extra STATE_DATUM_VALs onto the stack when the action was clicked. When you clicked back on the next screen, instead of returning to the previous command one of these data was removed, then the next, then finally we went to the previous screen on the third click. This simply peels off all datum values from the top of the stock before proceeding with removing the command.

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -344,7 +344,13 @@ public class CommCareSession {
     }
 
     public void stepBack(EvaluationContext evalContext) {
-        // Pop the first thing off of the stack frame, no matter what
+        // First peel off any data pushed onto the stack by the previous command
+        while(frame.getSteps().size() > 0 &&
+                frame.getSteps().lastElement().getType().equals(SessionFrame.STATE_DATUM_VAL)){
+            popSessionFrameStack();
+        }
+
+        // Then, pop the previous command
         popSessionFrameStack();
 
         // Keep popping things off until the value of needed data indicates that we are back to


ctsims:
@amstone326 You've done a few passes on this, can you possibly point will at the test suite you've been using for suite navigation issues so we can add this replication and find a fix?

phillipm:
Might not be understanding problem/how sessions work, but if a pre-form nav session collects 2 case ids via 2 case list selections, wouldn't the session have 2 `STATE_DATUM_VAL` entries in a row? If so, popping both of those off in one go sounds incorrect, because that would skip showing the user the second case select screen.

ctsims:
@phillipm I think it's an ordering issue in this case. IE: you pop off a 'real' select, but based on the order of selections, the "current" select is still the same.

Either way, our strategy should be (/I think is supposed to be) not "go back one step at a time", it's "See what the 'current' select is (IE: select "case_id"), and go back until the 'current' select changes, since it eliminates ambiguity from these situations.

amstone326:
Planning on chatting with Will about this when he has time; I'm a bit confused by the solution as well.

phillipm:
I wasn't able to reproduce this issue on Jeremy's app or on an example app I made. Is anyone else able to reproduce this issue?

wpride:
@phillipm I was able to - was planning on picking this up tomorrow with Aliza

PR #358, opened 2016-08-08
Respect default lang in CLI

phillipm:
Default app language wasn't being respected in the CLI.

Fix for http://manage.dimagi.com/default.asp?234438

Small enough change to add to 2.29


ctsims:
Hey,

I think there's some confusion here. "default" is the name of a locale, not the actual 'default' locale for Kriti's app. That locale needs to get set by respecting the parameter from the Properties in the app profile.

phillipm:
Ah, was wondering if that was the case. Updated

PR #359, opened 2016-08-08
Offline demo user restore

phillipm:
Also extends CLI to take in `-d` flag, which tries to log in as the demo user using the demo user restore present in the app.

cross-request: https://github.com/dimagi/commcare-android/pull/1438


backend/src/org/commcare/xml/SuiteParser.java
```diff
@@ -105,8 +105,8 @@ public Suite parse() throws InvalidStructureException, IOException,
                             entries.put(viewEntry.getCommandId(), viewEntry);
                             break;
                         case EntryParser.REMOTE_REQUEST_TAG:
-                            Entry syncEntry = EntryParser.buildRemoteSyncParser(parser).parse();
-                            entries.put(syncEntry.getCommandId(), syncEntry);
+                            Entry remoteRequestEntry = EntryParser.buildRemoteSyncParser(parser).parse();
+                            entries.put(remoteRequestEntry.getCommandId(), remoteRequestEntry);
```
phillipm: 👍

backend/src/org/commcare/suite/model/OfflineUserRestore.java
```diff
@@ -0,0 +1,135 @@
...
+    public OfflineUserRestore(String reference) {
+        this.reference = reference;
+        this.username = parseUsernameFromRestoreFile(reference);
+        this.password = PropertyUtils.genUUID();
+    }
+
+    public static OfflineUserRestore buildInMemoryUserRestore(InputStream restoreStream) throws IOException {
+        OfflineUserRestore offlineUserRestore = new OfflineUserRestore();
+        offlineUserRestore.restore = new String(StreamsUtil.inputStreamToByteArray(restoreStream));
+        return offlineUserRestore;
```
phillipm: I sorta wonder if for parity it would make sense to set the username/password in this builder, similar to how it is done in `OfflineUserRestore(String reference)`

amstone326: That sounds right to me, although I'm not actually that clear on what this method does. What does the `restore` field that's being built here represent?

backend/src/org/commcare/xml/SuiteParser.java
amstone326: Ha, pretty sure this was your change :)

backend/src/org/commcare/resources/model/installers/OfflineUserRestoreInstaller.java
```diff
@@ -0,0 +1,58 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean install(Resource r, ResourceLocation location,
+                           Reference ref, ResourceTable table,
+                           CommCareInstance instance, boolean upgrade)
+            throws UnresolvedResourceException, UnfullfilledRequirementsException {
+        try {
+            OfflineUserRestore offlineUserRestore = OfflineUserRestore.buildInMemoryUserRestore(ref.getStream());
```
ctsims: Where are we catching exceptions for invalid file contents and things like that? Currently it feels like they'll fail over by throwing generic exceptions that look like IO exceptions or other network traffic problems

backend/src/org/commcare/suite/model/OfflineUserRestore.java
```diff
@@ -0,0 +1,135 @@
...
+                if (eventType == KXmlParser.START_TAG) {
+                    if (parser.getName().toLowerCase().equals("username")) {
+                        parser.next();
+                        return parser.getText();
+                    }
+                }
+                eventType = parser.next();
+            } while (eventType != KXmlParser.END_DOCUMENT);
+        } catch (IOException | XmlPullParserException | InvalidReferenceException e) {
+            return null;
```
ctsims: This doesn't seem like the right way to handle all of these exceptions to me, precisely. Wrapping like this accounts for a lot of our "silent" issues with install/restore that are super frustrating to people, we should make sure we aren't swallowing these problems

backend/src/org/commcare/suite/model/OfflineUserRestore.java
```diff
@@ -0,0 +1,135 @@
...
+    public void setID(int ID) {
+        recordId = ID;
+    }
+
+    @Override
+    public int getID() {
+        return recordId;
+    }
+
+    private static String parseUsernameFromRestoreFile(String localRestoreReference) {
```
ctsims: Is there a reason we don't use the normal user parser and parser factory here? It seems like that re-uses more stable code rather than redoing the parsing on its own.

backend/src/org/commcare/util/CommCarePlatform.java
```diff
@@ -170,4 +172,13 @@ public String getMenuDisplayStyle(String menuId) {
         }
         return commonDisplayStyle;
     }
+
+    public OfflineUserRestore getDemoUserRestore() {
+        return offlineUserRestore;
+    }
+
+    @Override
+    public void registerDemoUserRestore(OfflineUserRestore offlineUserRestore) {
```
ctsims: It seems dangerous to assume there is only one of them like this, given that the spec doesn't outline (and the parser doesn't validate) that there is only one restore to process

tests/resources/app_structure/suite.xml
```diff
@@ -6,6 +6,12 @@
     </resource>
   </xform>
 
+  <user-restore username="test" password="123">
+    <resource id="foo" version="1" descriptor="demo user restore">
+      <location authority="local">./user_restore.xml</location>
```
ctsims: where is this file? It doesn't seem inlcuded

ctsims:
It seems like it would be helpful to also have a test at this level, since we have all of the harness code to do so. One thing I think would be a good move is for us to also start testing expectations around *incorrect* behavior as well as the correct behavior, IE: Trying to parse an app install/update which has malformed demo data, and seeing whether it breaks when and how we expect.

backend/src/org/commcare/xml/SuiteParser.java
phillipm: lol

backend/src/org/commcare/suite/model/OfflineUserRestore.java
amstone326: Ah ok nevermind makes more sense now, although adding a related question below

backend/src/org/commcare/suite/model/OfflineUserRestore.java
```diff
@@ -0,0 +1,135 @@
...
+        this.password = PropertyUtils.genUUID();
+    }
+
+    public static OfflineUserRestore buildInMemoryUserRestore(InputStream restoreStream) throws IOException {
+        OfflineUserRestore offlineUserRestore = new OfflineUserRestore();
+        offlineUserRestore.restore = new String(StreamsUtil.inputStreamToByteArray(restoreStream));
+        return offlineUserRestore;
+    }
+
+    public InputStream getRestoreStream() {
```
amstone326: @phillipm Could you add a brief comment as to when each of the 2 scenarios here will be used? Would be helpful for my understanding and I think future users of this as well

backend/src/org/commcare/resources/model/installers/OfflineUserRestoreInstaller.java
amstone326: @ctsims Well right now we're actually just not parsing the contents of the file at all until we go to use it when logging in, so no errors would be thrown until then. But I think it makes sense for both `OfflineUserRestore.buildInMemoryUserRestore` and the constructor used by Android to try to parse the file contents as a restore file at this point, and throw any resulting parse errors from that. Does that sound right?

amstone326: @ctsims This actually doesn't look simple to hook up (since the parsing and committing of data from a restore file is currently very coupled, and we would want to be able to very clearly parse but not commit any of the data at this stage), so going to wait for confirmation that this is the best approach before attempting.

backend/src/org/commcare/util/CommCarePlatform.java
amstone326: Not sure I'm following why it's dangerous. Agreed that the spec doesn't say this explicitly right now, but it seems reasonable to add that as a constraint?

backend/src/org/commcare/suite/model/OfflineUserRestore.java
amstone326: Mostly fixed this up, will finalize once we decide to what extent we're parsing the restore file at install time

tests/resources/app_structure/suite.xml
amstone326: That's because it already exists prior to this PR. I'm actually not sure why this block was even added to the suite.xml here though, since it's not being used anywhere for tests

backend/src/org/commcare/suite/model/OfflineUserRestore.java
amstone326: @phillipm Bumping on this

amstone326:
Ok @ctsims. I resolved 2 of the issues you raised:
1. Added a test for parsing the demo user restore, including very limited checks for a malformed restore.
2. The install-time parsing that is done in `OfflineUserRestore.checkThatRestoreIsValid` is now using an instance of `DataModelPullParser`. 

Right now, `checkThatRestoreIsValid` does the bare minimum amount of validation on the restore file, making sure that it is indeed for a user of type demo, and that a username was parsed successfully. There is definitely a lot more that could be done here as far as parsing the restore file more completely at install time, in order to ensure that it is properly formed and will be usable at the point of login, but I think I'm still a bit unsure on how extensively it is worth doing that, given the challenge of doing so, and what stands to be gained. If you can take a look at what is done already and give your opinion on how much more is worth trying to validate, that would be really helpful.

backend/src/org/commcare/suite/model/OfflineUserRestore.java
phillipm: I added some comments and consolidated some of the code around this in https://github.com/dimagi/commcare-core/pull/359/commits/f075d42460dd3627dee93e7b65d94225fbb1a2ef

amstone326: Cool, thanks

backend/src/org/commcare/suite/model/OfflineUserRestore.java
```diff
@@ -0,0 +1,182 @@
...
+        parser.parse();
+    }
+
+    private TransactionParser buildUserParser(KXmlParser parser) {
+        return new UserXmlParser(parser) {
+
+            @Override
+            protected void commit(User parsed) throws IOException, UnfullfilledRequirementsException {
+                if (!parsed.getUserType().equals(User.TYPE_DEMO)) {
+                    throw new UnfullfilledRequirementsException(
```
ctsims: Traditionally we use UnfillfilledRequrements as a signal for a configuration resource which is currently invalid but the user could make valid through some action (IE: "You need to update CommCare before proceeding"). I think "InvalidStructureException" is the more appropriate error here, which means "The config file that was requested for processing isn't created correctly, and needs to be changed before it will ever worlk.

backend/src/org/commcare/suite/model/OfflineUserRestore.java
```diff
@@ -0,0 +1,182 @@
...
+            public TransactionParser getParser(KXmlParser parser) {
+                String name = parser.getName();
+                if ("registration".equals(name.toLowerCase())) {
+                    return buildUserParser(parser);
+                }
+                return null;
+            }
+        };
+
+        DataModelPullParser parser = new DataModelPullParser(getRestoreStream(), factory, true, false);
```
ctsims: This looks great, by the way. Glad this ended up working.

backend/src/org/commcare/suite/model/OfflineUserRestore.java
```diff
@@ -0,0 +1,182 @@
...
+    public void setID(int ID) {
+        recordId = ID;
+    }
+
+    @Override
+    public int getID() {
+        return recordId;
+    }
+
+    private void checkThatRestoreIsValid()
```
ctsims: There's a potential argument to be made that we should actually be including the full restore semantics that we'll use in the app but in "read only" mode somehow. For now that's almost certainly overkill, though.

backend/src/org/commcare/suite/model/OfflineUserRestore.java
amstone326: Cool that makes sense, updated

PR #361, opened 2016-08-09
Wsp/stack push datum fix

wpride:
Fix for http://manage.dimagi.com/default.asp?233398

Two issues here:
We were parsing push datum's as STATE_DATUM_VALs instead of STATE_DATUM_COMPUTEDs which was messing with the stepBack() function
When iterating backwards in stepBack() we weren't checking the popped stack - only the "needed" datum which won't necessarily tell us that we're on a computed.

UPDATE
Instead of interpreting all stack push datums as computed type, push as a new unknown type. Then when the session encounters these, determine the "best" interpretation as either a command or computed based on the available entries of the extant stack. 

code buddy @amstone326


tests/resources/case_title_form_loading/suite.xml
```diff
@@ -53,6 +53,24 @@
...
+        <text>
+          <locale id="case_list_form.m0"/>
+        </text>
+        <text form="image">
+          <locale id="case_list_form.m0.icon"/>
+        </text>
+      </display>
+      <stack>
+        <push>
+          <command value="'m0-f0'"/>
```
wpride: Needed to push to a command that requires a nodeset, since otherwise getNeededData() will be null in the old (incorrect) loop predicate and would still work correctly

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -344,14 +344,52 @@ public Menu getMenu(String id) {
...
+                String childCommand = entry.getCommandId();
+                if (childCommand.equals(poppedId)) {
+                    return SessionFrame.STATE_COMMAND_ID;
+                }
+            }
+        }
+        return SessionFrame.STATE_DATUM_COMPUTED;
+    }
+
+    private boolean popNext(EvaluationContext evalContext){
```
amstone326: Can you rename this to better describe what it does? It's not actually popping anything off, just checking the thing that was popped. Something like `lastPoppedWasComputed()`?

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -427,7 +465,9 @@ public void syncState() {
         this.popped = null;
 
         for (StackFrameStep step : frame.getSteps()) {
-            if (SessionFrame.STATE_DATUM_VAL.equals(step.getType())) {
+            if (SessionFrame.STATE_DATUM_VAL.equals(step.getType()) ||
+                    SessionFrame.STATE_UNKNOWN.equals(step.getType()) &&
```
amstone326: why did you add this other condition?

backend/src/org/commcare/session/SessionInstanceBuilder.java
```diff
@@ -39,7 +39,8 @@ private static void addSessionNavData(TreeElement sessionRoot, SessionFrame fram
      */
     private static void addDatums(TreeElement sessionData, SessionFrame frame) {
         for (StackFrameStep step : frame.getSteps()) {
-            if (SessionFrame.STATE_DATUM_VAL.equals(step.getType())) {
+            if (SessionFrame.STATE_DATUM_VAL.equals(step.getType()) ||
```
amstone326: what is this change for?

amstone326:
@wpride Good to merge once `popNext` is renamed and I understand those 2 changes I asked about.

backend/src/org/commcare/session/CommCareSession.java
wpride: Because the new STATE_UNKNOWN type can also possibly be a computed that we want to put into the collectedDatum table

backend/src/org/commcare/session/SessionInstanceBuilder.java
wpride: Similar to above, we have this new type that we can also possibly want to move into the sessionData instance

amstone326: but the type you added here was `STATE_DATUM_COMPUTED`, which isn't the new type. I understand the addition you made above now, but shouldn't this one be the same as that?

wpride: Ahh gotcha. If you look at this line https://github.com/dimagi/commcare-core/blob/d98f844c7c95b17ec278d8ecc4cd8650845261a2/backend/src/org/commcare/session/CommCareSession.java#L563-L563

I changed the getFrame() call to perform a copy AND evaluate the unknowns to either commands or computeds, hence when they get to this point of the code they will have been evaluated to that type and need to be added.

wpride: Resolution to earlier: so previously every stack push was actually marked as a DATUM (Ie CASE_ID) type:

https://github.com/dimagi/commcare-core/blob/e91993ad92b3c445f95f48e31de63a6eee9daee5/backend/src/org/commcare/xml/StackFrameStepParser.java#L28-L28

So the logic here was still working. Now those are marked as UNKNOWN and later evaluated to COMPUTED types here.

PR #362, opened 2016-08-10
Fix Nepali date widget change on swipe... again

phillipm:
My description from [an old PR](https://github.com/dimagi/commcare-android/pull/1129) sums it up pretty well:

> Currently, the Nepali date widget changes the date back one day every time you scroll to and from the widget. This is due to some sort of timezone/calendar day issue where if you are in Nepal and it is, say 10am on the 12th, and hence the 11th in Greenwich Mean Time, then restoring the stored date value restores to the 11th, not the 12th.

More specifically, I think my old PR was only a partial fix. Instead of storing an millisecond offset used to convert dates to and from millis since epoch in a timezone agnostic way, we should use the timezone offset. This is because the millisecond offset info is lost when you reload the question widget.

I've moved the calendar tests from `commcare-android` to here because the calendar code now lives at core. I added a failing test (`testBaseDateSerialization`) to expose the issue and then applied the fixing changes.

cross-request: https://github.com/dimagi/commcare-android/pull/1440

### DIFF
--- a/javarosa/src/main/java/org/javarosa/xform/util/CalendarUtils.java
+++ b/javarosa/src/main/java/org/javarosa/xform/util/CalendarUtils.java
@@ -276,12 +276,20 @@ public class CalendarUtils {
         );
     }
 
+    /**
+     * @param millisFromJavaEpoch Argument must be normalized to UTC to prevent
+     *                            timezone issues when casting to a calendar date
+     */
     public static UniversalDate fromMillis(long millisFromJavaEpoch, TimeZone currentTimeZone) {
-        // Since epoch calculations are relative to GMT, take current timezone
+        // Since epoch calculations are relative to UTC, take current timezone
         // into account. This prevents two time values that lie on the same day
         // in the given timezone from falling on different GMT days.
-        int timezoneOffsetFromGMT = currentTimeZone.getOffset(millisFromJavaEpoch);
-        long millisFromMinDay = timezoneOffsetFromGMT + millisFromJavaEpoch - MIN_MILLIS_FROM_JAVA_EPOCH;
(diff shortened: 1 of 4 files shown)


javarosa/src/main/java/org/javarosa/xform/util/CalendarUtils.java
```diff
@@ -359,14 +359,19 @@ public static UniversalDate incrementYear(UniversalDate date) {
...
+    public static long toMillisFromJavaEpoch(int year, int month, int day) {
+        return toMillisFromJavaEpoch(year, month, day, TimeZone.getDefault());
+    }
+
+    public static long toMillisFromJavaEpoch(int year, int month, int day, TimeZone currentTimeZone) {
         int daysFromMinDay = countDaysFromMinDay(year, month, day);
         long millisFromMinDay = daysFromMinDay * UniversalDate.MILLIS_IN_DAY;
-        return millisFromMinDay + MIN_MILLIS_FROM_JAVA_EPOCH + millisOffset;
+        int timezoneOffsetFromGMT = currentTimeZone.getOffset(millisFromMinDay);
+        return millisFromMinDay + MIN_MILLIS_FROM_JAVA_EPOCH - timezoneOffsetFromGMT;
```
amstone326: Definitely just not following this completely, but can you explain why you always want to be subtracting the `timezoneOffsetFromGMT`

phillipm: The idea is to normalize the 'milliseconds since epoch' to not include timezone info, since we are representing a date not a datetime. I guess I did half of this in my last fix; the half when converting from millis to a date. The issue is that I didn't do the other half: removing the timezone offset when converting a date to 'milliseconds since epoch'. That is what this PR does.

PR #363, opened 2016-08-10
Flexible case tiles in case list

amstone326:
Used by https://github.com/dimagi/commcare-android/pull/1444


backend/src/org/commcare/suite/model/Detail.java
```diff
@@ -70,31 +84,13 @@ public Detail() {
     }
 
     public Detail(String id, DisplayUnit title, String nodeset,
```
amstone326: refactored this to only use 1 constructor because nothing was calling the other 2

phillipm:
Good to go once conflicts are resolved

amstone326:
@phillipm Whoops hadn't noticed, resolved

PR #366, opened 2016-08-11
Some linting

phillipm:


javarosa/src/main/java/org/javarosa/core/util/SizeBoundVector.java
```diff
@@ -34,26 +30,14 @@ public int getAdditional() {
     }
 
     public void addBadImageReference() {
-        badImageReferenceCount++;
+        // TODO: remove this
```
amstone326: why are you not just removing this?

phillipm: It is sorta hard to fully remove it because it is used by the core versions of the suite/xform installer's `verifyInstallation`. That code isn't run currently on the CLI, but we might want to enable it. Figuring out the right path forward seems like way more work than I want to do for this right now. I can revert the changes in this file if you'd like.

amstone326: Got it, why are you removing the method body then? I don't understand the purpose of leaving the method in, but leaving it empty.

phillipm: `badImageReferenceCount` is never read, so keeping it around didn't seem useful

amstone326: Ok, I think it probably makes more sense to revert the changes to this file for now than leaving it in a half-but-not-fully-deleted state

PR #367, opened 2016-08-11
executeStackOperations refactor

phillipm:
Break `CommCareSession.executeStackOperations` up into helper methods.
And other misc. refactoring.

Review commit-by-commit

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -554,19 +554,11 @@ public class CommCareSession {
                 case StackOperation.OPERATION_CREATE:
                     // Ensure no frames exist with this ID
                     if (matchingFrame == null) {
-                        Boolean currentFramePushedOrNull =
-                                performPush(op, new SessionFrame(frameId), true, currentFramePushed, onDeck, ec);
-                        if (currentFramePushedOrNull != null) {
-                            currentFramePushed = currentFramePushedOrNull;
-                        }
+                        currentFramePushed = performPush(op, new SessionFrame(frameId), true, currentFramePushed, onDeck, ec);
                     }
                     break;
                 case StackOperation.OPERATION_PUSH:
-                    Boolean currentFramePushedOrNull =
-                            performPush(op, matchingFrame, false, currentFramePushed, onDeck, ec);
-                    if (currentFramePushedOrNull != null) {
-                        currentFramePushed = currentFramePushedOrNull;
-                    }
+                    currentFramePushed = performPush(op, matchingFrame, false, currentFramePushed, onDeck, ec);
                     break;
                 case StackOperation.OPERATION_CLEAR:
                     performClearOperation(matchingFrame, op, ec);
@@ -579,7 +571,7 @@ public class CommCareSession {
         return popOrSync(onDeck);
     }
 
-    private Boolean performPush(StackOperation op, SessionFrame matchingFrame,
+    private boolean performPush(StackOperation op, SessionFrame matchingFrame,
                                 boolean isNewFrame, boolean currentFramePushed,
                                 SessionFrame onDeck, EvaluationContext ec) {
         if (op.isOperationTriggered(ec)) {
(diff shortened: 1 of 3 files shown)


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -206,7 +207,8 @@ public String getNeededData(EvaluationContext evalContext) {
                 && entries.elementAt(0) instanceof SyncEntry
                 && ((SyncEntry)entries.elementAt(0)).getSyncPost().isRelevant(evalContext)) {
             return SessionFrame.STATE_SYNC_REQUEST;
-        } else if (entries.size() > 1 || !entries.elementAt(0).getCommandId().equals(currentCmd)) {
+        } else if (entries.size() > 1 ||
+                (!entries.isEmpty() && !entries.elementAt(0).getCommandId().equals(currentCmd))) {
```
phillipm: Adding `entries.isEmpty()` prevents a index out of range crash, but it might take us into undefined territory in terms of session nav...

amstone326: Have you seen this crash on ACRA?

backend/src/org/commcare/xml/StackOpParser.java
```diff
@@ -17,38 +14,34 @@
 /**
  * @author ctsims
  */
-public class StackOpParser extends ElementParser<StackOperation> {
```
amstone326: why are you removing `public` here and in the constructor?

backend/src/org/commcare/session/CommCareSession.java
phillipm: No, ran across it when dealing with a misconfigured suite file. I think I'll just remove it a leave it to crash for now

amstone326: yea I think that makes more sense

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -604,6 +595,25 @@ public boolean executeStackOperations(Vector<StackOperation> ops, EvaluationCont
...
+                matchingFrame = onDeck;
+            }
+
+            for (StackFrameStep step : op.getStackFrameSteps()) {
+                matchingFrame.pushStep(step.defineStep(ec));
+            }
+
+            return pushNewFrame(matchingFrame, isNewFrame, currentFramePushed);
+        }
+        return null;
```
amstone326: why not stick to your original strategy of returning `currentFramePushed` in this case, so that you don't have to do the null check after calling this method?

phillipm: We only want to update `currentFramePushed` if `op.isOperationTriggered(ec)` is true

phillipm: At least, that must be done to preserve the existing behavior. Not sure if that behavior is necessary or not

amstone326: Right, but you're passing in the original value of `currentFramePushed` to this method, so if you just return that value in the case where `op.isOperationTriggered(ec)` is false, you won't be updating the value.

backend/src/org/commcare/xml/StackOpParser.java
phillipm: When I upgraded to Android Studio 2.2 it started highlighting things that can be made package local (the same as `protected`). My resistance to ignoring lint warnings is very low... I should probably just disable that specific highlighting because it may break FormPlayer imports. In this case we should be fine

amstone326: Gotcha. I know this is nitpicky, but given that we use the convention of putting `public` before our class names and constructors uniformly throughout the entire codebase, I think I'd rather not start changing it in random places just because the linter says so. It gives (at least me) the impression that it was done for a particular reason, since it's such a departure from our standard. If we want to make a separate pass at converting everything that can be to to package local, we can do that.

PR #368, opened 2016-08-11
Extend session stack frame with 'mark' and 'rewind' constructs

phillipm:
Extension to the stack logic to allow 'register from the case list' and 'case claim' to select the case that was registered/claimed and continue forward navigation when in the presence of advanced modules.

The current issue is that when you perform 'registration from the case list' in an advanced module, you may have gone through several case list selections. Hence it is very cumbersome to copy those selected cases over to new session frames; the only way to do so is to have HQ analyze the navigation workflow and hard-code the data propagation. The reason we have copy data between frames is that when the registration completes, a new frame is created to forward the user to the next place.

So the idea behind this fix is to extend sessions in the following 2 ways:

 - Add a new `mark` frame step. This step marks a point to rewind to. When it is added to a stack frame the current needed datum for that frame is computed and stored with it. Once a `rewind` targeting a `mark` is processed, that datum is set with the rewind value.
 - Add a `rewind` frame step which has a value associated with it. When rewinds are encountered all steps in between it and the nearest mark above it will be cleared and the datum in the `mark` will be turned into a `case_id`/`SessionFrame.STATE_DATUM_VAL` entry.

Here is an example of the syntax:
```{xml}
<stack>
    <push>
        <command value="'m1'"/>
        <mark/>
        <datum id="child_name" value="'steve'"/>
        <rewind value="'billy'"/>
        <!-- results in child_name getting set to 'billy' -->
    </push>
</stack>
```

How do these 2 constructs come together to solve the problem?
Say you have a form entry that requires 2 case ids:
```{xml}
<entry>
  <command id="child-visit-form"/>
  ...
  <session>
    <datum id="mother_id" nodeset="..." value="./@case_id" detail-select="m1_case_short" detail-confirm="m1_case_long"/>
    <datum id="child_id" nodeset="..." value="./@case_id" detail-select="m1_case_short" detail-confirm="m1_case_long"/>
```

# Case Claim
and associated with the `m1_case_short` case list is the case claim action:

```{xml}
<action>
  <display>
    <text>
      <locale id="case_search.m1"/>
    </text>
  </display>
  <stack>
    <push>
      <mark/>
      <command value="'search_command.m1'"/>
    </push>
  </stack>
</action>
```

The action, when triggered, runs the `search_command.m1` claim process:
```{xml}
<sync-request>
  <post url="https://india.commcarehq.org/..." relevant="...">
     ...
  </post>
  <command id="search_command.m1">
    <display>
      <text>
        <locale id="case_search.m1"/>
      </text>
    </display>
  </command>
  <instance id="querysession" src="jr://instance/session"/>
  <session>
    <query url="https://india.commcarehq.org/a/enikshay-test/phone/search/" storage-instance="results">
      ...
    </query>
    <datum id="case_id" nodeset="instance('results')/results/case[@case_type='person'][@status='open']" value="./@case_id" detail-select="m1_case_short" detail-confirm="m1_case_long"/>
  </session>
  <stack>
    <push>
      <rewind value="instance('querysession')/session/data/case_id"/>
    </push>
  </stack>
</sync-request>
```

Now you start your workflow
 - you select a mother case and realize you need to claim the child
   your stack looks like:
``` 
current frame:
  (COMMAND_ID "child-visit-form")
  (CASE_ID mother_id "nancy")
```

 - you trigger the claim command from the child case list. This will copy the current frame onto the stack and push the claim command onto the current frame:
``` 
current frame:
  (COMMAND_ID "child-visit-form")
  (CASE_ID mother_id "nancy")
  (MARK)
  (COMMAND_ID "search_command.m1")
```
 - you find a case and claim it, this pushes the claimed case onto the current frame as the return value:
``` 
current frame:
  (COMMAND_ID "child-visit-form")
  (CASE_ID mother_id "nancy")
  (MARK)
  (COMMAND_ID "search_command.m1")
  (REWIND "billy")
```
When the step pushes are processed the session logic immediately resolves that frame to 
``` 
current frame:
  (COMMAND_ID "child-visit-form")
  (CASE_ID mother_id "nancy")
rewind value: "billy"
```
- The session then sees what the next needed datum is for the `child-visit-form` command, which is `child_id`. It notices that there is a rewind value set, and uses that for the `child_id`
- 🎉 🎈 

# Double Management
and associated with the `m1_case_short` case list is the case claim action:

```{xml}
<action>
  <display>
    <text>
      <locale id="case_list_form.m1"/>
    </text>
  </display>
  <stack>
    <push>
      <mark/>
      <command value="'child-registration-form'"/>
      <datum id="case_id_new_case_0" value="uuid()"/>
      <datum id="child_double_management" value="true()"/>
    </push>
  </stack>
</action>
```

and when you trigger the action you are taken to the `child-registration-form`, with the following entry:
```{xml}
<entry>
  <form>http://openrosa.org/formdesigner/11FAC65A-F2CD-427F-A870-CF126336AAB5</form>
  <command id="child-registration-form">
    <text>
      <locale id="forms.m0f0"/>
    </text>
  </command>
  <instance id="commcaresession" src="jr://instance/session"/>
  <session>
    <datum id="case_id_new_case_0" function="uuid()"/>
  </session>
  <stack>
    <push if="count(instance('commcaresession')/session/data/person_double_management) = 1 and 
              count(instance('casedb')/casedb/case[@case_id=instance('commcaresession')/session/data/case_id_new_case_0]) &gt; 0">
      <rewind value="instance('commcaresession')/session/data/case_id_new_case_0"/>
    </push>
    <!-- normal, non-double management form linking can occur here -->
    <!-- NOTE: these stack operations will be skipped if a rewind occurs -->
  </stack>
</entry>
```

# Other Implementation Details:
 - A rewind that doesn't have an associated mark is skipped over
 - If a rewind is processed, subsequent stack frame steps and operations will be disregarded.
 - mark/rewinds can occur in `create` stack operations but will just get simplified out at the time that the `create` is processed.
 - `rewind`s will find the nearest `mark`. Hence, you can have more than one more `mark`, in the case that you want to `rewind` more than once.


snopoke:
What problems is this trying to solve?

phillipm:
@snopoke Selecting the created case after performing 'Registration from the case list' in an advanced module doesn't work. Also, resuming the current session with the claimed case doesn't work. The issue can be solved by HQ manually encoding every possible return path via conditional frame creations, but I saw this as a potentially smaller lift. I'll look into expanding the PR description to clarify exactly how the 2 approaches would work so we can contrast them.

snopoke:
Nice, thanks for the explanation.

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -203,6 +203,8 @@ public String getNeededData(EvaluationContext evalContext) {
 
         if (needDatum != null) {
             return needDatum;
+        } else if (entries.isEmpty()) {
+            throw new RuntimeException("Collected datums don't match required datums for entries at command " + currentCmd);
```
phillipm: Originally talked about in https://github.com/dimagi/commcare-core/pull/367#discussion_r74509690 this came up again and was confusing the hell out of me. Failing explicitly instead of with an NPE would probably help avoid such confusion, so opting to add it back

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -612,31 +626,18 @@ private SessionFrame updateMatchingFrame(String frameId) {
...
-            // somehow.
-            cleanStack();
-
-            // OK, now we want to take the current frame and put it up on the frame stack unless
-            // this frame is dead (IE: We're closing it out). then we'll push the new frame
-            // on top of it.
-            if (!frame.isDead() && !currentFramePushed) {
-                frameStack.push(frame);
-                return true;
-            }
```
phillipm: @ctsims I removed this code, which pushes the current frame onto the top of the stack frame because it didn't make sense to me and was complicating my implementation of this extension.  (note that when I refactored this code in https://github.com/dimagi/commcare-core/pull/367 I made a mistake, the `return true` should really be `currentFramePushed = true` to persist the pre-refactor behavior)
Can you verify that this isn't needed? If it is needed, can you explain why?

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -317,7 +320,14 @@ private SessionDatum getFirstMissingDatum(OrderedHashtable datumsCollectedSoFar,
...
             if (!datumsCollectedSoFar.containsKey(datum.getDataId())) {
-                return datum;
+                if (nextDatumValue != null) {
+                    // set the needed datum to the return value from the last
+                    // popped frame and continue looking for needed datums
+                    setDatum(datum.getDataId(), nextDatumValue);
+                    nextDatumValue = null;
+                } else {
+                    return datum;
+                }
```
phillipm: Sorta feel like this is a bad place to be doing this work, but couldn't think of a better place

phillipm:
@ctsims Can you review this one when you get a chance? @proteusvacuum is going to start the HQ side of it soon because the UATBC timelines are pretty tight, so would be good to get feedback on whether you see any critical issues with this approach.

ctsims:
@phillipm So the basic idea here is that:

1) <copy> always moves a full copy of the frame onto the stack
2) When a <return> is processed, it goes into the "next" element requested (<datum>, or <command>, or whatnot)

A few questions for you:
1) So I click "claim" or "new case (dbl mgmt)" or whatever other stack operation I'm doing and then after I get to the form I press "back". I am still on the 'new' frame, right? Is the other frame still on the stack? How does the app know to invalidate that frame based on my actions?  

If I press "Claim -> Back -> Claim", I'm popping a second frame on, right?

2) The <return> construct executes in the post-form-entry/claim context, right? That's the context after processing has occurred but before the stack frame is "finalized". <return> seems like a really helpful construct, but we should make sure the semantically it's tied to the right "side" of the split.

proteusvacuum:
Tracking double management HQ side here: manage.dimagi.com/default.asp?235199

phillipm:
@ctsims This is ready to be reviewed again

backend/src/org/commcare/session/SessionFrame.java
```diff
@@ -108,6 +110,28 @@ public StackFrameStep popStep() {
         return recentPop;
     }
 
+    protected boolean rewindToMarkAndSet(String value) {
+        int markIndex = getLatestMarkPosition(steps);
+
+        if (markIndex >= 0) {
+            String markDatumId = steps.get(markIndex).getId();
+            steps = new Vector<>(steps.subList(0, markIndex));
+            steps.addElement(new StackFrameStep(SessionFrame.STATE_DATUM_VAL, markDatumId, value));
```
phillipm: Does it make sense to turn a `mark`/`rewind` pair into `STATE_DATUM_VAL` or `STATE_DATUM_COMPUTED`?

backend/src/org/commcare/session/CommCareSession.java
ctsims: Managing the "new Frame" code is important. If you are using ID's you can only have one frame with a given ID on the stack at a time. Removing this code will prevent that from working right.

PR #369, opened 2016-08-12
Rename case claim code to not use 'sync'

phillipm:
Realized that my naming for the case claim feature was super confusing. Usage of 'sync' is unneeded; changed them to 'remote request'

Requires HQ changes, but @proteusvacuum agreed to get those in on Monday.

cross-request: https://github.com/dimagi/commcare-android/pull/1451

### DIFF
--- a/backend/src/org/commcare/xml/EntryParser.java
+++ b/backend/src/org/commcare/xml/EntryParser.java
@@ -30,7 +30,7 @@ import java.util.Vector;
 public class EntryParser extends CommCareElementParser<Entry> {
     private static final String FORM_ENTRY_TAG = "entry";
     private static final String VIEW_ENTRY_TAG = "view";
-    private static final String SYNC_REQUEST_TAG = "sync-request";
+    protected static final String SYNC_REQUEST_TAG = "remote-request";
     private final String parserBlockTag;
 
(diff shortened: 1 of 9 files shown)


amstone326:
@phillipm If the purpose of this is just to avoid confusion in code readability, what is the advantage of sneaking it into 2.29?

phillipm:
The reasoning is that nobody uses case claim currently, so these changes won't break anyone's apps. If we wait, then we will have to have everyone redeploy their apps from HQ once they upgrade their version of CommCare

amstone326:
@phillipm Ok makes sense. We should just make sure to re-run the case claim integration test on 2.29 once this is merged and Farid's changes are deployed.

backend/src/org/commcare/xml/EntryParser.java
```diff
@@ -30,7 +30,7 @@
 public class EntryParser extends CommCareElementParser<Entry> {
     private static final String FORM_ENTRY_TAG = "entry";
     private static final String VIEW_ENTRY_TAG = "view";
-    private static final String SYNC_REQUEST_TAG = "sync-request";
+    protected static final String SYNC_REQUEST_TAG = "remote-request";
```
amstone326: can you rename this variable to not include "sync", since nothing else does anymore

PR #374, opened 2016-08-17
Allow menu screen to handle ViewEntry

wpride:
fix for http://manage.dimagi.com/default.asp?235260

cc @benrudolph


benrudolph:
looks good, should i merge?

wpride:
sure, but good to note that what's merged into master in commcare-core doesn't really affect formplayer since it targets specific commits/branches

PR #375, opened 2016-08-18
Refactoring before demo user work

phillipm:
Moving refactors from https://github.com/dimagi/commcare-core/pull/359 into a clean PR

cross-request: https://github.com/dimagi/commcare-android/pull/1458


backend/src/org/commcare/resources/model/installers/CacheInstaller.java
```diff
@@ -46,24 +47,20 @@
         return cacheStorage;
     }
 
-    int cacheLocation;
+    @Override
```
amstone326: I know you didn't add this but what is the purpose of having this abstract method declaration here, when it's just overriding an abstract method of its super class?

phillipm: Hmm, yeah, it isn't useful here since the parent implementation is also abstract. We can take it later

PR #376, opened 2016-08-19
Get correct needed datum when using child modules

phillipm:
Fix for some issues in http://manage.dimagi.com/default.asp?234442

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -125,22 +125,21 @@ public class CommCareSession {
      */
     private Vector<Entry> getEntriesForCommand(String commandId,
                                                OrderedHashtable<String, String> currentSessionData) {
+        Vector<Entry> entries = new Vector<>();
         for (Suite s : platform.getInstalledSuites()) {
             for (Menu m : s.getMenus()) {
                 // We need to see if everything in this menu can be matched
                 if (commandId.equals(m.getId())) {
-                    return getEntriesFromMenu(m, currentSessionData);
+                    entries.addAll(getEntriesFromMenu(m, currentSessionData));
                 }
             }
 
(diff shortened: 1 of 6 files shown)


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -125,22 +126,21 @@ public CommCareSession(CommCareSession oldCommCareSession) {
...
+        Vector<Entry> entries = new Vector<>();
         for (Suite s : platform.getInstalledSuites()) {
-            for (Menu m : s.getMenus()) {
-                // We need to see if everything in this menu can be matched
-                if (commandId.equals(m.getId())) {
-                    return getEntriesFromMenu(m, currentSessionData);
+            List<Menu> menusWithId = s.getMenusWithId(commandId);
+            if (menusWithId != null) {
+                for (Menu menu : menusWithId) {
+                    entries.addAll(getEntriesFromMenu(menu, currentSessionData));
```
amstone326: Is this the fix? curious how this resolves Kriti's issue of displaying the wrong menu name. Or is that not what this PR is addressing? (tried to read through the ticket you linked but there were a few different things going on there)

phillipm: This PR resolves a session ordering issue with when a case list is displayed. If there are 2 forms that use different case lists and are associated with a single menu command, then it used to just take the first one, which resulted in the case list being shown before form selection.

amstone326: Gotcha, thanks. Do we have plans to resolve the menu naming issue, just out of curiosity?

phillipm: Not sure, I wanted to get the big ones out of the way, but I'll reasses that issue soon. Might be the case that Kriti can use advanced modules instead.

PR #380, opened 2016-08-26
rename Style to GridStyle 

amstone326:
Currently, the `style` block of a `field` within a `detail` looks like this:

screen shot 2016-08-26 at 2 15 42 pm

which makes it look like the attributes of a `style` block (alignment, font size, etc.) are generically applicable to all detail fields, when in fact they are only used if there is a `grid` element within the `style` block as well. Ideally, I think it would make sense to invert the elements so that `grid` is the outer element, and `style` is a sub-element of that. However, that isn't practical to do without breaking all case tiles that are already out in the wild, so for now I'm settling for just renaming `style` to `grid-style`. The parser will still accept an element name of `style` OR `grid-style` in order to preserve backwards-compatibility, but I'm going to call it `grid-style` in the documentation here https://github.com/dimagi/commcare-core/wiki/Suite20#detail, and try to make sure we use that going forward.

### DIFF
--- a/backend/src/org/commcare/xml/DetailFieldParser.java
+++ b/backend/src/org/commcare/xml/DetailFieldParser.java
@@ -47,7 +47,7 @@ public class DetailFieldParser extends CommCareElementParser<DetailField> {
             }
         }
         if (nextTagInBlock("field")) {
-            parseStyle(builder);
+            parseGridStyle(builder);
             checkNode("header");
 
             builder.setHeaderWidthHint(parser.getAttributeValue(null, "width"));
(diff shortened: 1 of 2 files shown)


backend/src/org/commcare/xml/DetailFieldParser.java
```diff
@@ -84,16 +84,15 @@ public DetailField parse() throws InvalidStructureException, IOException, XmlPul
...
 
-    private void parseStyle(DetailField.Builder builder) throws InvalidStructureException, IOException, XmlPullParserException {
-        //style
-        if (parser.getName().toLowerCase().equals("style")) {
-            StyleParser styleParser = new StyleParser(builder, parser);
-            styleParser.parse();
-            //Header
+    private void parseGridStyle(DetailField.Builder builder) throws InvalidStructureException, IOException, XmlPullParserException {
+        if (parser.getName().toLowerCase().equals("style") ||
+                parser.getName().toLowerCase().equals("grid-style")) {
```
amstone326: updating the parser to accept either `style` or `grid-style` as the element name

PR #381, opened 2016-08-29
Extend case list actions to allow for relevancy condition

phillipm:
Add `relevant=xpath_expr` to case list actions, allowing them to be hidden for certain mobile workers.

```
<action relevant="true()">
    <display>
        ...
    </display>
    <stack>
        ...
    </stack>
</action>
```

cross-request: https://github.com/dimagi/commcare-android/pull/1473

### DIFF
--- a/tests/resources/complex_stack/suite.xml
+++ b/tests/resources/complex_stack/suite.xml
@@ -60,6 +60,17 @@
         <title>
             <text>Case List</text>
         </title>
+        <action>
+            <display>
+                <text>Jump to Menu 2 Form 1</text>
+            </display>
+            <stack>
+                <push>
+                    <command value="'m1-f0'"/>
+                    <datum id="calculated_data" value="'new'"/>
+                </push>
+            </stack>
+        </action>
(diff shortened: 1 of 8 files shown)


phillipm:
@proteusvacuum this is ready from the mobile side

tests/resources/complex_stack/suite.xml
```diff
@@ -56,6 +56,49 @@
...
+            <display>
+                <text>Jump to Menu 2 Form 1</text>
+            </display>
+            <stack>
+                <push>
+                    <command value="'m1-f0'"/>
+                    <datum id="calculated_data" value="'new'"/>
+                </push>
+            </stack>
+        </action>
```
amstone326: @phillipm Would be nice to add an action here that doesn't have a relevancy condition at all and make sure it gets included

amstone326:
Good to go after that test condition is added

phillipm:
@amstone326 updated

proteusvacuum:
Thanks!

PR #382, opened 2016-08-29
Show case name for saved follow-up forms when there isn't a case detail screen

phillipm:
Looks like the calculation that loads the case name for saved forms that update a case couldn't handle the case where there wasn't a case detail screen for the case list. Don't think there was any good reason for this.

Fix for http://manage.dimagi.com/default.asp?235678

### DIFF
--- a/backend/src/org/commcare/util/FormDataUtil.java
+++ b/backend/src/org/commcare/util/FormDataUtil.java
@@ -2,7 +2,6 @@ package org.commcare.util;
 
 import org.commcare.cases.model.Case;
 import org.commcare.core.interfaces.UserSandbox;
-import org.commcare.modern.util.Pair;
 import org.commcare.session.CommCareSession;
 import org.commcare.suite.model.ComputedDatum;
 import org.commcare.suite.model.EntityDatum;
@@ -13,6 +12,9 @@ import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.TreeReference;
 
 /**
+ * Use the session state descriptor attached to saved forms to load case
+ * information, such as the case name.
+ *
  * @author Phillip Mates (pmates@dimagi.com)
  */
 public class FormDataUtil {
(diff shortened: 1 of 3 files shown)


backend/src/org/commcare/util/FormDataUtil.java
```diff
@@ -48,12 +48,16 @@ public static String getTitleFromSession(UserSandbox userSandbox,
...
                 datumValue = session.getPoppedStep().getValue();
-            } else if (datum instanceof EntityDatum && ((EntityDatum)datum).getLongDetail() != null) {
+            } else if (datum instanceof EntityDatum) {
                 String tmpDatumValue = session.getPoppedStep().getValue();
                 if (tmpDatumValue != null) {
                     datumValue = tmpDatumValue;
                 }
-                return new Pair<>((EntityDatum)datum, datumValue);
+                if (((EntityDatum)datum).getLongDetail() == null) {
+                    return new Pair<>(null, datumValue);
```
amstone326: I'm not following why we can't return the `EntityDatum` here just because it doesn't have a long detail

phillipm: When there isn't a long detail we don't want to fire https://github.com/dimagi/commcare-core/pull/382/files/6d0f6bf1ae7ee116b0d3d2cf00fb24ece89a834e#diff-ff89d6b08ec41ebd600e0eefe0d93ebfR36 but rather https://github.com/dimagi/commcare-core/pull/382/files/6d0f6bf1ae7ee116b0d3d2cf00fb24ece89a834e#diff-ff89d6b08ec41ebd600e0eefe0d93ebfR33 to get the name associated with a case

amstone326: Gotcha. I would advocate for adding some sort of comment here to indicate that, and maybe renaming the method a bit as well. With the method just called `getDatumAndCaseId`, it appears counterintuitive that you wouldn't return the datum just because it's missing a long detail

backend/src/org/commcare/util/FormDataUtil.java
```diff
@@ -13,55 +12,46 @@
...
-                    return new Pair<>(null, datumValue);
+                    // In the absence of a case detail, use the plain case name as the title
+                    break;
                 } else {
-                    return new Pair<>((EntityDatum)datum, datumValue);
+                    return loadTitleFromEntity((EntityDatum)datum, datumValue, evalContext, sessionCopy, userSandbox);
                 }
             }
-            session.popStep(evaluationContext);
+            sessionCopy.popStep(evalContext);
```
amstone326: what is the purpose of this line?

backend/src/org/commcare/util/FormDataUtil.java
```diff
@@ -13,55 +12,46 @@
...
                 // session.getPoppedStep().getId().startsWith("case_id")
-                datumValue = session.getPoppedStep().getValue();
+                datumValue = sessionCopy.getPoppedStep().getValue();
             } else if (datum instanceof EntityDatum) {
-                String tmpDatumValue = session.getPoppedStep().getValue();
+                String tmpDatumValue = sessionCopy.getPoppedStep().getValue();
                 if (tmpDatumValue != null) {
                     datumValue = tmpDatumValue;
                 }
                 if (((EntityDatum)datum).getLongDetail() == null) {
```
amstone326: can this if/else statement be re-written as just `if (((EntityDatum)datum).getLongDetail() != null) { return loadTitleFromEntity((EntityDatum)datum, datumValue, evalContext, sessionCopy, userSandbox);`, and remove the oddness of the `break`?

backend/src/org/commcare/util/FormDataUtil.java
phillipm: The algorithm steps backwards through the session until it finds a case id to use

PR #383, opened 2016-08-31
Fix for Crash Querying Non-Homogenous Fixtures

ctsims:
Bug: http://manage.dimagi.com/default.asp?233467#1196349

Fixes issue where querying a tree with attributes that aren't consistent would crash.


ctsims:
@phillipm Moved tests here and removed X-request

phillipm:
Nice, thanks!

PR #387, opened 2016-09-01
Add computeds to stack

wpride:
This broke a Formplayer test that I only noticed when I merged us up to master today. Definitely seems correct to add this, though I'm loathe to open this can of forms again.


phillipm:
Do you think this should go into 2.30 as well?

wpride:
@phillipm I think yes since we haven't started QA yet, will remake

wpride:
better addressed in https://github.com/dimagi/commcare-core/pull/403, closing

PR #391, opened 2016-09-06
Make max and min work on datestrings

amstone326:
For the 2.30 PR, I'm only extending max() and min() to accept date-strings, but we should talk about making all uses of `toNumeric()` behave this way (Possible that it's totally straightforward, but I wanted to make the minimum amount of changes necessary for 2.30 here)

### DIFF
--- a/javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -32,6 +32,7 @@ import org.javarosa.xpath.parser.XPathSyntaxException;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.text.DateFormat;
 import java.util.Date;
 import java.util.Enumeration;
 import java.util.Hashtable;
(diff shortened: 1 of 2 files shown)


javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -660,6 +659,20 @@ public static Double toDouble(Object o) {
...
     }
 
+    private static Double toNumeric_inclusiveOfDatestrings(Object o) {
+        Double d = toNumeric(o);
+        if (Double.isNaN(d.doubleValue())) {
+            o = unpack(o);
+            if (o instanceof String) {
+                Date dateFromString = DateUtils.parseDate((String)o);
+                if (dateFromString != null) {
+                    return new Double(DateUtils.daysSinceEpoch(dateFromString));
```
ctsims: For cohesiveness, can we make sure that the process by which a string turns into a date, and a date turns into a double are following the same code paths as they would be normally? IE: seems like we should be doing 

if(NaN) { 
try
toNumeric(toDate(val))
catch()
}

So that if there's any complexity (your unpack() call is a good example), that complexity is compartmentalized.

ctsims:
Looks like the right approach in general. Left a comment about general compartmentalization that would be good to check

amstone326:
@ctsims Hey, I changed this as much as I understood was possible, which was just to replace `return new Double(DateUtils.daysSinceEpoch(dateFromString));` with `return toNumeric(dateFromString);`. As far as I can tell, the call to `unpack()` and the subsequent conversion from `String` to `Date` are still unavoidable, since the point of implementing this method was that `toNumeric` doesn't currently convert strings to dates. Let me know if you were thinking of something more than this though

ctsims:
@amstone326 Pulling for now. We should follow up about how to address post-2.30

PR #393, opened 2016-09-07
Wsp/delete invalidstoragestructure

wpride:
Follow up to https://github.com/dimagi/commcare-android/pull/1485 for master, I can't really see a use for having these be separate exceptions. Proposing to merge two the unless - @phillipm was there a specific reason we needed both? Only thing that jumps out was that InvalidStorageStructureException was a Runtime while InvalidStructureException is not, though I couldn't see a place where that was useful. 

Cross https://github.com/dimagi/commcare-android/pull/1486


phillipm:
Sounds good! Not exactly sure why I added it, maybe to have a distinct error for storage, not xml structure issues. Regardless, consolidating seems good.

PR #394, opened 2016-09-07
Handle datestrings in toNumeric

amstone326:
Extension of the work in https://github.com/dimagi/commcare-core/pull/391 to make all usages of `toNumeric` accept datestrings.


javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -687,22 +673,16 @@ public static Double toNumeric(Object o) {
...
             double d;
             try {
-                s = s.trim();
-                for (int i = 0; i < s.length(); i++) {
-                    char c = s.charAt(i);
-                    if (c != '-' && c != '.' && (c < '0' || c > '9'))
-                        throw new NumberFormatException();
+                checkForInvalidNumericCharacters(s);
+                try {
+                    val = attemptDateConversion(s);
```
ctsims: I think we should try to parse to a double _first_ (which is a much cheaper operation and much more common) before falling back to the attempted date conversion.

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -687,22 +673,16 @@ public static Double toNumeric(Object o) {
...
-            String s = (String)o;
+            String s = ((String)o).trim();
             double d;
             try {
-                s = s.trim();
-                for (int i = 0; i < s.length(); i++) {
-                    char c = s.charAt(i);
-                    if (c != '-' && c != '.' && (c < '0' || c > '9'))
-                        throw new NumberFormatException();
+                checkForInvalidNumericCharacters(s);
```
ctsims: Somewhat awkwardly it's probably for the best if this is "checkforInvalidNumericOrDateStringCharacters()" since both are necessary for the code below to make sense. Otherwise someone is likely to think the date checking is out of order.

javarosa/src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
```diff
@@ -172,6 +172,14 @@ public void doTests() {
         testEval("number(date('1969-12-31'))", null, null, new Double(-1.0));
         testEval("number(date('2008-09-05'))", null, null, new Double(14127.0));
         testEval("number(date('1941-12-07'))", null, null, new Double(-10252.0));
+        testEval("number('1970-01-01')", null, null, new Double(0.0));
+        testEval("number('1970-01-02')", null, null, new Double(1.0));
+        testEval("number('1969-12-31')", null, null, new Double(-1.0));
+        testEval("number('2008-09-05')", null, null, new Double(14127.0));
+        testEval("number('1941-12-07')", null, null, new Double(-10252.0));
+        testEval("number('1970-01')", null, null, new Double(Double.NaN));
```
ctsims: I'm curious, what happens when we try to cast over a "date looking" date that just doesn't exist, like "2016-13-13" or something? (Just trying to think of areas where we might end up with uncaptured exceptions)

amstone326: Added 2 tests below- they evaluate to NaN as I assume we would want

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
amstone326: Good call, changed

javarosa/src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
```diff
@@ -68,6 +68,8 @@ private void testEval(String expr, FormInstance model, EvaluationContext ec, Obj
                     fail("Doubles outside of tolerance [" + o + "," + t + " ]");
                 } else if (Double.isNaN(o) && !Double.isNaN(t)) {
                     fail("Result was NaN when not expected");
+                } else if (Double.isNaN(t) && !Double.isNaN(o)) {
```
amstone326: just added this because I realized we were not actually checking for this failure case... @ctsims Are there other test files where I might want to check that we were overlooking this?

amstone326:
hold off on merge, still fixing 1 other thing

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -674,11 +674,11 @@ public static Double toNumeric(Object o) {
             val = (Double)o;
         } else if (o instanceof String) {
             String s = ((String)o).trim();
-            double d;
```
amstone326: The change in this commit seemed like the cleanest way to make sure we don't attempt a date conversion on a string we already know has invalid characters. Otherwise both the `checkForInvalidNumericOrDatestringCharacters()` method and the `Double.parseDouble()` call were both potentially throwing a `NumberFormatException`, with no way to distinguish between them.

amstone326:
ok, g2g

javarosa/src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
ctsims: Good catch, not sure, I think we evaluate equality on these internally when we do xpath expressions, so I can't think of any.

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -720,6 +700,28 @@ public static Double toNumeric(Object o) {
...
+        }
+        return false;
+    }
+
+    private static Double attemptDateConversion(String s) {
+        if (s.length() == 0) {
+            // This check is needed because toDate() returns a string if given the empty string
+            throw new XPathTypeMismatchException();
+        }
+        return toNumeric(toDate(s));
```
ctsims: Hm, since toDate() here isn't guaranteed to return a date, I think we should probably manually check the return type. You caught one of the special cases above, but any others that got added to the toDate() method may not be mirrored back here, which would be a tough infinite loop to catch.

amstone326: makes sense, updated

ctsims:
This looks great. I think this will make a lot of people's lives way easier. Thanks for the fine detail work, Aliza!

amstone326:
Woo! Do you think this is something that is worth announcing explicitly to implementers in some way?

ctsims:
It'll be off-by-one release (2.31 v. 2.30), but I'd add a tag for a release note, yeah.

amstone326:
Release Note: All xpath functions that operate on numeric values now work directly upon date-strings (i.e. "2012-05-13"), rather than having to cast the string to date first.

PR #395, opened 2016-09-08
Properly consider group relevancy condition in a repeat while creating triggerable DAG

phillipm:
There was an issue where a relevancy condition attached to a group inside of a repeat wasn't being considered when building the triggerable DAG that controls when/how data propagates through the form. This was because the condition was inside a template (because it was in a repeat) and the code that cascades dependencies wasn't correctly taking templates into account.

Fix for http://manage.dimagi.com/default.asp?236848


javarosa/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -774,16 +777,16 @@ private void fillTriggeredElements(Triggerable t,
...
+                                                           ArrayList<TreeReference> updatedNodes) {
+        ArrayList<TreeReference> cachedNodes = cachedCascadingChildren.retrieve(target);
         if (cachedNodes == null) {
             if (target.getMultLast() == TreeReference.INDEX_ATTRIBUTE) {
                 // attributes don't have children that might change under
                 // contextualization
                 cachedCascadingChildren.register(target, updatedNodes);
             } else {
-                Vector<TreeReference> expandedRefs = exprEvalContext.expandReference(target);
+                Vector<TreeReference> expandedRefs = exprEvalContext.expandReference(target, true);
```
phillipm: The fix, the rest is linting and cleanup.

tests/test/org/commcare/test/utilities/MockApp.java
```diff
@@ -30,23 +33,37 @@ public MockApp(String resourcePath) throws Exception {
...
+        mSandbox.setLoggedInUser(mSandbox.getUserStorage().read(0));
 
         mSessionWrapper = new SessionWrapper(mEngine.getPlatform(), mSandbox);
     }
 
+    /**
+     * Loads the provided form and properly initializes external data instances,
+     * such as the casedb and commcare session.
+     */
+    public FormParseInit loadAndInitForm(String formFileInApp) {
```
phillipm: Not used now that I found a minimal test case, but it was helpful in copying a form and user restore from HQ and loading all that data into a `commcare-core` level test. Leaving it here in hopes that we use it soon. Also willing to delete it if that sounds better.

phillipm: Not used now that I found a minimal test case, but it was helpful in copying a form and user restore from HQ and loading all that data into a `commcare-core` level test. Leaving it here in hopes that we use it soon. Also willing to delete it if that sounds better.

PR #396, opened 2016-09-12
Allow detail to tell entity list to focus to bottom

amstone326:


backend/src/org/commcare/suite/model/Detail.java
```diff
@@ -65,6 +65,10 @@
     // Force the activity that is showing this detail to show itself in landscape view only
     private boolean forceLandscapeView;
 
+    // If this detail is being shown in an entity list, set focus to the LAST item in the list
+    // when it is loaded, instead of the first
+    private boolean focusToBottomOfEntityList;
```
phillipm: Do you want to save/restore this in the `Externalizable` framework?

amstone326: Hm, it seems like I should be doing that for all of the custom fields I added here recently, but they have definitely been working without that, which is odd

phillipm: I'm not entirely sure why any of the suite stuff is externalizable. We re-parse it from xml every time the app launches.

amstone326: yea that sounds right. can we kill it then?

ctsims: It's old J2ME Stuff. Will may still be using it, actually, I'd check with
them before killing.

-Clayton

amstone326: @wpride Can you speak to whether this is still used in CloudCare?

wpride: Yeah we actually need it

PR #399, opened 2016-09-12
Fix for Nested Repeats with different lengths

ctsims:
This fixes a bug using nested repeats with counts. 

Previously, an "inner" repeat would trigger the creation of its automated children IFF the count of children of the _outer_ repeat was less than the inner repeat's count.

### DIFF
--- a/javarosa/src/test/java/org/javarosa/core/model/test/FormDefTest.java
+++ b/javarosa/src/test/java/org/javarosa/core/model/test/FormDefTest.java
@@ -212,7 +212,7 @@ public class FormDefTest {
         stepThroughEntireForm(fec);
 
         ExprEvalUtils.assertEqualsXpathEval("Nested repeats did not evaluate to the proper outcome",
-                30.0,
+                60.0,
                 "/data/sum",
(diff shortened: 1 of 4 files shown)


javarosa/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -212,7 +212,7 @@ public void testNestedRepeatActions() throws Exception {
         stepThroughEntireForm(fec);
 
         ExprEvalUtils.assertEqualsXpathEval("Nested repeats did not evaluate to the proper outcome",
-                30.0,
+                60.0,
```
ctsims: This test change expands the repeats out far enough that it triggers the bug. Previously the outer repeat count was always smaller than the inner repeat count.

amstone326:
@ctsims was there a ticket that triggered this or did you just discover it independently?

ctsims:
@amstone326 Discovered independently while testing this form:
https://india.commcarehq.org/a/icds-test/apps/view/8e6ee6e3290ddc3ac114211b8af9d58d/modules-0/forms-0/source/

PR #403, opened 2016-09-15
Fix for going back over case_id after end of form navigation

wpride:
Fixes http://manage.dimagi.com/default.asp?237462

Essentially we weren't evaluating unknowns to case_ids (DATUM_VALS) when they possibly could have been and accepting these as valid user actions/datums.


wpride:
(Cleaning up test resources before adding)

tests/test/org/commcare/backend/session/test/StackRegressionTests.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by willpride on 9/15/16.
+ */
+
+public class StackRegressionTests {
+    /**
+     * Load form title from session where the case id is computed, not
```
amstone326: Looks like this never got updated after copying from elsewhere

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -355,6 +355,13 @@ private String guessUnknownType(StackFrameStep popped) {
                 if (childCommand.equals(poppedId)) {
                     return SessionFrame.STATE_COMMAND_ID;
                 }
```
amstone326: Can you update the method comment to include the logic you added here?

backend/src/org/commcare/session/SessionDescriptorUtil.java
```diff
@@ -16,7 +16,8 @@ public static void loadSessionFromDescriptor(String sessionDescriptor,
             if (action.equals(SessionFrame.STATE_COMMAND_ID)) {
                 session.setCommand(tokenStream[++current]);
             } else if (action.equals(SessionFrame.STATE_DATUM_VAL) ||
-                    action.equals(SessionFrame.STATE_DATUM_COMPUTED)) {
```
amstone326: What is this change for?

tests/test/org/commcare/backend/session/test/StackRegressionTests.java
```diff
@@ -0,0 +1,37 @@
...
+     * selected, and the case name is loaded from detail referenced by m0-f0,
+     * which is implicitly referenced
+     */
+    @Test
+    public void testGoBackAfterEndOfFormNavigation() throws Exception {
```
amstone326: Can you explain how this is testing an end of form nav scenario? It looks to me like all that happens before the `stepBack()` commands is selecting a module and a case.

backend/src/org/commcare/session/SessionDescriptorUtil.java
wpride: We want to allow STATE_UNKNOWN type actions to be instantiated from the descriptor string in the same way that computed and datums are

tests/test/org/commcare/backend/session/test/StackRegressionTests.java
wpride: This is the state that the given form would end up in after end of form navigation, so we want to test that in this situation it would behave correctly (IE determine that the case_id value would have been added by the case selection screen and properly resolve the unknown datum as such).

backend/src/org/commcare/session/CommCareSession.java
wpride: udpated

tests/test/org/commcare/backend/session/test/StackRegressionTests.java
amstone326: Gotcha ok, can you update the method comment to explain that?

wpride: Done

PR #404, opened 2016-09-15
Convert with timezone

benrudolph:
@wpride date times crashed on prod cause browser was in a different timezone than the server


wpride:
oh god here we go with the timezones on Formplayer

benrudolph:
😖 😵

PR #405, opened 2016-09-16
Use safe double parse function

wpride:
Fixes http://manage.dimagi.com/default.asp?237000

This is without a doubt the most obscure bug I've fixed in my lifetime. 

Basically way deep down in the XForm engine we have this code that attempts to parse as a double every string literal we process and, if successful, stores the value as in a CacheTable with the parsed double as the value and the string as a key. This includes things like fixture ID's that we use for indexing and later lookups. 

Unfortunately, until version 4.4 Android had a bug with the parseDouble function that you can read about here: https://android-review.googlesource.com/#/c/102376/. Because Java doubles use the exponential format MeP the "e" character could be in a valid double in certain circumstances. In the bug above case, the fixture ID was 1d9098a0c23a0c83740547dd946**e3783**

So, the bug would parse the trailing 946e3783 as a valid exponential with mantissa even though the rest of the string was complete garbage. This would evaluate to infinity (or Double.POSITIVE_INFINITY to be precise) and then be stored in the cache table as a valid double mapping.

Then, all later lookups would use this value when trying to match against the filter expression in the XForm, which would of course fail.

This fix simply uses code from SO to parse the double and is the fastest possible (according to SO). Another smaller-touch fix would be to manually check for and reject the infinity value after parsing. 

There's also a great post in Russian about the original discovery of the bug https://habrahabr.ru/post/234193/

### DIFF
--- a/javarosa/src/main/java/org/javarosa/core/util/MathUtils.java
+++ b/javarosa/src/main/java/org/javarosa/core/util/MathUtils.java
@@ -28,7 +28,6 @@ public class MathUtils {
     }
 
     // double parsing from http://stackoverflow.com/questions/8564896/fastest-way-to-check-if-a-string-can-be-parsed-to-double-in-java
-
     final static String Digits     = "(\\p{Digit}+)";
     final static String HexDigits  = "(\\p{XDigit}+)";
 
@@ -59,7 +58,7 @@ public class MathUtils {
 
                     // Hexadecimal strings
                     "((" +
-                    // 0[xX] HexDigits ._opt BinaryExponent FloatTypeSuffix_opt
+                    // 0[xX] HlxDigits ._opt BinaryExponent FloatTypeSuffix_opt
                     "(0[xX]" + HexDigits + "(\\.)?)|" +
 
                     // 0[xX] HexDigits_opt . HexDigits BinaryExponent FloatTypeSuffix_opt
@@ -69,6 +68,7 @@ public class MathUtils {
                     "[fFdD]?))" +
                     "[\\x00-\\x20]*");// Optional trailing "whitespace"
 
+    // Safe double parsing - see PR https://github.com/dimagi/commcare-core/pull/405
     public static Double parseDoubleSafe(String string) {
         if (Pattern.matches(fpRegex, string))
             return Double.valueOf(string);
--- a/javarosa/src/main/java/org/javarosa/core/util/MathUtils.java
+++ b/javarosa/src/main/java/org/javarosa/core/util/MathUtils.java
@@ -27,6 +27,8 @@ public class MathUtils {
         return r;
     }
 
+    // double parsing from http://stackoverflow.com/questions/8564896/fastest-way-to-check-if-a-string-can-be-parsed-to-double-in-java
+
     final static String Digits     = "(\\p{Digit}+)";
     final static String HexDigits  = "(\\p{XDigit}+)";
 
--- a/javarosa/src/main/java/org/javarosa/core/util/MathUtils.java
+++ b/javarosa/src/main/java/org/javarosa/core/util/MathUtils.java
@@ -1,7 +1,6 @@
 package org.javarosa.core.util;
 
 import java.util.Random;
-import java.util.regex.Pattern;
 
 /**
  * Static utility functions for mathematical operations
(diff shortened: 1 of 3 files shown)


czue:
haha, wow this is amazing. i didn't even know you spoke russian.

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1449,7 +1449,7 @@ public static Object InferType(String attrValue) {
         }
 
         try {
-            Double ret = new Double(Double.parseDouble(attrValue));
+            Double ret = MathUtils.parseDoubleSafe(attrValue);
```
czue: are there any performance implications of this? just wondering if we should only use our custom thing conditionally on android version or something.

javarosa/src/main/java/org/javarosa/core/util/MathUtils.java
```diff
@@ -25,4 +26,54 @@ public static Random getRand() {
         }
         return r;
     }
+
+    // double parsing from http://stackoverflow.com/questions/8564896/fastest-way-to-check-if-a-string-can-be-parsed-to-double-in-java
```
czue: might also be worth linking to the ticket or PR to explain why this exists

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
wpride: This is at the Javarosa level so would be hard to make contingent on Android version. According to SO and my understanding of regex's this should actually be more performant. Maybe @ctsims can weigh in - can also just manually check for the Infinity value if we think the parseDouble function is more performant.

amstone326:
@wpride Doesn't this need to be hotfixed if enikshay needs it?

wpride:
@amstone326 no Cory had confirmed it does not since it only affects one user

czue:
(we're going to just give them an apk)

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
amstone326: Maybe we could try profiling this to be sure?

amstone326: Although if SO says it's more performant I'm not that worried

phillipm: The results of this evaluation are cached, so hopefully that would help soften any adverse performance effects. Would be cool to test this vs. the infinity check on a few thousand unique ID strings though.

phillipm:
> This would evaluate to infinity (or Double.POSITIVE_INFINITY to be precise)
Are you sure it will always be positive infinity? 
Also, do we want to be casting things that or IDs but also valid double to doubles? Like what happens if we have an ID it is just `100E5`? Do we actually want to consider it a double in `tryBatchChildFetch`?

wpride:
@amstone326 @phillipm @ctsims ok so some back of the envelope performance testing using this test https://gist.github.com/wpride/d36082328832ecab52b8b1d90b22b24d that would always miss the cache with 100000 successful double casts and 100000 unsuccessful. The old code ran this in .401 seconds and the new code in 1.867 - so, looks like there's a non-trivial performance hit. (Note this ran on my machine, not a device - don't know of any reason to think it'd be much different?)

Given that, we should probably stick with the old parseDouble function and handle the Infinity and Negative Infinity values deliberately. Seems to me that we probably never want to actually treat "Infinity" as a double in the cache table.

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1449,7 +1449,12 @@ public static Object InferType(String attrValue) {
         }
 
         try {
-            Double ret = new Double(Double.parseDouble(attrValue));
+            // Don't process strings with scientific notation or +/- Infinity as doubles
+            if (checkForInvalidNumericOrDatestringCharacters(attrValue)) {
```
phillipm: Nice!

PR #407, opened 2016-09-19
Serialize and restore SessionFrame stack

wpride:
Needed for Formplayer


amstone326:
@phillipm Will this have any implications for how session restoration currently works? I vaguely thought that there was a reason the restoration code didn't already include the frame stack.

phillipm:
As far as I remember it was just that I didn't get around to implementing it at the time.

PR #409, opened 2016-09-19
Fix for '+' being reserved in SQL, test fixes

wpride:
cc @benrudolph

### DIFF
--- a/modern/org/commcare/modern/database/TableBuilder.java
+++ b/modern/org/commcare/modern/database/TableBuilder.java
@@ -180,7 +180,7 @@ public class TableBuilder {
 
     //sqlite doesn't like dashes
     public static String scrubName(String input) {
-        return input.replace("-", "_");
+        return input.replace("-", "_").replace(".", "_");
     }
 
     public static byte[] toBlob(Externalizable externalizable){
--- a/modern/org/commcare/modern/database/TableBuilder.java
+++ b/modern/org/commcare/modern/database/TableBuilder.java
@@ -180,7 +180,7 @@ public class TableBuilder {
 
     //sqlite doesn't like dashes
     public static String scrubName(String input) {
-        return input.replace("-", "dash").replace(".", "dot").replace("+", "plus");
+        return input.replace("-", "_dash_").replace(".", "_dot_").replace("+", "_plus_");
     }
 
     public static byte[] toBlob(Externalizable externalizable){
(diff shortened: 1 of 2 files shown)


modern/org/commcare/modern/database/TableBuilder.java
```diff
@@ -180,7 +180,7 @@ public String getTableCreateString() {
 
     //sqlite doesn't like dashes
     public static String scrubName(String input) {
-        return input.replace("-", "_");
+        return input.replace("-", "_dash_").replace(".", "_dot_").replace("+", "_plus_");
```
phillipm: Pretty sure this will require a bunch of data migrations. For example, this is used to get db column names in https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/database/SqlStorage.java#L152

phillipm: Oh, I guess if you have verified that no current scrubbed columns etc use `-`, `.`, or `+`, then we should be good.

tests/test/org/commcare/util/test/SqlStorageIndexedTests.java
```diff
@@ -150,13 +150,13 @@ public void testSqlLedgerStorage() {
             ledgerStorage.write(l2);
             ledgerStorage.write(l3);
 
-            Vector ids = ledgerStorage.getIDsForValue("entity_id", "ledger_entity_id");
+            Vector ids = ledgerStorage.getIDsForValue("entity-id", "ledger_entity_id");
```
phillipm: Can you explain why this change needed?

modern/org/commcare/modern/database/TableBuilder.java
benrudolph: I would've expected android to crash if it did have usernames like that. though it seems plausible that it would have come across `.` in usernames. Do you know if Android's sqlite version is just able to handle that? Or is there some other escaping process

benrudolph: Ah nvm, you're saying it's used for column names (not username) I would expect that to fail

phillipm: Yeah, and it would only crash if you try to update CommCare with an app installed from `2.30` to the dev build. Otherwise the tables would just get created with the new naming structure.

PR #410, opened 2016-09-19
Move getQuestionPrompts into FormEntryController

wpride:
Needed for live preview

cross-request: https://github.com/dimagi/commcare-android/pull/1495

cc @benrudolph

### DIFF
--- a/javarosa/src/main/java/org/javarosa/form/api/FormEntryController.java
+++ b/javarosa/src/main/java/org/javarosa/form/api/FormEntryController.java
@@ -29,6 +29,8 @@ public class FormEntryController {
     public static final int EVENT_REPEAT = 16;
     public static final int EVENT_REPEAT_JUNCTURE = 32;
 
+    public final static String FIELD_LIST = "field-list";
+
     private final FormEntryModel model;
     private final FormEntrySessionRecorder formEntrySession;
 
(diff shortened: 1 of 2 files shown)


javarosa/src/main/java/org/javarosa/form/api/FormEntryController.java
```diff
@@ -369,4 +374,77 @@ public void setLanguage(String language) {
...
+            return new FormEntryPrompt[]{this.getModel().getQuestionPrompt(currentIndex)};
+        }
+    }
+
+    /**
+     * A convenience method for determining if the current FormIndex is a group that is/should be
+     * displayed as a multi-question view of all of its descendants. This is useful for returning
+     * from the formhierarchy view to a selected index.
+     */
+    private boolean isFieldListHost(FormIndex index) {
```
amstone326: This method is now duplicated in `FormController` and `FormEntryController`, right? Can you make the implementation in core static and have the android code call that?

wpride: That's what this PR does basically, right? https://github.com/dimagi/commcare-android/pull/1495/files

Making it static and passing in a FormEntryController doesn't seem particularly cleaner to me

wpride: Ignore previous comment - see you were pointing to a different method than the main one

PR #413, opened 2016-09-25
Add 'cond' function to JavaRosa XPath engine

phillipm:
Our `if(bool_expr, expr, expr)` syntax can be quite cumbersome for complicated conditional statements. For example, if you have several mutually exclusive conditional branches, you need to create nested `if`-statements, which can be hard to read.

This PR introduces `cond(bool_expr, expr, bool_expr, expr, ..., expr)` to the language.

For example, here is some XPath logic I wrote recently:
```
date(
  if(member_profile = 'ecd', next_ecd_visit_date,
    if(member_profile = 'hiv_exp_infant', next_visit_hiv_exp_infant,
      if(member_profile = 'hiv_pos_infant', next_visit_hiv_pos_infant,
        if(member_profile = 'hiv_pos_child', next_visit_hiv_pos_child,
          if(member_profile = 'hiv_exp_child', next_visit_hiv_exp_child,
            if(member_profile = 'hiv_unexp_child', next_visit_hiv_unexp_child,
              if(member_profile = 'hiv_pos_not_on_arts', next_visit_hiv_pos_not_on_arts, next_visit_date_max))))))))
```
And now with `cond`:
```
date(
  cond(member_profile = 'ecd', next_ecd_visit_date,
       member_profile = 'hiv_exp_infant', next_visit_hiv_exp_infant,
       member_profile = 'hiv_pos_infant', next_visit_hiv_pos_infant,
       member_profile = 'hiv_pos_child', next_visit_hiv_pos_child,
       member_profile = 'hiv_exp_child', next_visit_hiv_exp_child,
       member_profile = 'hiv_unexp_child', next_visit_hiv_unexp_child,
       member_profile = 'hiv_pos_not_on_arts', next_visit_hiv_pos_not_on_arts,
       next_visit_date_max))
```

__Release Notes__ Add 'cond()' function to our XPath engine as an alternative to nested if-statements.


amstone326:
Seems like it would be a huge improvement to me. One question- is there any reason it's necessary to require an ":else" statement?

phillipm:
@amstone326 we definitely need a default case unless we want to have the `cond()` return `""` when none of the conditional clauses are true, which sounds not great.

The `:else` is not needed. My reason for adding it is because it helps distinguish that the last argument is the default case.. and I've been hanging out with Clojure, which has this syntax. We could leave it out, but it messes with the visual symmetry a bit :)

amstone326:
Ah gotcha yea no I think that makes the most sense then

emord:
@phillipm is `:else` valid xpath? Right now the javascript xpath parser chokes on it, so it would show up as an error in vellum

phillipm:
@emord updated to exclude `:else` so that vellum won't need any parser changes

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -59,10 +59,11 @@ public XPathFuncExpr() {
...
     public XPathFuncExpr(XPathQName id, XPathExpression[] args) throws XPathSyntaxException {
-
-        if (id.name.equals("if")) {
-            if (args.length != 3) {
-                throw new XPathSyntaxException("if() function requires 3 arguments but " + args.length + " are present.");
+        if ("if".equals(id.name) && args.length != 3) {
+            throw new XPathSyntaxException("if() function requires 3 arguments but " + args.length + " are present.");
+        } else if ("cond".equals(id.name)) {
+            if (args.length < 3 || args.length % 2 != 1) {
+                throw new XPathSyntaxException("cond() function requires an odd number of arguments. " + args.length + " are present.");
```
amstone326: Technically someone could have tried to call `cond()` with 1 argument here, and then the error message would be confusing. Probably best to split up the 2 checks and have a different error message for `args.length < 3`, even though it's unlikely

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -873,10 +878,23 @@ private Double position(TreeReference refAt) {
         return new Double(refAt.getMultLast());
     }
 
-    public static Object ifThenElse(DataInstance model, EvaluationContext ec, XPathExpression[] args, Object[] argVals) {
```
amstone326: Do you understand why this weird `argVals` param was being used? It's such an odd pattern that I'm afraid there was some obscure reason for it

javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -873,10 +878,23 @@ private Double position(TreeReference refAt) {
...
-        return (b ? args[1].eval(model, ec) : args[2].eval(model, ec));
+    private static Object ifThenElse(DataInstance model, EvaluationContext ec, XPathExpression[] args) {
+        if (toBoolean(args[0].eval(model, ec))) {
+            return args[1].eval(model, ec);
+        } else {
+            return args[2].eval(model, ec);
+        }
+    }
+
+    private static Object condEval(DataInstance model, EvaluationContext ec,
```
amstone326: this is sweet!

phillipm:
@amstone326 updated

PR #416, opened 2016-09-28
Don't clear stack if rewind occurred when terminating session

phillipm:
Follow up to https://github.com/dimagi/commcare-core/pull/368

Turns out that mark/rewind for case claim session nav didn't work in normal contexts when there wasn't another frame being popped. I think this came up now because I only tested the mark/rewind implementation on the enikshay app.


phillipm:
Self-merging, will re-ping next week for a real review.

PR #418, opened 2016-10-02
don't swallow error message when writing blob

wpride:
Fix dumb error handling

### DIFF
--- a/modern/org/commcare/modern/database/TableBuilder.java
+++ b/modern/org/commcare/modern/database/TableBuilder.java
@@ -187,9 +187,9 @@ public class TableBuilder {
         ByteArrayOutputStream bos = new ByteArrayOutputStream();
         try {
             externalizable.writeExternal(new DataOutputStream(bos));
-        } catch (IOException e1) {
-            e1.printStackTrace();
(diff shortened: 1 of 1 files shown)


modern/org/commcare/modern/database/TableBuilder.java
```diff
@@ -187,9 +187,9 @@ public static String scrubName(String input) {
         ByteArrayOutputStream bos = new ByteArrayOutputStream();
         try {
             externalizable.writeExternal(new DataOutputStream(bos));
-        } catch (IOException e1) {
-            e1.printStackTrace();
-            throw new RuntimeException("Failed to serialize externalizable for content values");
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to serialize externalizable " + externalizable +
```
ctsims: Hey,

One thing that helps a _lot_ in cases like this is to not just throw the message, but to include the full trace. The best way to do so is to create the runtime exception, and set its "cause" to the original throwable, which comes with a whole host of benefits. This should definitely be a best-practice when we deal with this kind of error handling.

PR #419, opened 2016-10-03
Update coalesce to accept more than 2 arguments

phillipm:
Sometimes users want to be able to coalesce more than 2 values. This PR extends the [`coalesce` function](https://confluence.dimagi.com/display/commcarepublic/CommCare+Functions#CommCareFunctions-coalesce) to accept more than 2 arguments.

For example:

```
coalesce(#case/child_name, #form/child_name, "unknown")
```

For context, the current documentation for coalesce:

> Useful for choosing which of two values to return.  Will return the non-null/empty value.  If both are not null, will return the first argument. 

__Release Notes__ Allow `coalesce` to accept more than 2 arguments.

### DIFF
--- a/javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -174,7 +174,7 @@ public class XPathFuncExpr extends XPathExpression {
 
         //TODO: Func handlers should be able to declare the desire for short circuiting as well
         if (name.equals("if") && args.length == 3) {
-            return ifThenElse(model, evalContext, args, argVals);
+            return ifThenElse(model, evalContext, args);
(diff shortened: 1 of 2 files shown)


javarosa/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -170,15 +170,8 @@ public Object evalRaw(DataInstance model, EvaluationContext evalContext) {
         //TODO: Func handlers should be able to declare the desire for short circuiting as well
         if (name.equals("if") && args.length == 3) {
             return ifThenElse(model, evalContext, args, argVals);
```
amstone326: only comment is same as on the `cond` PR- are you sure there wasn't a reason for the `argVals` thing?

amstone326:
@phillipm This has conflicts now that I merged https://github.com/dimagi/commcare-core/pull/413

PR #430, opened 2016-10-20
Allow super big strings to be serialized

phillipm:
There is a limitation in our serialization framework where strings over `65535` bytes cannot be serialized. The fix I've implemented checks the length of the string as bytes and dispatches to more robust serialization logic if the string is above the limit. This does not require a migration because of some sneaky uses of tagging. This does come with a limitation that if you try to directly read out a large string that was serialized using `readString`, it doesn't work. On the plus side, in the case of case properties, which are stored using `ExtWrapMapPoly`, everything works because `ExtWrapMapPoly` utilizes tags.

Targeting 2.31, but can change it to master if we think the change is too big for 2.31.

Fix for http://manage.dimagi.com/default.asp?237711

### DIFF
--- a/javarosa/src/main/java/org/javarosa/core/util/externalizable/ExtUtil.java
+++ b/javarosa/src/main/java/org/javarosa/core/util/externalizable/ExtUtil.java
@@ -1,5 +1,6 @@
 package org.javarosa.core.util.externalizable;
 
+import org.javarosa.core.model.data.LargeString;
 import org.javarosa.core.services.PrototypeManager;
 import org.javarosa.core.util.Interner;
 import org.javarosa.core.util.OrderedHashtable;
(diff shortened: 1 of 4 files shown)


javarosa/src/main/java/org/javarosa/core/util/externalizable/ExtUtil.java
```diff
@@ -127,7 +129,14 @@ public static void writeBool(DataOutputStream out, boolean val) throws IOExcepti
     }
 
     public static void writeString(DataOutputStream out, String val) throws IOException {
-        out.writeUTF(val);
+        if (val.getBytes("UTF-8").length > MAX_UNSIGNED_SHORT_VALUE) {
```
phillipm: I'm wondering if this line has bad memory overhead. Is there a way to count the bytes in a string without allocating the whole string into memory again, which is what this is doing, right?

phillipm:
Closing out due to memory / runtime concerns

PR #431, opened 2016-10-20
Show user-facing error when serializing too large of a string

phillipm:
Throw error and wipe form upon form completion if you try to save a case property (or really any case field) that is too large.

screen


amstone326:
@phillipm Curious what prompted this- did a user report an issue related to this?

amstone326:
ah nevermind, just saw the other PR that explains this

amstone326:
@phillipm Is there no better way we can handle this than wiping the user's entire form? Would it not be better to try to save everything besides that value? Or is that just too hard to do to make it worth it?

phillipm:
@amstone326 I think it is hard to be worth it. This is failing in the case commit phase, meaning that the form has already been written with the super large value in it. If we don't wipe the record, then it will get submitted to HQ and re-synced down to the phone, causing more issues later. We could add checks into the form processing code to prevent this, but I think it is best to just fail hard and tell the user so they can adjust their workflow to stop running into this case. 
I think if we invest dev time into this, it should be to make case attachments work, so people have an alternative to storing huge blobs of data into case properties.

amstone326:
@phillipm Ok, that makes sense. I think one thing that would improve this then is a more actionable error message; I don't think the average user will know what they did wrong or how to fix it in the future from the message shown above. Information about the offending case ID and the size of the blob are useful to us when trying to debug it, so it is worth logging those pieces of information or else putting them in a pinned notification, but those values won't mean anything to the user. The user-facing message should probably say something more like "The value saved for case property X is too large and your form could not save. Please try filling it out again with a smaller value for that property."

phillipm:
@amstone326 updated the error message to say:
`One of the property values for the case named 'some case name' is too large (by 20                  %). Please show your supervisor.`

I don't know if it is worth the complexity to move details into a pinned notification, because this should almost never come up unless the app is doing something really advanced, in which case, the people managing the app should have the know-how to act on this message / contact us.

PR #433, opened 2016-10-20
Remove regexp dummy implementation in test file

phillipm:
the `XPathEvalTest` test overrode the `regex` function with a dummy implementation, which is bad because `regex` test cases weren't testing prod code.


phillipm:
implemented in https://github.com/dimagi/commcare-core/pull/444

PR #434, opened 2016-10-25
remove 1mb blob limit from commcare-core

wpride:
@benrudolph @czue @phillipm turns out this "just works", pretty neat. Don't pull until I can rework this check into the android repo

cross-request https://github.com/dimagi/commcare-android/pull/1547

### DIFF
--- a/modern/org/commcare/modern/database/DatabaseHelper.java
+++ b/modern/org/commcare/modern/database/DatabaseHelper.java
@@ -96,9 +96,6 @@ public class DatabaseHelper {
     private static void addDataToValues(HashMap<String, Object> values,
                                         Externalizable e) throws RecordTooLargeException {
         byte[] blob = TableBuilder.toBlob(e);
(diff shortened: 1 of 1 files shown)


wpride:
Actually @phillipm is that check even needed anymore since we'll use the file system for any oversized fixture on Android and now know we don't need it outside Android?

phillipm:
@wpride We have to be a little careful about it because you have to enable the hybrid db/filesystem storage setup manually. Currently it is only enabled for fixtures because I suspect there is some amount of overhead involved using the hybrid system and I don't foresee it being needed elsewhere for the time being.
I sort of elaborate on this issue in the associated ticket:

> One thing to keep in mind is that my fix for Android is only for fixtures, so if you remove that check and then try to store a form, or other data blob greater than 1mb, then it will crash silently. This is probably okay as long as we know what to look for; I remember Will saying it was hard to initially track down the 1mb limitation issue.

PR #440, opened 2016-11-04
Add template arg to remote query result

phillipm:
Fix for http://manage.dimagi.com/default.asp?240813

When we query HQ for a remote instance, we need to know if the results are a set of cases or not. Currently, for the only use case, case claim, query results are always a set of cases, but we want to be flexible to other workflows going forward.

We need this distinction because internally we treat xpath references that look into a case database differently than normal references: In certain contexts, case db lookups shouldn't fail hard if they don't find a case property, because case properties aren't always set.

This was manifesting in xpath errors in the case detail screen shown when choosing a case to claim. 

After this change is merged, you can include the `template="case"` attribute in a `<query url="http://fake.com/patient_search/" storage-instance="patients" template="case">` element to mark it as a query whose results should adhere to the [casedb template structure](https://github.com/dimagi/commcare-core/blob/master/cases/resources/casedb_instance_structure.xml)

cross-request: https://github.com/dimagi/commcare-android/pull/1552


phillipm:
@wpride might want to check that this refactor doesn't mess with https://github.com/dimagi/commcare-core/blob/nimbus/util/src/org/commcare/util/cli/QueryScreen.java#L83-L93

PR #443, opened 2016-11-08
Use java classes to represent and dispatch XPath functions

phillipm:
Redo how we do XPath function dispatch so that at parse time we create different class instances for different functions (instead of a bunch of `XPathFuncExpr` instances that do dispatch at runtime). This pushes runtime dispatch into the JVM, which should hopefully be faster.

In prep to add in-code/repl xpath function documentation for @benrudolph , as well as a method that returns all available functions for FormPlayer code completion!  

cross-request: https://github.com/dimagi/commcare-android/pull/1558


benrudolph:
wow! you're the man! this is a hell of PR too. 103 files changed!

javarosa/src/main/java/org/javarosa/xpath/expr/FunctionUtils.java
```diff
@@ -0,0 +1,364 @@
...
+import org.javarosa.core.model.utils.DateUtils;
+import org.javarosa.core.util.CacheTable;
+import org.javarosa.core.util.MathUtils;
+import org.javarosa.xpath.IExprDataType;
+import org.javarosa.xpath.XPathNodeset;
+import org.javarosa.xpath.XPathTypeMismatchException;
+
+import java.util.Date;
+
+public class FunctionUtils {
```
wpride: Totally subjective but XPathUtils feels more apt to me

javarosa/src/main/java/org/javarosa/xpath/parser/ast/ASTNodeFunctionCall.java
```diff
@@ -29,6 +27,140 @@ public XPathExpression build() throws XPathSyntaxException {
             xargs[i] = args.get(i).build();
         }
 
-        return new XPathFuncExpr(name, xargs);
+        return buildFuncExpr(name.toString(), xargs);
+    }
+
+    private static XPathFuncExpr buildFuncExpr(String name, XPathExpression[] args)
```
wpride: beastly

wpride:
this gets a 👍 from me though the diff is so huge its hard to grok the changes

ctsims:
@wpride @phillipm Have we done that performance analysis test on this code yet? It sits uneasily with me that we pulled a massive non-blocking change to how we are processing forms without having any evaluation of the impact.

PR #446, opened 2016-11-09
Move cli screen related code into own dir for clean formplayer import

phillipm:


wpride:
Thanks @phillipm good call

PR #451, opened 2016-11-10
WIP Implement case claim for the CLI

wpride:
@phillipm I notice that in the android build.gradle file you exclude the httpclient library - is that because it conflicts with the Android version?


phillipm:
@wpride yeah, I think you get some sort of double import error otherwise

amstone326:
@wpride Looks like this has been open for a long time. Do you still want to finish it or can it be closed?

PR #456, opened 2016-11-16
Fix QueryRequestActivityTest test failure

phillipm:
The changes in https://github.com/dimagi/commcare-core/pull/455 caused a sloppy `NullPointerException` catch to fail here https://github.com/dimagi/commcare-core/blob/master/backend/src/org/commcare/session/RemoteQuerySessionManager.java#L49-L51 which caused an `commcare-android` test to fail

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -303,8 +303,11 @@ public class CommCareSession {
      * an entry on the stack
      */
     public SessionDatum getNeededDatum() {
-        Entry entry = getEntriesForCommand(getCommand()).elementAt(0);
-        return getNeededDatum(entry);
+        Vector<Entry> entries = getEntriesForCommand(getCommand());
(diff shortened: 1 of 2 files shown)


backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -303,8 +303,11 @@ private String getDataNeededByAllEntries(Vector<Entry> entries) {
      * an entry on the stack
      */
     public SessionDatum getNeededDatum() {
-        Entry entry = getEntriesForCommand(getCommand()).elementAt(0);
```
phillipm: Exactly, that is why https://github.com/dimagi/commcare-core/blob/master/backend/src/org/commcare/session/RemoteQuerySessionManager.java#L49-L51 was not hit, because it used to trigger an NPE and now it was triggering a different exception.

amstone326: Gotcha, thanks

PR #457, opened 2016-11-16
Method for getting list of base XPath functions

phillipm:
Exposes method that returns a list of all the hard coded XPath functions (i.e. `sin, uuid, date, substring,...`)

For use in Form Player to allow for auto-completing text entry.

The list is hard-coded because I don't like reflection but includes tests that ensure the list is kept up-to-date if functions are added or deleted.

cc @benrudolph 

contains commits from https://github.com/dimagi/commcare-core/pull/443

### DIFF
--- a/javarosa/src/main/java/org/javarosa/xpath/expr/FunctionUtils.java
+++ b/javarosa/src/main/java/org/javarosa/xpath/expr/FunctionUtils.java
@@ -7,76 +7,77 @@ import org.javarosa.xpath.IExprDataType;
 import org.javarosa.xpath.XPathNodeset;
 import org.javarosa.xpath.XPathTypeMismatchException;
 
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 
 public class FunctionUtils {
-    public static final HashMap<Class, String> funcList = new HashMap<>();
+    private static final HashMap<Class, String> funcList = new HashMap<>();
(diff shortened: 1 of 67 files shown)


javarosa/src/main/java/org/javarosa/xpath/expr/FunctionUtils.java
```diff
@@ -0,0 +1,447 @@
...
+        return s.toLowerCase();
+    }
+
+    /**
+     * Get list of base xpath functions
+     *
+     * (Used in formplayer for function auto-completion)
+     */
+    @SuppressWarnings("unused")
+    public static List<String> xPathFuncList() {
```
phillipm: @benrudolph this is the method you will want to use in formplayer

benrudolph: wow this is awesome. thanks for setting this up. fyi @emord we may be able to expose this list via an api call to formplayer if you think it'd be good to have in vellum

benrudolph: one thing for vellum, since users don't have to use the latest commcare version (unlike formplayer) some of the functions may not be available for the version they're using. having versions attached to the functions would be useful, but probably not a huge priority

wpride:
@phillipm yikes at merge conflicts

PR #458, opened 2016-11-16
Expose doc() XPath function in CLI that prints out function documentation

phillipm:
Exposes the contents of the [CommCare Functions documentation webpage](https://confluence.dimagi.com/display/commcarepublic/CommCare+Functions) to CLI users via the `doc()` xpath function (only made available in debugging environments)

For example:

```
> doc('true')
--------------------
true
--------------------
Return:  Returns the boolean value True
Arguments: None
Syntax: true()
Example: You may want to use true() when you have some advanced logic for a display or validation condition.  You could also use them if you want to calculate true for a hidden value.  For example, if(/data/question1 = "yes" and /data/question2 > 30, true(), false())
```

### DIFF
--- a/javarosa/src/main/java/org/javarosa/xpath/expr/XPathFormatDateForCalendarFunc.java
+++ b/javarosa/src/main/java/org/javarosa/xpath/expr/XPathFormatDateForCalendarFunc.java
@@ -50,6 +50,11 @@ public class XPathFormatDateForCalendarFunc extends XPathFuncExpr {
 
     @Override
     public String getDocumentation() {
-        return "?";
+        return getDocHeader()
+                + "Behavior: Converts a Gregorian formatted date to a different calendar format. \n"
+                + "Return: a String of the input date formatted for the specified calendar. \n"
+                + "Arguments: The date to be converted and the format to convert to "
+                    + "(currently 'ethiopian' or 'nepali'). \n"
+                + "Example: format-date-for-calendar('19070-01-1', 'nepali')";
     }
 }
--- a/javarosa/src/main/java/org/javarosa/xpath/expr/XPathFormatDateForCalendarFunc.java
+++ b/javarosa/src/main/java/org/javarosa/xpath/expr/XPathFormatDateForCalendarFunc.java
@@ -50,6 +50,6 @@ public class XPathFormatDateForCalendarFunc extends XPathFuncExpr {
 
     @Override
     public String getDocumentation() {
-        return "";
+        return "?";
     }
 }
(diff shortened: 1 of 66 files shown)


javarosa/src/main/java/org/javarosa/xpath/expr/XPathFormatDateForCalendarFunc.java
```diff
@@ -47,4 +47,9 @@ private static String formatDateForCalendar(Object dateObject, Object format) {
             throw new XPathUnsupportedException("Unsupported calendar type: " + format);
         }
     }
+
+    @Override
+    public String getDocumentation() {
+        return "?";
```
phillipm: @wpride I had trouble finding examples of this function. Can you fill out the documentation here and update [CommCare Functions confluence page](https://confluence.dimagi.com/display/commcarepublic/CommCare+Functions)?

javarosa/src/main/java/org/javarosa/xpath/expr/XPathPositionFunc.java
```diff
@@ -54,4 +54,9 @@ private static Double position(TreeReference refAt) {
         return new Double(refAt.getMultLast());
     }
 
+
+    @Override
+    public String getDocumentation() {
+        return "?";
```
phillipm: I don't really know all the intricacies of the `position` function and there is no documentation for it on the CommCare functions confluence page. If anyone else wants to take a crack at documenting it, that would be great

benrudolph:
😂

javarosa/src/main/java/org/javarosa/xpath/expr/XPathPositionFunc.java
wpride: @ctsims took my best stab at this

javarosa/src/main/java/org/javarosa/xpath/expr/XPathPositionFunc.java
```diff
@@ -54,4 +54,12 @@ private static Double position(TreeReference refAt) {
         return new Double(refAt.getMultLast());
     }
 
+    @Override
+    public String getDocumentation() {
+        return getDocHeader()
+                + "Behavior: Returns the current index of the given reference. \n"
+                + "Return: the greatest extant multiplicity of the reference. \n"
+                + "Arguments: The reference to be checked, or none for current position "
+                + "Example: position(/data/repeat) returns '2' if two iterations have been created";
```
ctsims: @wpride We should remove documentation for the position($arg) form of this method. It is unsupported and has ambiguous confusing behavior.

phillipm:
@ctsims good point, I'll refactor it to load data from a json file that is only included JAR artifacts

PR #460, opened 2016-11-18
Fix session state... again

wpride:
This issue just reared its head again, this time in caught by a Formplayer end of form navigation test. I think what's happening is that the "return_to" session datum wasn't being added as a collectedDatum because its not a STATE_DATUM_VAL or STATE_UNKNOWN (its a STATE_DATUM_COMPUTED). I'm surprised this hasn't come up elsewhere, but I feel that way every time I "fix" a "bug" in this part of the code.

### DIFF
--- a/backend/src/org/commcare/session/CommCareSession.java
+++ b/backend/src/org/commcare/session/CommCareSession.java
@@ -425,7 +425,11 @@ public class CommCareSession {
     }
 
     public void setDatum(String keyId, String value) {
-        frame.pushStep(new StackFrameStep(SessionFrame.STATE_DATUM_VAL, keyId, value));
+        setDatum(SessionFrame.STATE_DATUM_VAL, keyId, value);
+    }
+
+    public void setDatum(String action, String keyId, String value) {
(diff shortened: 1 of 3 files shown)


phillipm:
Can you add a test that covers this behavior?

backend/src/org/commcare/session/CommCareSession.java
```diff
@@ -425,7 +425,11 @@ private void popStepInCurrentSessionFrame() {
     }
 
     public void setDatum(String keyId, String value) {
-        frame.pushStep(new StackFrameStep(SessionFrame.STATE_DATUM_VAL, keyId, value));
+        setDatum(SessionFrame.STATE_DATUM_VAL, keyId, value);
+    }
+
+    public void setDatum(String action, String keyId, String value) {
+        frame.pushStep(new StackFrameStep(action, keyId, value));
```
wpride: @phillipm adding a test required changing the loadSessionFromDescriptor method - previously every datum was set as a CASE_ID so these ambiguities between unknown, computed, and case_id were lost

PR #465, opened 2016-11-23
Cache fixture bases for massive performance wins

ctsims:
Implement a cache for fixtures at the same level as the casedb/ledger caches which ensures that we don't retrieve a fixture instance more than once.

The lifecycle for the fixture cache is aligned with the same lifecycle/process as the casedb/ledgerdb cache, so there should be no risk that it gets invalidated via sync differently than those two.

Huge, huge performance wins for apps with locations or other persistent fixtures, now that evaluation contexts are requested in many different situations during app execution

Release Note: Major performance improvements for large lookup tables

### DIFF
--- a/core/org/commcare/core/process/CommCareInstanceInitializer.java
+++ b/core/org/commcare/core/process/CommCareInstanceInitializer.java
@@ -15,10 +15,11 @@ import org.javarosa.core.model.instance.FormInstance;
 import org.javarosa.core.model.instance.InstanceInitializationFactory;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.services.locale.Localization;
+import org.javarosa.core.util.CacheTable;
 
 /**
- *  Initializes a CommCare DataInstance against a UserDataInterface and (sometimes) optional
- *  CommCareSession/Platform
+ * Initializes a CommCare DataInstance against a UserDataInterface and (sometimes) optional
+ * CommCareSession/Platform
  *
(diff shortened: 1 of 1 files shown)


core/org/commcare/core/process/CommCareInstanceInitializer.java
```diff
@@ -28,9 +29,11 @@
     protected final CommCareSession session;
     protected CaseInstanceTreeElement casebase;
     protected LedgerInstanceTreeElement stockbase;
+    protected CacheTable<String, TreeElement> fixtureBases = new CacheTable();
```
phillipm: To get ride of lint warnings: 
`private final CacheTable<String, TreeElement> fixtureBases = new CacheTable<>();`

core/org/commcare/core/process/CommCareInstanceInitializer.java
```diff
@@ -114,13 +117,21 @@ protected AbstractTreeElement setupFixtureData(ExternalDataInstance instance) {
...
 
-            TreeElement root = fixture.getRoot();
+                if (fixture == null) {
+                    throw new FixtureInitializationException(ref);
+                }
+
+                root = fixture.getRoot();
+
+                fixtureBases.register(key, root);
+            }
```
phillipm: Would be nice to lift all this logic into a helper like `loadFixtureRoot`

ctsims:
@phillipm Feedback addressed

PR #467, opened 2016-11-23
New XPath Function Handlers crash out with blank args

ctsims:
Adding custom functions seems to work fine, adding custom functions to override methods with no arguments seems to crash. Added test which illustrates and should pass when fixed. Didn't quite have time to dig in to what was up, so assigning over to @phillipm , since he wrote the substem.

Note: Prevents form entry from working in CLI


phillipm:
Sorry about that, I have a fix for that in [this commit](https://github.com/dimagi/commcare-core/pull/458/commits/54db5cda861ffed72ccefd5feb9c348d3da125c1) of [this PR](https://github.com/dimagi/commcare-core/pull/458) up for review. I'll add it here so we don't have to wait for that PR to get reviewed...

PR #468, opened 2016-11-24
CaseInstanceTreeElement & User refactors

phillipm:
Remove some unused code and fix lint warnings in `CaseInstanceTreeElement`

Remove unused code in `User`


cases/src/org/commcare/cases/instance/CaseInstanceTreeElement.java
```diff
@@ -57,12 +51,9 @@
      */
     final boolean reportMode;
 
-    public CaseInstanceTreeElement(AbstractTreeElement instanceRoot, IStorageUtilityIndexed storage, String[] caseIDs) {
-        this(instanceRoot, storage, false);
-        this.caseRecords = caseIDs;
-    }
```
phillipm: Appears to be unused in commcare, formplayer, and touchforms. Removed code associated with `caseRecords` because this is the only place it is set

cases/src/org/commcare/cases/instance/CaseInstanceTreeElement.java
```diff
@@ -161,38 +145,22 @@ protected synchronized void getCases() {
...
+            cases.addElement(new CaseChildElement(this, id, null, mult));
+            objectIdMapping.put(DataUtil.integer(id), DataUtil.integer(mult));
+            mult++;
         }
     }
 
-    public void setState(String syncToken, String stateHash) {
-        this.syncToken = syncToken;
-        this.stateHash = stateHash;
-    }
```
phillipm: Appears to be unused. Removed all code associated with `syncToken` and `stateHash` because these are the only places they are set.

src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
```diff
@@ -232,38 +196,24 @@ public String getAttributeNamespace(int index) {
...
     @Override
     public CaseChildElement getAttribute(String namespace, String name) {
-        //Oooooof, this is super janky;
         return null;
     }
 
     @Override
     public String getAttributeValue(String namespace, String name) {
-        return getAttributeValue("syncToken".equals(name) ? 0 : "stateHash".equals(name) ? 1 : -1);
+        return null;
```
amstone326: The default behavior here before was to return -1, so we probably want to preserve that

phillipm: Isn't it the same as `getAttributeValue(-1)` which is the same as `null`?

amstone326: haha oops my bad

PR #470, opened 2016-11-24
Remove IStorageUtilityIndexed.registerIndex

phillipm:
Remove `IStorageUtilityIndexed.registerIndex` and implementations because almost all of them are empty, the non-empty implementation is in the CLI/test code, and there is only one usage that isn't covered in any tests.

@ctsims can you confirm that these changes are okay? I'm fine closing this PR if you aren't sure; this is a sorta unclear part of code, so figured getting rid of it might be helpful for code comprehension.

cross-request: https://github.com/dimagi/commcare-android/pull/1582


ctsims:
@phillipm Can confirm that that method is only used on J2ME, so as long as this is a clean cut it should be totally fine.

PR #472, opened 2016-11-28
update error message for missing instance

snopoke:
@phillipm


snopoke:
@wspride @phillipm is `MessageFormat` a thing we can use?

wpride:
@snopoke should be totally fine now that we split the repos

snopoke:
@sheelio FYI

PR #474, opened 2016-11-29
add test asserting that false condition isn't evaluated

wpride:


javarosa/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -567,4 +569,49 @@ private static FormEntryController initFormEntry(FormParseInit fpi, String local
...
     }
+
+    /**
+     * Test that the untrue portion of an if() statement won't evaluate
+     */
+    @Test
+    public void testFormShortCircuit() {
+        FormParseInit fpi = new FormParseInit("/IfShortCircuitTest.xhtml");
+
+        // Custom func to double the numeric argument passed in.
```
amstone326: comment seems wrong

javarosa/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -567,4 +569,49 @@ private static FormEntryController initFormEntry(FormParseInit fpi, String local
...
+        IFunctionHandler functionFailer = new IFunctionHandler() {
+            @Override
+            public String getName() {
+                return "fail_function";
+            }
+
+            @Override
+            public Object eval(Object[] args, EvaluationContext ec) {
+                throw new RuntimeException("You failed the test!");
```
amstone326: would probably be helpful to be more specific about what test was failed

PR #476, opened 2016-11-30
Module badges

amstone326:
cross-request: https://github.com/dimagi/commcare-android/pull/1590


tests/resources/app_with_good_numeric_badge/suite.xml
```diff
@@ -0,0 +1,180 @@
...
+
+  <menu id="root" style="grid">
+    <text/>
+  </menu>
+
+  <menu id="m1" style="list">
+    <display>
+      <text>Menu 1 Text</text>
+      <text form="badge">
+        <xpath function="8"/>
```
ctsims: To Fix: this won't actually test whether the evaluation context for the badge is set correctly. I'd make sure this evaluating against session or case data or something similar.

amstone326: @ctsims Sorry about that I meant to fix that, updated now

PR #478, opened 2016-12-05
Restructure repo to follow gradle java standards

phillipm:
The `commcare-core` repo has lots of disparate source directories that go against normal [code structure conventions](https://docs.gradle.org/current/userguide/java_plugin.html). This PR moves all the (java) source code into `src/` and makes use of sourceset naming conventions.

```
src/
 - api
 - cli
 - main
 - test
 - translate
```

@wpride I checked that merging this with the `nimbus` branch results in only a few, probably managable merge conflicts.


build.gradle
```diff
@@ -16,9 +16,6 @@ sourceSets {
             srcDir 'javarosa/util/schema-gen/src'
             compileClasspath += main.output
         }
-        resources{
-            srcDir 'javarosa/src/test/resources'
-        }
```
phillipm: I'm fairly certain that the schema-gen code doesn't need to point to these test resources. Does that sound right?

PR #479, opened 2016-12-05
Better Error Message for calculates during parse

ctsims:


wpride:
any reason not to merge this with the catch statement above?

phillipm:
@wpride I like to isolate statements in `try` blocks because if we merged the two statements into one block, we wouldn't know which one throws which exception.

wpride:
@phillipm wouldn't the built message still be different?

phillipm:
What I was getting at is that, later, when you revisit and reread the following code, you won't know which line contributes to which exception, making comprehension and refactoring difficult.
```
try {
  r = buildCalculate(xpathCalc, ref);
  r = (Recalculate)_f.addTriggerable(r);
} catch (XPathSyntaxException | XPathException e) {
  throw buildParseException(nodeset, xpe.getMessage(), xpathCalc, "calculate");
}
```
My point isn't especially valid with this example because the code above is pretty concise/readable; I guess I was getting at that this is a pattern I usually like to follow..

ctsims:
Yeah, in this specific case I didn't want to conflate the two, since those errors could potentially come from either line, and might be handled differently in the future.  the notion that the second line can't throw the syntax, exception, might make it easier for someone to do something different with the thrown exception.

Having done some refactors lately, I agree with Phillip specifically for some runtimeexceptions that otherwise it's hard to tell what to expect might be thrown.

PR #480, opened 2016-12-06
Setup java performance benchmarking suite (JMH) 

phillipm:
`gradle jmh` runs the performance benchmark suite, which currently just loads and inits a large form (child classification) used by the TDH REC app.

I haven't been able to resolve Android Studio import issues, so when editing benchmark classes you won't get any IDE help.

NOTE: for whatever reason `gradle jmh` needs `compile 'org.openjdk.jmh:jmh-core:1.17.3'`, which causes android apk build issues, so I'm leaving it commented out. You'll have to uncomment those dependency lines whenever you want to use jmh...

Example of output:
```
Benchmark                        Mode  Cnt  Score   Error  Units
LoadAndInitLargeForm.form_init  thrpt   20  0.955 ± 0.019  ops/s
```


amstone326:
@phillipm Hey, just curious what we're envisioning the general long-term purpose of this infrastructure to be? I don't remember if there was a specific reason that you were setting it up.

build.gradle
```diff
@@ -79,6 +93,15 @@ dependencies {
       translateCompile files('lib/json-simple-1.1.1.jar')
       translateCompile 'joda-time:joda-time:2.9.3'
       translateCompile 'org.json:json:20140107'
+
+    // 'compile' dep also needed to run 'gradle jmh' for some reason,
```
amstone326: A little confused by this comment -- what is dep?

src/jmh/java/org/commcare/LoadAndInitLargeForm.java
```diff
@@ -0,0 +1,53 @@
...
+        session.setDatum("case_id_new_imci_visit_0",
+                "593ef28a-34ff-421d-a29c-6a0fd975df95");
+        sessionNavigator.startNextSessionStep();
+
+        FormEntryController fec = mockApp.loadAndInitForm("large_tdh_form.xml");
+        fec.getModel().getEvent();
+    }
+
+    // Execute using:
+    //  java -jar build\libs\commcare-core-jmh.jar -wi 5 -i 5 form_init
```
amstone326: can you add a short explanation of what the `-wi` and `-i` arguments do?

src/jmh/java/org/commcare/LoadAndInitLargeForm.java
```diff
@@ -0,0 +1,53 @@
...
+import org.commcare.modern.session.SessionWrapper;
+import org.commcare.test.utilities.MockApp;
+import org.commcare.test.utilities.MockSessionNavigationResponder;
+import org.javarosa.form.api.FormEntryController;
+
+public class LoadAndInitLargeForm {
+
+    @Benchmark()
+    @Group("form_init")
+    public void recForm() throws Exception {
```
amstone326: Does profiling for this method already automatically occur when the program is run because of the `@Benchmark()` tag? If so, where would it be reported to? If not, do you know what needs to be added to do so?

phillipm: yeah it runs automatically given the benchmark annotation and the output is printed to the console

build.gradle
phillipm: I meant it to stand for 'dependency' (I updated the comment to hopefully be more readable)

src/jmh/java/org/commcare/LoadAndInitLargeForm.java
phillipm: -wi is for warmup iteration count
-i is for iteration count

PR #483, opened 2016-12-07
update xpath error messages

snopoke:
@wpride 

cc @millerdev @nickpell @sheelio

### DIFF
--- a/src/main/java/org/javarosa/xpath/XPathTypeMismatchException.java
+++ b/src/main/java/org/javarosa/xpath/XPathTypeMismatchException.java
@@ -22,6 +22,6 @@ public class XPathTypeMismatchException extends XPathException {
     }
 
     public XPathTypeMismatchException(String s) {
-        super("type mismatch \n" + s);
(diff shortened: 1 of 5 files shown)


src/main/java/org/javarosa/xpath/XPathTypeMismatchException.java
```diff
@@ -22,6 +22,6 @@ public XPathTypeMismatchException() {
     }
 
     public XPathTypeMismatchException(String s) {
-        super("type mismatch \n" + s);
```
wpride: Was the thinking here that "type mismatch" was not useful information? Seems like we might want to keep some part of the message as indicative that this came from this exception class

snopoke: It's only every useful to developers who can also just look up the exception from the message. I also found that there wasn't a whole lot of consistency in terms of where this exception class was used, some of the places didn't seem like type mismatches.

PR #484, opened 2016-12-08
[GS Efficiency] update error message for cyclical references

snopoke:


src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2875,14 +2875,14 @@ private void checkDependencyCycles() {
         }
 
         if (!acyclic) {
-            String errorMessage = "XPath Dependency Cycle:\n";
+            String errorMessage = "Logic Dependency Cycle:\n";
             for (int i = 0; i < edges.size(); i++) {
                 TreeReference[] edge = (TreeReference[])edges.elementAt(i);
-                errorMessage += edge[0].toString() + " => " + edge[1].toString() + "\n";
+                errorMessage += edge[0].toString() + " references " + edge[1].toString() + "\n";
```
wpride: Not cleanly that I can see from the code. The conditionsTriggeredByRef() function for a reference would give you the Triggerables that depend on this ref which you could then check whether they're Conditions or Recalculates but there might be N Triggerables so impossible to tell.

snopoke:
:white_check_mark:

PR #485, opened 2016-12-08
TreeReference hash collision regression

wpride:
Will fail until we resolve the bug from http://manage.dimagi.com/default.asp?243592


wpride:
fixed in https://github.com/dimagi/commcare-core/pull/486, closing

PR #486, opened 2016-12-09
Fix for TreeElement Child Fetch Hash Collision Bug

ctsims:
Fix for  http://manage.dimagi.com/default.asp?243592

Tree Element children whose names have hash collisions could result in incorrect fetches.

Backported Tests from https://github.com/dimagi/commcare-core/pull/485 to include here. That PR can be closed.


phillipm:
Pretty sure this didn't make it into the `commcare_2.32.0` tag, so these changes are lost now that the `commcare_2.32` branch is closed

amstone326:
@phillipm Why do you think that? Both commits are right here: https://github.com/dimagi/commcare-core/commits/commcare_2.32.0

phillipm:
haha, you're right! Probably still rusty from vacation. I think I ran `git tag --contains 76b4b14b0c9e846b26e082c9bf3ec28838be3668` in the wrong repository locally...

PR #489, opened 2016-12-14
[NOTE - Needs HotFix] XPath Message formatting error

ctsims:
Fix and test for a bug introduced in https://github.com/dimagi/commcare-core/pull/483 

NOTE: this should get hotfixed back to 2.32, and we should test whether that PR has any similar issues (they may have been copy/pasted)

### DIFF
--- a/src/main/java/org/javarosa/xpath/XPathException.java
+++ b/src/main/java/org/javarosa/xpath/XPathException.java
@@ -32,11 +32,11 @@ public class XPathException extends RuntimeException {
     @Override
     public String getMessage() {
         if (prefix != null) {
-            return MessageFormat.format("{0}\n{1)", prefix, super.getMessage());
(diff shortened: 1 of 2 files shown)


snopoke:
Thanks and sorry. Re-checked the other PR and didn't see any other issues. It would be the one error I didn't test locally :(

ctsims:
@snopoke It happens to all of us, thanks for reviewing the other push!

amstone326:
FYI to all that no hotfix is needed, the original PR never made it's way into 2.32 in the first place

ctsims:
🎉🎉🎉 Woo! For some reason I thought I remembered the error msg thing getting pulled

czue:
nice! here I was thinking that @snopoke was going to sabotage the mobile team's hot streak... :smiling_imp:

PR #490, opened 2016-12-15
Fix for nested repeats not showing up (again)

amstone326:
Fix for http://manage.dimagi.com/default.asp?241001

Sadly, https://github.com/dimagi/commcare-core/pull/399 did not quite fix the issue of nested repeats not being shown when the outer repeat's count becomes greater than the inner repeat's count. `getLastRepeatInstanceIndex` was only getting the multiplicity of the direct next level to the current FormIndex, so in a form index with more than 2 levels, this didn't have the desired outcome, and was still reverting back to the old behavior of returning the multiplicity of the outer repeat. The solution is to make sure we have the deepest level being referenced by the current index before getting its multiplicity.

@ctsims, a few things:
1. This definitely fixes the issue reported above (from the broad app), but I was a little surprised because your comment on this method specifically included an example that I don't think would have worked ("1, 1_2, 2, 1_3"), since it has more than 2 levels. The format of the FormIndex in this app that was breaking was something like "1_8, 0, 4_1". So it was grabbing the 2nd level there, which doesn't have an instance index, and therefore defaulting back to the 8 from the first level. I just want to make sure that I'm not missing something more complicated here as to what went wrong in this specific case. 
2. I'm also not 100% how to go about writing a test for this new case, would you mind pointing me in the right direction?
3. I think we should probably do a point release for this, I don't think we can tell the broad team to wait for 2.33 to have this working properly. Does that sound right?

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormIndex.java
+++ b/src/main/java/org/javarosa/core/model/FormIndex.java
@@ -160,14 +160,11 @@ public class FormIndex {
      * @return The multiplicity of the "most specific" repeated instance of a question or group.
      */
     public int getLastRepeatInstanceIndex() {
-        if(this.getNextLevel() == null) {
-            return getInstanceIndex();
-        }
-        int childIndex = this.getNextLevel().getInstanceIndex();
-        if(childIndex == -1) {
+        int lastRepeatIndex = getDeepestLevel().getInstanceIndex();
+        if (lastRepeatIndex == -1) {
             return getInstanceIndex();
         } else {
-            return childIndex;
+            return lastRepeatIndex;
         }
     }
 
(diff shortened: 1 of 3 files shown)


amstone326:
@ctsims Updated `getLastRepeatInstanceIndex` to do what we discussed, updated the test xform to cover this new case, and took a stab at simplifying the datatype production definition. I'd like to write some unit tests for `FormIndex` but there are too many other interrupt tickets in the queue for me to dig into that right now, so it's on my backlog.

ctsims:
@amstone326 This looks good to me. This should go to a hotfix branch, though, right?

amstone326:
Yea I'll create one that includes this and https://github.com/dimagi/commcare-core/pull/491, but is there any reason not to merge this into master?

ctsims:
@amstone326 If you moved everything to the hotfix, the only reason not to merge would be that we should backmerge the hotfix to master anyway, so it would be redundant.

amstone326:
Closing in favor of https://github.com/dimagi/commcare-core/pull/492

PR #491, opened 2016-12-16
Actually respect not equals op in tryBatchChildFetch

amstone326:
Fix for http://manage.dimagi.com/default.asp?244045. This was certainly the hardest to track down bug I've had in a while, thanks @proteusvacuum 😈 

`tryBatchChildFetch` is a shortcut that we attempt to use to expand a nodeset reference with predicates. Any predicates that are an `XPathEqExpr` get applied in this method rather than somewhere further down the line that is presumably slower. The problem was that in this method we were just assuming that the `XPathEqExpr` in question was always an '=' expression, and not accounting for the possibility that it was '!='. 

Don't pull yet, need to add tests and probably re-target as a hotfix.


proteusvacuum:
yay! Thanks!

ctsims:
Here's the last one we fixed as well https://github.com/dimagi/commcare-core/pull/351/files

phillipm:
Haha, this is a great bug, nice work tracking it down! 👍 👍

amstone326:
Closing in favor of https://github.com/dimagi/commcare-core/pull/492

PR #494, opened 2016-12-20
Consolidate code in case & ledger instance tree elem implementations

phillipm:
Lots of duplicate code in `LedgerInstanceTreeElement` and `CaseInstanceTreeElement`. Moved that code into `StorageInstanceTreeElement`

Also moved duplicate code from `CaseChildElement` and `LedgerChildElement` to `StorageBackedChildElement`

cross-request: https://github.com/dimagi/commcare-android/pull/1601


src/main/java/org/commcare/cases/instance/CaseChildElement.java
```diff
@@ -321,7 +146,7 @@ private void setCaseProperties(Case c, TreeElement cacheBuilder) {
                 continue;
             }
 
-            TreeElement scratch = new TreeElement(parent.intern(key));
```
phillipm: I think I was probably in destructive mood and decided to rip out the unused string interner; going against a previous discussion we'd had about potentially trying to hook up the string interner again. I'll look into adding it back :)

phillipm: @amstone326 Good to go now

PR #497, opened 2016-12-23
new case relationship test cases

czue:
@proteusvacuum @ctsims is this right?

wait for / merge with https://github.com/dimagi/commcare-hq/pull/14327


czue:
i'm getting the impression these tests aren't actually running.

phillipm:
The extension case tests aren't running on jenkins!
Looks like there is some custom test runners that aren't getting picked up by `gradle test`.

czue:
i have all i need now, and these tests are wrong, closing. thanks for the help @phillipm!

PR #498, opened 2016-12-23
add more relationship test cases with extensions/children of extensions

czue:
https://github.com/dimagi/commcare-hq/pull/14329

@phillipm can you run these locally for me? I think they should be good and we can merge when https://github.com/dimagi/commcare-hq/pull/14329 is merged


phillipm:
tests passed

PR #502, opened 2017-01-03
Allow for storage-backed fixtures

phillipm:
Currently, when fixtures are stored by CommCare they are parsed into a large TreeElement and serialized as a big blog in the DB. This means that XPath queries over these fixtures must traverse the TreeElement, which is slow for large trees. An improvement was developed before my time to store TreeElements with uniform structures in DB tables and expose certain elements and attributes as table columns so that XPath queries can make use of speedy DB query tech. This was done for cases and ledgers. This PR applies this approach to semi-structured fixtures.

By semi-structured I mean the fixture must have a root and then every immediate child has the same name. For example:
```
<fixture id="some-fixture-name">
  <products>
    <product id="one">
      ...
    </product>
    <product id="two">
      ...
    </product>
    ...
  </products>
</fixture>
```

With that structure we can now create a new DB table for every semi-structured fixture and each row will contain a serialized TreeElement for the child (like a `product` from above) as well as a series of columns that expose elements and attributes to DB queries.

## Table Indices
Since we allow children to have any structure, we must select what elements and attributes we want to include in the table. This is done via a new XML structure that should precede the fixture in the user restore: 

```
<schema id="some-fixture-name">
  <indices>
    <index>@id</index>
    <index>id,@id</index>
    <index>nonexistent-element</index>
  </indices>
</schema>
```

Several restrictions are currently in place on what is a valid index in a `schema` element. We can currently only index elements/attributes that are direct children of the entry. For instance `.../products/product/name` is a valid index, but `.../products/product/name/@lang` isn't. Additionally, indexes can will be applied to the first entry found, so if you have a structure like:
```
<product>
  <code lang="en">pen</code>
  <code lang="pt">caneta</code>
</product>
```
The index `<index>code</index>` will blindly be applied to `<code lang="en>...`. This should be avoided, because I'm not sure how it would effect the resolution of queries...

Both of these constraints can possibly be remedied in the future with more thought and coding.

A notable flexibility is that indices can be created on non-homogenous elements/attributes and even non-existent elements/attributes.

## Example fixture and table expression
```
<fixture id="some-fixture-name" indexed="true">
  <products>
    <product id="one">
      <id>1</id>
      <value>capybara</value>
    </product>
    <product id="two">
      <id>2</id>
      <value>pudú</value>
    </product>
  </products>
</fixture>
```
This fixture will result in the following SQL DB statements
```
CREATE TABLE IF NOT EXISTS IND_FIX_some_fixture_name (_$id, _0id, _0nonexistent$$element);

CREATE INDEX _$id_index on IND_FIX_some_fixture_name (_$id);
CREATE INDEX _0id__$id_index on IND_FIX_some_fixture_name (_0id,_$id);
CREATE INDEX _0nonexistent$$element_index on IND_FIX_some_fixture_name (_0nonexistent$$element);
```
Note: since fixtures might have elements and attributes that conflict with SQL keywords, we escape attributes with a preceding `_$` and elements with a preceding `_0`. Dashes, which are invalid, are turned into `$$`.

## Fixture to base and child name DB
Since the base and child names for semi-structured fixtures are variable between fixtures we must store those names in a persistent store in order to instantiate the associated `IndexedFixtureInstanceTreeElement`. To do this we create a DB table in and insert/query it using `UserSandbox.setIndexedFixturePathBases` and `UserSandbox.getIndexedFixturePathBases`

cross-request: https://github.com/dimagi/commcare-android/pull/1610

Lengths time taken to process restore because each child TreeElement is serialized separately and inserted into the DB. I profiled this and the bulk of the time spent was on serialization, so I don't think there are easy ways to speed it up. I tried grouping all the DB inserts in an overarching DB transaction but that didn't lead to speed-ups.

## HQ Changes
Fixtures that we want to store in a storage indexed manner need to have the `indexed="true"` attribute: `<fixture id="some-fixture-name" indexed="true">`

Indices must be specified in the user restore using the `schema` xml element described above. These `schema` elements must appear before their associated fixture in the restore.


src/api/java/org/commcare/api/persistence/UserSqlSandbox.java
```diff
@@ -53,6 +62,32 @@ public UserSqlSandbox(String username) {
...
+                                           Set<String> indices) {
+        // TODO PLM: delete table if it already exists
+        // TODO PLM: create indexes over table
+        String tableName = StorageIndexedTreeElementModel.getTableName(fixtureName);
+        new SqliteIndexedStorageUtility<>(exampleEntry, username, tableName, path);
+    }
+
+    @Override
+    public Pair<String, String> getIndexedFixturePathBases(String fixtureName) {
+        throw new RuntimeException("implement in similar fashion as AndroidSandbox implementation");
```
amstone326: I'm guessing this isn't still supposed to be like this?

src/api/java/org/commcare/api/persistence/UserSqlSandbox.java
```diff
@@ -53,6 +62,32 @@ public UserSqlSandbox(String username) {
...
+    public IStorageUtilityIndexed<StorageIndexedTreeElementModel> getIndexedFixtureStorage(String fixtureName) {
+        String tableName = StorageIndexedTreeElementModel.getTableName(fixtureName);
+        return new SqliteIndexedStorageUtility<>(StorageIndexedTreeElementModel.class, username, tableName, path);
+    }
+
+    @Override
+    public void setupIndexedFixtureStorage(String fixtureName,
+                                           StorageIndexedTreeElementModel exampleEntry,
+                                           Set<String> indices) {
+        // TODO PLM: delete table if it already exists
```
amstone326: Should these TODO's happen now or are they just notes for later?

src/main/java/org/commcare/cases/instance/FixtureIndexSchema.java
```diff
@@ -0,0 +1,81 @@
...
+    private final Set<String> indices;
+    public final String fixtureName;
+
+    public FixtureIndexSchema(TreeElement schemaTree) {
+        this.fixtureName = schemaTree.getAttributeValue(null, "id");
+
+        indices = buildIndices(schemaTree.getChildrenWithName("index"));
+    }
+
+    private static Set<String> buildIndices(Vector<TreeElement> indexElements) {
```
amstone326: Super nitpicking, but this could just be an instance method called `setIndices` instead of a static method that returns something

src/api/java/org/commcare/api/persistence/UserSqlSandbox.java
phillipm: I was hoping to defer these implementations to @wpride because I'm not sure how this (formplayer specific?) DB setup works. I'll sync up with Will about this tomorrow.

src/api/java/org/commcare/api/persistence/UserSqlSandbox.java
phillipm: same as https://github.com/dimagi/commcare-core/pull/502#discussion_r95878366

src/main/java/org/commcare/cases/instance/FixtureIndexSchema.java
phillipm: 👍

src/main/java/org/commcare/cases/instance/IndexedFixtureChildElement.java
```diff
@@ -0,0 +1,74 @@
...
+                parent.getChildHintName());
+
+        StorageIndexedTreeElementModel modelTemplate = parent.getModelTemplate();
+        empty = modelTemplate.getRoot();
+        empty.setMult(TreeReference.INDEX_TEMPLATE);
+        // NOTE PLM: do we need to do more to convert a regular TreeElement into a template?
+    }
+
+    @Override
+    protected TreeElement cache() {
```
amstone326: Purely for my understanding -- what exactly is this method doing?

src/main/java/org/commcare/cases/instance/IndexedFixtureChildElement.java
```diff
@@ -0,0 +1,74 @@
...
+
+    protected IndexedFixtureChildElement(StorageInstanceTreeElement<StorageIndexedTreeElementModel, ?> parent,
+                                         int mult, int recordId) {
+        super(parent, mult, recordId, parent.getName(), parent.getChildHintName());
+    }
+
+    /**
+     * Template constructor (For elements that need to create reference nodesets but never look up values)
+     */
+    private IndexedFixtureChildElement(StorageInstanceTreeElement<StorageIndexedTreeElementModel, ?> parent) {
```
amstone326: Since this is only ever called by `buildFixtureChildTemplate` I would be in favor of not making it a constructor and just putting this code into that method.

phillipm: 👍

src/main/java/org/commcare/cases/instance/IndexedFixtureChildElement.java
phillipm: I think it caches the result of deriving the underlying TreeElement from the current object

src/main/java/org/commcare/cases/model/StorageIndexedTreeElementModel.java
```diff
@@ -0,0 +1,151 @@
...
+        ExtUtil.writeNumeric(out, recordId);
+        ExtUtil.writeString(out, ExtUtil.emptyIfNull(entityId));
+        ExtUtil.write(out, root);
+        ExtUtil.write(out, new ExtWrapList(indices));
+    }
+
+    /**
+     * Turns a column name into the corresponding attribute or element for the TreeElement
+     */
+    public static String getEntryFromCol(String col) {
```
amstone326: This method and the one below could probably have more descriptive names (which might remove the necessity of the method comments). Something like `getElementOrAttributeFromSqlColumnName()` and the reverse.

src/main/java/org/commcare/cases/model/StorageIndexedTreeElementModel.java
```diff
@@ -0,0 +1,151 @@
...
+            // are only made over entries with multiplicity 0
+            TreeElement child = root.getChild(getEntryFromCol(fieldName), 0);
+            IAnswerData value = child.getValue();
+            if (value == null) {
+                return "";
+            } else {
+                return value.uncast().getString();
+            }
+        }
+        return null;
```
amstone326: It looks like most other implementations of `getMetaData()` throw an `IllegalArgumentException` when they are called with a `fieldName` that doesn't have an indexed column.

src/main/java/org/commcare/core/parse/CommCareTransactionParserFactory.java
```diff
@@ -159,6 +174,12 @@ public TransactionParser getParser(KXmlParser parser) {
         };
     }
 
+    public IndexedFixtureXmlParser buildIndexedFixtureParser(KXmlParser parser,
+                                                             String fixtureName,
+                                                             FixtureIndexSchema schema) {
+        return new IndexedFixtureXmlParser(parser, fixtureName, schema, sandbox);
```
amstone326: since this is just a one-liner can we call it directly?

src/main/java/org/commcare/core/parse/CommCareTransactionParserFactory.java
```diff
@@ -80,20 +85,31 @@ public TransactionParser getParser(KXmlParser parser) {
...
+        } else if (FixtureIndexSchemaParser.INDICE_SCHEMA.equalsIgnoreCase(name)) {
+            return new FixtureIndexSchemaParser(parser, fixtureSchemas);
         } else if ("fixture".equalsIgnoreCase(name)) {
+            String id = parser.getAttributeValue(null, "id");
+            String isIndexedAttr = parser.getAttributeValue(null, "indexed");
+            boolean isIndexed = "true".equals(isIndexedAttr);
             req();
-            return fixtureParser.getParser(parser);
+            if (isIndexed) {
+                FixtureIndexSchema schema = fixtureSchemas.get(id);
```
phillipm: I purposefully designed it this way because the setup will still work it just won't have any meta data columns. I don't know if there is any use to this, but it would mean that instead of storing one huge fixture blob in one row of the normal fixture db it would create a dedicated table and store a bunch of little blobs for each entry in the fixture.

amstone326: Hm, I'm not following what you mean by that I don't think. We can follow up offline

src/main/java/org/commcare/cases/instance/IndexedFixtureInstanceTreeElement.java
```diff
@@ -0,0 +1,70 @@
...
+import org.javarosa.core.model.instance.AbstractTreeElement;
+import org.javarosa.core.model.instance.InstanceBase;
+import org.javarosa.core.services.storage.IStorageUtilityIndexed;
+import org.javarosa.model.xform.XPathReference;
+import org.javarosa.xpath.expr.XPathPathExpr;
+
+import java.util.Hashtable;
+
+/**
+ * The root element for the a indexed fixture data instance: instance('some-indexed-fixture').
```
ctsims: for clarify in the documentation, this element represents 

instance('some-indexed-fixture')/fixtureroot

doesn't it?

src/main/java/org/commcare/cases/model/StorageIndexedTreeElementModel.java
```diff
@@ -0,0 +1,151 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.Set;
+import java.util.Vector;
+
+/**
+ * A DB model for storing TreeElements such that particular attributes and
+ * elements are indexed and queryable using the DB.
+ *
+ * Indexed attributes/elements get their own table columns, and the rest of
```
ctsims: Woah. I legitimately didn't realize this would literally fit into the existing model set. That's actually pretty cool

src/main/java/org/commcare/cases/model/StorageIndexedTreeElementModel.java
```diff
@@ -0,0 +1,151 @@
...
+        String[] escapedIndexList = new String[indices.size()];
+        int i = 0;
+        for (String index : indices) {
+            escapedIndexList[i++] = getColFromEntry(index);
+        }
+        return escapedIndexList;
+    }
+
+    public static String getTableName(String fixtureName) {
+        String cleanedName = fixtureName.replace(":", "_").replace(".", "_").replace("-", "_");
```
ctsims: I don't know if it's possible, but it's generally better to apply this transform once, then store the result (the place to do this may come up if we start storing the full fixture), rather than do this dynamically, as if we need to change the logic for it later, we can just keep using the retained name, rather than keeping both around and being confused about possibly using multiple ways to refer to the fixture table and checking each one.

src/main/java/org/commcare/cases/model/StorageIndexedTreeElementModel.java
```diff
@@ -0,0 +1,151 @@
...
+
+    @Override
+    public String[] getMetaDataFields() {
+        return metaDataFields;
+    }
+
+    @Override
+    public Object getMetaData(String fieldName) {
+        if (fieldName.startsWith(ATTR_PREFIX)) {
+            return root.getAttributeValue(null, getEntryFromCol(fieldName).substring(1));
```
ctsims: .substring(1) seems like a very dangerous way to eliminate the prefix. at the very least I'd do substring(ATTR_PREFIX.length())

src/main/java/org/commcare/cases/model/StorageIndexedTreeElementModel.java
ctsims: +1

src/main/java/org/commcare/xml/IndexedFixtureXmlParser.java
```diff
@@ -0,0 +1,120 @@
...
+    private final String fixtureName;
+    private IStorageUtilityIndexed<StorageIndexedTreeElementModel> indexedFixtureStorage;
+
+    public IndexedFixtureXmlParser(KXmlParser parser, String fixtureName,
+                                   FixtureIndexSchema schema, UserSandbox sandbox) {
+        super(parser);
+        this.sandbox = sandbox;
+        this.fixtureName = fixtureName;
+
+        if (schema == null) {
```
ctsims: Flagging that it's going to be hard for users to identify when they've made the mistake of sending down the index data _after_ the fixture data, since it will work, it'll just be super slow.

We should possibly make "including a schema element _after_ a fixture element with the same name" an error, even if having a fixture element with no schema is not.

src/test/java/org/commcare/fixtures/test/StorageBackedFixtureTest.java
```diff
@@ -0,0 +1,37 @@
...
+    public void setUp() {
+        sandbox = MockDataUtils.getStaticStorage();
+    }
+
+    @Test
+    public void queryIndexedLookup() throws XPathSyntaxException, UnfullfilledRequirementsException,
+            XmlPullParserException, IOException, InvalidStructureException {
+        ParseUtils.parseIntoSandbox(this.getClass().getResourceAsStream("/indexed-fixture.xml"), sandbox);
+
+        EvaluationContext ec =
```
ctsims: It might be worth some manual test in here (somehow!) to ensure that this isn't using the old fixture format.

src/test/resources/mixed_instance_initializers/user_restore.xml
```diff
@@ -1,21 +1,44 @@
...
+            <item id="2">two</item>
+            <item id="3">three</item>
+            <item id="4">four</item>
+        </list>
+    </fixture>
+    <table-indices id="bike-list">
+        <index>@id</index>
+        <index>id</index>
+    </table-indices>
+    <fixture indexed="true" id="bike-list" user_id="test_uuid">
```
ctsims: Would be worth adding tests for the non-homogenous schemas.

src/test/resources/mixed_instance_initializers/user_restore.xml
```diff
@@ -1,21 +1,44 @@
...
+        <list>
+            <item id="1">one</item>
+            <item id="2">two</item>
+            <item id="3">three</item>
+            <item id="4">four</item>
+        </list>
+    </fixture>
+    <table-indices id="bike-list">
+        <index>@id</index>
+        <index>id</index>
```
ctsims: Would also be worth testing the compound schemas, right?

src/main/java/org/commcare/cases/instance/IndexedFixtureInstanceTreeElement.java
phillipm: yeah, good call!

src/main/java/org/commcare/xml/IndexedFixtureXmlParser.java
phillipm: 👍

src/main/java/org/commcare/cases/model/StorageIndexedTreeElementModel.java
phillipm: That sounds like a good idea but I didn't get to this. It could be stored in the fixture base/child paths table.

src/api/java/org/commcare/api/persistence/UserSqlSandbox.java
```diff
@@ -53,6 +66,65 @@ public UserSqlSandbox(String username) {
...
+        String tableName = StorageIndexedTreeElementModel.getTableName(fixtureName);
+        SqliteIndexedStorageUtility<StorageIndexedTreeElementModel> sqlUtil =
+                createFixturePathsTable(tableName);
+
+        Map<String, String> contentVals = new HashMap<>();
+        contentVals.put(IndexedFixturePathsConstants.INDEXED_FIXTURE_PATHS_COL_BASE, baseName);
+        contentVals.put(IndexedFixturePathsConstants.INDEXED_FIXTURE_PATHS_COL_CHILD, childName);
+        contentVals.put(IndexedFixturePathsConstants.INDEXED_FIXTURE_PATHS_COL_NAME, fixtureName);
+
+        sqlUtil.basicInsert(contentVals);
```
phillipm: @wpride implemented this but didn't test it at all. Can you take over on this?

src/api/java/org/commcare/api/persistence/UserSqlSandbox.java
phillipm: @wpride Wasn't able to get this implemented.

phillipm:
@ctsims / @amstone326 updated the pr to reflect your comments. The formplayer DB code is half-finished so that will need to be implemented before formplayer QA starts. In the meantime this PR can be merged and app QA can start.

I didn't check performance implications of when filtering a locations fixture based on non-indexed element.

src/test/java/org/commcare/fixtures/test/StorageBackedFixtureTest.java
phillipm: Wasn't able to think of a way to do this. I did add a check that the fixture gets added to both the new db tables as long as the traditional db table.

ctsims:
@wspride  Flagging that there's some backend work still for web apps before this will work in touchforms, we' are pulling for cutting release on mobile for nikshay testing, but Phillip mentioned we should still review the touchforms parts specifically

czue:
hey - confirming that the `table-indices` node will have fine/non-breaking behavior before 2.33? if that's not the case we will need to bump the OTA restore version attribute so we can selectively include it. did we ever start passing the app version up in headers? that could also work.

czue:
thanks for this last hoorah @phillipm ! you don't know how huge a contribution this is

PR #506, opened 2017-01-12
Filter rules for combo box

amstone326:
Supporting changes for https://github.com/dimagi/commcare-android/pull/1613

### DIFF
--- a/src/main/java/org/javarosa/core/model/ComboboxFilterRule.java
+++ b/src/main/java/org/javarosa/core/model/ComboboxFilterRule.java
@@ -1,0 +1,23 @@
+package org.javarosa.core.model;
+
+/**
+ * @author Aliza Stone
+ */
+public interface ComboboxFilterRule {
+
+    /**
+     *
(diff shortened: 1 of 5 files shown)


src/main/java/org/javarosa/core/model/ComboboxFilterRule.java
```diff
@@ -0,0 +1,23 @@
...
+public interface ComboboxFilterRule {
+
+    /**
+     *
+     * @param choice - an answer choice available in this adapter
+     * @param textEntered - the text entered by the user in the combobox's edittext field
+     * @return If the given choice should be displayed in combobox's dropdown menu, based upon
+     * the text that the user currently has entered
+     */
+    boolean choiceShouldBeShown(String choice, CharSequence textEntered);
```
ctsims: CharSequence is an Android Specific class, is it not?

amstone326: @ctsims Nope it's actually not! It's part of `java.lang`

PR #507, opened 2017-01-18
First pass example - Optimize "selected()" 

ctsims:


ctsims:
Moving to https://github.com/dimagi/commcare-core/pull/508

PR #508, opened 2017-01-19
First pass example - Optimize "selected()

ctsims:


src/main/java/org/commcare/cases/util/PredicateProfile.java
```diff
@@ -0,0 +1,27 @@
...
+ * Created by ctsims on 1/18/2017.
+ */
+
+public interface PredicateProfile {
+
+    /**
+     * @return The index the predicate is operating over. This will be moved soon and isn't
+     * a fundamental part of the structure.
+     */
+    public String getKey();
```
wpride: public is unnecessary

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
```diff
@@ -101,40 +100,68 @@ private void collectProcessablePredicates(Vector<XPathExpression> predicates,
...
+                //go check whether this key is loaded into cached memory.
+                keyMapping = getKeyMapping(key);
+            }
+
+            //TODO: Move key mapping and "Index Match" to be their own optimization implementations
             if (keyMapping != null) {
                 //If so, go fetch that element's record id and skip the storage
                 //lookup
-                Integer uniqueValue = keyMapping.get(FunctionUtils.toString(o));
+                Integer uniqueValue = keyMapping.get(FunctionUtils.toString(((IndexedValueLookup)profiles.elementAt(0)).value));
```
wpride: If you move the valueLook  declaration outside of the if you can reuse that var here

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
```diff
@@ -101,40 +100,68 @@ private void collectProcessablePredicates(Vector<XPathExpression> predicates,
...
+                //Get the first set of values.
+                String key = valueLookup.key;
+
+                //Some storage roots will collect common iterative mappings ahead of time,
+                //go check whether this key is loaded into cached memory.
+                keyMapping = getKeyMapping(key);
+            }
+
+            //TODO: Move key mapping and "Index Match" to be their own optimization implementations
             if (keyMapping != null) {
```
wpride: Maybe outside this PR scope but this whole if/else might belong in the if at line 148?

ctsims: The logic is kind of awkward because you need to be able to fall over to the second body _either_ because the keyMapping doesn't exist for this key, or because it's the wrong kind of optimization (IE: If the keyMapping does exist, we want to never execute the second body, but it could not exist for one of two reasons). 

Definitely needs refactoring, but I think I'm going to refactor the optimization set entirely to make this make more sense next (outside of the 2.33 release)

ctsims:
@wpride made one change and answered the question from the review.

PR #515, opened 2017-02-02
Include question ref in error msg

amstone326:
Addresses insufficient error msg reported here: http://manage.dimagi.com/default.asp?246752


wpride:
👍

PR #516, opened 2017-02-03
Install app from list of available apps for a user

amstone326:


src/main/java/org/commcare/xml/AvailableAppsParser.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final String APP_TAG = "app";
+    private static final String DOMAIN_TAG = "domain";
+    private static final String APP_NAME_TAG = "name";
+    private static final String APP_VERSION_TAG = "version";
+    private static final String PROFILE_REF_TAG = "profile";
+    private static final String MEDIA_PROFILE_REF_TAG = "media-profile";
+    private static final String ENVIRONMENT_TAG = "environment";
+    private static final String PROD_VALUE = "Production";
+
+    public AvailableAppsParser(KXmlParser parser) {
```
wpride: Nice!

src/main/java/org/commcare/suite/model/AppAvailableForInstall.java
```diff
@@ -0,0 +1,73 @@
...
+    public AppAvailableForInstall() {
+        // for deserialization
+    }
+
+    public AppAvailableForInstall(String domain, String appName, String appVersion,
+                                  boolean isOnProd, String profileRef, String mediaProfileRef) {
+        this.domain = domain;
+        this.appName = appName;
+        this.appVersion = appVersion;
+        this.isOnProd = isOnProd;
```
wpride: doesn't seem like `isOnProd` is ever used?

wpride: in fact seems like only a few of these values are ever used?

src/main/java/org/commcare/xml/AvailableAppsParser.java
```diff
@@ -0,0 +1,60 @@
...
+        parser.next();
+        int eventType = parser.getEventType();
+        do {
+            if (eventType == KXmlParser.START_TAG) {
+                String tagName = parser.getName().toLowerCase();
+                if (APP_TAG.equals(tagName)) {
+                    String domain = parser.getAttributeValue(null, DOMAIN_TAG);
+                    String appName = parser.getAttributeValue(null, APP_NAME_TAG);
+                    String appVersion = parser.getAttributeValue(null, APP_VERSION_TAG);
+                    boolean isOnProd = PROD_VALUE.equals(parser.getAttributeValue(null, ENVIRONMENT_TAG));
```
wpride: confused about the value of this boolean

src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -72,26 +82,35 @@ private void setupAuthentication(boolean isAuth, User user, String domain) {
...
+            setupAuthentication(username, password, user);
         } else {
             // clear any prior set authenticator to make unauthed requests
             Authenticator.setDefault(null);
         }
     }
 
+    private void setupAuthentication(final String username, final String password, User user) {
+        if (username == null || password == null ||
+                (user != null && User.TYPE_DEMO.equals(user.getUserType()))) {
```
wpride: Would password not be null if its a demo user?

wpride: I'm a bit confused what the various paths through the two setupAuthentication methods are - are they:

1. Web user, not logged in
2. Mobile user, not logged in
3. Mobile user, logged in
4. Demo user, logged in

?

Either way feels strange to check whether its a demo user this far along - might be better to not expose this feature to demo users from the outset?

src/main/java/org/commcare/suite/model/AppAvailableForInstall.java
amstone326: Yea.. wasn't sure what to do about this; I have all of these here because Farid is sending down all of this info in the response, so I felt like I should store it in case it became useful later. But I guess YAGNI, and I should just ignore the fields that I'm not using.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
amstone326: So if you look at the full changes to this class, that check for demo user was already there. The `ModernHttpRequester` class is used for a lot more than just this feature, it's just a generic HTTP requester. All that I've done here is to add a way to use this class to make a request from a user who hasn't already logged into CommCare, by creating this overloaded method that accepts just `username` and `password` String arguments, instead of a `User` object (which is only accessible once a user has logged in).

At the level of this class, there's also no distinction between a mobile user and a web user, we're just thinking in terms of authentication for an HTTP request, so we ultimately just need a username and password. If you read through the old implementation of `setupAuthentication()` that method accepted a `User` object, and all that that object was being used for was to grab the username and password from it, and then it used that to set up the `Authenticator`. So I've moved all of that work to my overloaded method, and changed the original version to simply call my new version with the username and password taken from the `User` object.

If this is still confusing, we can talk through in person, which might be easier.

amstone326:
@wpride Responded/made changes.

PR #517, opened 2017-02-08
Query Planning

ctsims:
Very simply query planner which give us a place to migrate optimizations we've been using for predicate evaluation and make informed decisions about which to use/not use when evaluating complex queries.

Also provides a Query Profiler which makes it possible to understand how a query was executed, and profile the runtime of that evaluation.

cross-request: https://github.com/dimagi/commcare-android/pull/1632


src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
```diff
@@ -100,7 +113,7 @@ private void collectPredicateProfiles(Vector<XPathExpression> predicates,
                             //sure the best way to do that....? Maybe tell the evaluation context to skip out here if it detects a request
                             //to resolve in a certain area?
                             Object o = FunctionUtils.unpack(((XPathEqExpr)xpe).b.eval(evalContext));
-                            optimizations.addElement(new IndexedValueLookup(filterIndex, o));
+                            optimizations.addElement(new org.commcare.cases.query.IndexedValueLookup(filterIndex, o));
```
amstone326: did you add this intentionally?

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
```diff
@@ -83,7 +85,18 @@ protected String translateFilterExpr(XPathPathExpr expressionTemplate, XPathPath
...
                                           Hashtable<XPathPathExpr, String> indices,
                                           EvaluationContext evalContext,
-                                          Vector<PredicateProfile> optimizations) {
+                                          Vector<org.commcare.cases.query.PredicateProfile> optimizations,
+                                          QueryContext queryContext) {
+
+        optimizations.addAll(getQueryPlanner().collectPredicateProfiles(predicates, queryContext, evalContext));
+
+        //For now we are going to skip looking deeper if we trigger
+        //any of the planned optimizations
```
amstone326: Why? Are they duplicates in some way? Or just less powerful?

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -297,15 +322,31 @@ private void expandReferenceAccumulator(TreeReference sourceRef, DataInstance so
...
         }
 
+        QueryContext subContext = queryContext.
+                checkForDerivativeContextAndReturn(childSet == null ? 0 : childSet.size());
+
+        //If we forked a new query body from above (IE: a new large query) and there wasn't an
+        //original context before, we can anticipate that the subcontext below will refernce
+        //into the returned body as the original context, which is ugly, but opens up
+        //intense optimizations
+        if(this.getOriginalContextForPropogation() == null && subContext != queryContext) {
```
amstone326: Did not successfully follow this :(

src/main/java/org/commcare/cases/util/QueryUtils.java
```diff
@@ -0,0 +1,36 @@
...
+        return null;
+    }
+
+    public static Vector<Integer> wrapSingleResult(Integer result) {
+        Vector<Integer> results = new Vector<>();
+        results.add(result);
+        return results;
+    }
+
+    public static void poke(Object o, QueryContext context) {
```
amstone326: Is this wrapper method necessary? Seems like `notifySensitive` could contain all of this logic. `poke()` is also a very ambiguous method name.

src/main/java/org/javarosa/core/model/trace/BulkEvaluationTrace.java
```diff
@@ -0,0 +1,82 @@
...
+import org.javarosa.xpath.expr.XPathExpression;
+
+import java.util.Collection;
+import java.util.Vector;
+
+/**
+ * Created by ctsims on 1/24/2017.
+ */
+
+public class BulkEvaluationTrace extends EvaluationTrace {
```
amstone326: class-level comment needed

src/main/java/org/javarosa/core/model/trace/BulkEvaluationTrace.java
```diff
@@ -0,0 +1,82 @@
...
+
+    boolean bulkEvaluationSucceeded = false;
+    long runtimeImpact = 0;
```
amstone326: unused

src/main/java/org/javarosa/core/model/trace/EvaluationTraceReduction.java
```diff
@@ -0,0 +1,115 @@
...
+
+    Vector<EvaluationTraceReduction> children = new Vector<>();
+
+    public EvaluationTraceReduction(EvaluationTrace trace) {
+        super(trace.getExpression());
+        this.expression = trace.getExpression();
+        foldIn(trace);
+    }
+
+    public void foldIn(EvaluationTrace trace) {
```
amstone326: comment needed

src/main/java/org/commcare/cases/query/queryset/CaseQuerySetLookup.java
```diff
@@ -0,0 +1,125 @@
...
+            return false;
+        } else {
+            ModelQuerySet set =
+                    context.getQueryCache(QuerySetCache.class).getModelQuerySet(CurrentModelQuerySet.CURRENT_QUERY_SET_ID);
+            if (set == null) {
+                //this optimization isn't useful because the current context doesn't contain
+                //a query set
+                return false;
+            }
+            return true;
```
wpride: None of the tests in CaseDbQueryTests handle the condition where this Lookup is actually valid (and neither of the other two tests cover this code path)

src/main/java/org/commcare/cases/query/handlers/ModelQueryLookupHandler.java
```diff
@@ -0,0 +1,86 @@
...
+            return null;
+        }
+
+        TreeReference ref = lookup.getLookupIdKey(evalContext);
+
+        if(!lookup.isValid(ref, context)) {
+            return null;
+        }
+
+        Vector<PredicateProfile> newProfile = new Vector<>();
```
wpride: The tests never reach this part of  the code path

src/main/java/org/commcare/cases/query/queryset/DerivedCaseQueryLookup.java
```diff
@@ -0,0 +1,77 @@
+package org.commcare.cases.query.queryset;
```
wpride: This class and its inheritor also seem an unnecessary split that makes the code hard to read. Unless you have a very explicit reason for keeping its parent abstract class the logic seems so paired that they could be merged. I'm not clear what logic belongs in here and what in its inheritors like CaseIndexQueryTransform. It also doesn't look like any of the QuerySetLookup methods are called in the tests so I can't tell what's going on.

wpride:
@ctsims I spent a good four hours on this this morning and still don't think I fully grasp the entire changeset. Partially I think that's because its just a large PR but I also think a lot of the abstractions are really hard to understand. If I were to implement a matcher or transform I'd really not be sure where to start. As it is even knowing, for example, that one of the tests is supposed to handle transforming "index/parent" into a more consumable DB lookup I'm still not sure how to trace that process. I think this PR needs some combination of 

1. Refactor of the abstractions 
2. Better class level commenting
3. Better test coverage with more explicitly description of what we're testing

Also was pushing some random linting/comments here as I went along: https://github.com/dimagi/commcare-core/pull/519

I'll take another look with fresh eyes on Monday but as it is I'd have a very hard time debugging any issues along this path or implementing this in Formplayer

src/main/java/org/commcare/cases/instance/StorageInstanceTreeElement.java
```diff
@@ -268,8 +269,16 @@ public String intern(String s) {
...
                                       int recordId, String id, int mult);
 
-    protected Model getElement(int recordId) {
-        return storage.read(recordId);
+    protected Model getElement(int recordId, QueryContext context) {
+        EvaluationTrace trace = new EvaluationTrace("Model Load[" + childName+"]");
+
+        Model m = storage.read(recordId);
+
+        trace.setOutcome(String.valueOf(recordId));
```
ctsims: Exactly. These traces are used to denote that a model was loaded directly from the DB for performance profiling (IE: how much time was spent pulling cases from storage one at a time).

src/main/java/org/commcare/cases/query/QueryCache.java
```diff
@@ -0,0 +1,47 @@
+package org.commcare.cases.query;
+
+import java.util.HashMap;
+
+/**
+ * Created by ctsims on 1/26/2017.
+ */
+
+public class QueryCache {
```
ctsims: Updated the docs to reflect this a bit better(push is coming with the updated docs)

src/main/java/org/commcare/cases/query/QueryCacheEntry.java
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * A query cache stores data local to a query context which will be relevant for recomputation
+ * within the lifecycle of the Query.
+ *
+ * Note that there can only be one instance of a query cache at any given time
+ *
+ * Created by ctsims on 1/26/2017.
+ */
+
+public interface QueryCacheEntry {
```
ctsims: Updated the docs. This is a tag interface used to specify the lifecycle of the object

src/main/java/org/commcare/cases/query/QueryContext.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private void reportContextEscalation(QueryContext queryContext, QueryContext newContext, String label) {
+        EvaluationTrace trace = new EvaluationTrace(label + " Query Context [" + newContext.contextScope +"]");
+        trace.setOutcome("");
+        reportTrace(trace);
+    }
+
+    public void reportTrace(EvaluationTrace trace) {
+        traceRoot.reportSubtrace(trace);
```
ctsims: yeah, I'm super unhappy about how the trace root gets revealed/passed along but it's quite complex with all the nesting, so I don't have a much better plan. I updated this in the new push to at least protect against hte null case. No reason we should ever crash there.

src/main/java/org/commcare/cases/query/QuerySensitive.java
```diff
@@ -0,0 +1,9 @@
+package org.commcare.cases.query;
+
+/**
+ * Created by ctsims on 1/27/2017.
+ */
+
+public interface QuerySensitive {
```
ctsims: That's a fair concern. My worry was that Will might need to implement the "poke" concept on some Web apps specific code, which would require making core changes to the libraries in order to proceed. If the ability to sign up for a poke comes from your own object it lets him (or android) accept that signal without needing a new build of the core libraries.

ctsims: for reference we will likely implement indexed storage refactors soon that will use this interface as well

src/main/java/org/commcare/cases/query/handlers/ModelQueryLookupHandler.java
```diff
@@ -0,0 +1,86 @@
...
+            trace.setOutcome("Results: " +lookupData.size());
+            queryContext.reportTrace(trace);
+        }
+
+        return lookupData;
+    }
+
+    @Override
+    public void updateProfiles(ModelQueryLookup querySet, Vector<PredicateProfile> profiles) {
+        profiles.removeElementAt(0);
```
ctsims: the only reason is that there are optimizations which are written in the old framework which need to be moved into the new framework and handle multiple queries.

The optimization which turns 

[@case_type = 'foo'][@case_status = 'open'] 

for instance will need to remove multiple predicates in one go, so I didn't want to shoot ourselves in the foot by needing to rewrite it.

I updated this to "profiles.remove(querySet) though, for consistency

src/main/java/org/commcare/cases/query/handlers/ModelQueryLookupHandler.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public Collection<PredicateProfile> collectPredicateProfiles(Vector<XPathExpression> predicates,
```
ctsims: Commented above

src/main/java/org/commcare/cases/query/queryset/CaseModelQuerySetMatcher.java
```diff
@@ -0,0 +1,127 @@
...
+                XPathReference.getPathExpr("instance('" + modelId + "')/casedb/case").getReference();
+
+        //Later on we need this to refer to a real element at casedb, not a virtual one
+        caseDbRoot.setMultiplicity(0, 0);
+        this.multiplicityMap = multiplicityMap;
+
+        membershipIndexes = new Vector<>();
+        membershipIndexes.add(CaseInstanceTreeElement.CASE_ID_EXPR);
+        membershipIndexes.add(CaseInstanceTreeElement.CASE_ID_EXPR_TWO);
+        querySetTransforms.add(new CaseIdentityQuerySetTransform());
```
ctsims: done

src/main/java/org/commcare/cases/query/queryset/DerivedCaseQueryLookup.java
ctsims: The tricky part about CaseIndexQueryTransform is that we can't put it in core, because core doesn't support the index table and the relevant operations needed to perform the transform. In core the only model query lookups that work are referential (IE:  anything that matches a case_id to a case_id).

I moved this here because it would let the bulk of the logic live in core for being re-implemented on other platforms. 

This code is triggered by  a test in the android codebase here:
https://github.com/dimagi/commcare-android/pull/1632/files#diff-c678f23327e97d00bb82c3717c08bbe0R105

ctsims: Updating docs for this method as well

src/main/java/org/commcare/cases/util/QueryUtils.java
ctsims: you're totally right, I had this in here originally for peformance profiling reasons. removing

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
ctsims: The old format for genreating predicateprofiles relied on being able to match things like @case_id on the left and some arbitrary value on the right that was generated by running the XPathExpression against the current evalContext. 

In cases where calculating the right hand side expression is the expensive part of the query this makes it ~impossible to optimize without preventing things from being generated.

In the future the best plan will be to delay the evaluation of the RightHandSide part of the predicate profile until it's ready for use in a specific handler, but I didn't have time to write/test that level of refactoring the process

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -297,15 +322,31 @@ private void expandReferenceAccumulator(TreeReference sourceRef, DataInstance so
...
 
+        this.openBulkTrace();
+
         // Use the reference's simple predicates to filter the potential
         // nodeset.  Predicates used in filtering are removed from the
         // predicate input argument.
-        Vector<TreeReference> childSet = node.tryBatchChildFetch(name, mult, predicates, this);
+        Collection<TreeReference> childSet = node.tryBatchChildFetch(name, mult, predicates, this);
+
+        this.reportBulkTraceResults(originalPredicates, predicates, childSet);
```
ctsims: Predicates is mutable and gets incrementally shaved down as predicates are "batch fetched". OriginalPredicates contains the predicates that were requested coming into the method.

src/main/java/org/javarosa/core/model/trace/ReducingTraceReporter.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.HashMap;
+import java.util.Vector;
+
+/**
+ * A Cummulative trace reporter collects and "folds" traces which execute over multiple elements.
+ * It is helpful for identifying how many times different expressions are evaluated, and aggregating
+ * elements of each execution
+ *
+ * Created by ctsims on 10/19/2016.
```
ctsims: that's... super weird. I probably copy/pasted from somewhere. Will fix

src/main/java/org/commcare/cases/query/QueryContext.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * @param newScope the magnitude of the new query
+     * @return either the existing QueryContext or a new (child) QueryContext
+     * if the magnitude of the new query exceeds the parent sufficiently
+     */
+    public QueryContext checkForDerivativeContextAndReturn(int newScope) {
+        QueryContext newContext;
+
+        //TODO: I think we may need to clear the spanwed context's spawned context (maybe?) if it
```
amstone326: I'm not sure if comments like this are helpful, we have a bunch of them all over the codebase that never get addressed because no one knows what to do with them. I would be in favor of either deciding this is important enough to address now, or removing this.

src/main/java/org/commcare/cases/query/QueryHandler.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Vector;
+
+/**
+ * A tunable host for areas in storage which need potentially complex caching semantics
+ * which rely on some structural knowledge. In many implementations will be quite simple
+ *
+ * Created by ctsims on 1/25/2017.
+ */
+
+public interface QueryHandler<T> {
```
ctsims: done

ctsims:
Pointed back to master. Sorry about the branch name now not making a lot of sense but will be resolved once we pull anyway.

amstone326:
Hey, just approved this so we can move forward with merging. I did want to note though that I still think a medium-level refactor of this could be done to make it a lot more readable. I know that it's not practical for that to happen now or even soon, but since it's likely that people besides Clayton will be working with this code in the future, I think even a very belated refactor would still have a ton of value in terms of mitigating tech debt and significantly improving the ability for others to debug, read, and use it. 

@ctsims What are your thoughts on marking this as something to come back to and work on further when time is less crunched? I think something as simple as a trello card to track this that sits on the backlog for a while would serve as a good reminder, and since our future is likely to include many more tight project deadlines, it's probably good for us to establish a precedent of tracking tech debt of this nature (meaning that which arises from tight deadlines and resource crunches) more explicitly.

ctsims:
@amstone326 Definitely agree that it would be great to do a technical debt refactor here, and I think there are also a ton of low-hanging fruit wins to still be won from adding the old optimizations in the new framework. Definitely happy to put that on the roadmap

amstone326:
@ctsims Cool sounds great, just added a card to the roadmap: https://trello.com/c/9fuGh0U3/193-tech-debt-refactor-of-add-back-old-optimizations-to-query-planner-work

src/main/java/org/commcare/cases/query/queryset/DerivedCaseQueryLookup.java
```diff
@@ -0,0 +1,98 @@
...
+        if(set == null) {
+            return null;
+        }
+
+        List<Integer> returnSet = new Vector<>();
+        for(Integer i : rootLookup) {
+            Collection<Integer> matching = set.getMatchingValues(i);
+            if(matching == null) {
+                return null;
+            }
```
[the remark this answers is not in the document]
ctsims: This is such a good question that I've lost the context to answer adequately :/. 

I think the core difficulty is differentiating "optimized but empty" conditions and "optimization unavailable / failed" conditions. In theory (as you called out over voice) It should be the query set result (and not its contents) that would be null, but I don't have a great answer to why I doubled up on the conservative response here other than that it probably was hard to positively confirm that is always the case. 

It *might* be because we are caching the internals in a volatile way that can get punted from memory if things get tight, but I'm probably overthinking it.

PR #518, opened 2017-02-08
Kill preloaders

wpride:
This kills the outdated anachronism preloaders


src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1362,50 +1357,14 @@ public void populateDynamicChoices(ItemsetBinding itemset, TreeReference curQRef
...
-        }
-        // }
-    }
-
-    public boolean postProcessInstance() {
+    public void postProcessInstance() {
         if(!isCompletedInstance) {
             actionController.triggerActionsFromEvent(Action.EVENT_XFORMS_REVALIDATE, this);
         }
-        return postProcessInstance(mainInstance.getRoot());
```
amstone326: why was this call removed?

czue:
haha i remember when I tried to kill these on old cloudcare and some random person was using them. good riddance.

src/main/java/org/javarosa/core/model/FormDef.java
wpride: thanks, I actually should have removed the (now unused) method too. All this function was doing was triggering post processing for preloaders which are now gone so, unneeded.

PR #520, opened 2017-02-15
Make CasePurgeTest use standard Parameterized test infrastructure

amstone326:
Addresses http://manage.dimagi.com/default.asp?244430


src/test/java/org/commcare/test/utilities/CasePurgeTest.java
```diff
@@ -24,58 +28,86 @@
...
+
+        return parameters;
+    }
+
+    private static void addJSONArrayIfPresent(JSONObject root, int index, String key,
+                                              Object[] parameterSet) {
+        if (root.has(key)) {
+            parameterSet[index] = root.getJSONArray(key);
+        } else {
+            parameterSet[index] = null;
```
wpride: The values in the parameters array will be initialized to null so this else clause is unnecessary

src/test/java/org/commcare/test/utilities/CasePurgeTest.java
```diff
@@ -24,58 +28,86 @@
...
+            addJSONArrayIfPresent(root, i+1, jsonArrayKeys[i], parameters);
+        }
+
+        return parameters;
+    }
+
+    private static void addJSONArrayIfPresent(JSONObject root, int index, String key,
+                                              Object[] parameterSet) {
+        if (root.has(key)) {
+            parameterSet[index] = root.getJSONArray(key);
```
wpride: If you change this to root.get() you can add "name" to the jsonArrayKeys and have the work done on line 55 done in this method instead (and allow you to simplify the loop). parameterSet is an array of objects so you're not gaining anything by using getJSONArray() instead of get() (unless you're trying to ensure that those objects are JSONArrays in which case, ignore me)

amstone326: Hm, that's a good point. The issues with that are a) what you mentioned about wanting to make sure those objects are JSONArrays, and b) also that "name" is required while the rest are optional, so I'd want to handle it slightly differently anyway. I think given those 2 things, I'm going to leave it the way it is, despite being a bit convoluted.

src/test/java/org/commcare/test/utilities/CasePurgeTest.java
amstone326: Ha yes, good call

PR #522, opened 2017-02-19
Tests around confusing indexed fixture behavior with subelements

ctsims:
Tests to replicate/confirm what is seen in this ticket
http://manage.dimagi.com/default.asp?247786

Also fixes a bug in which the second (rather than first) element was used as a model template, which would cause problems with indexed fixtures that only have one element.


amstone326:

src/main/java/org/commcare/cases/instance/StorageInstanceTreeElement.java
```diff
@@ -273,7 +273,7 @@ protected Model getElement(int recordId) {
     }
 
     protected Model getModelTemplate() {
-        return storage.read(1);
```
wpride: Was this intentional? Seems like this would've caused some serious issues

ctsims: It did indeed cause serious issues. This is just tests for the bugfix, which I applied in a separate PR. There are a couple of issues with the new fixture format which need to be addressed (Nestle submitted them as feedback)

PR #524, opened 2017-02-22
Datetime fix

benrudolph:
Been trying to get this guy in for a while now. This PRs the change against the `nimbus` branch. Will PR the update to the nimbus branch separately

### DIFF
--- a/src/api/java/org/commcare/api/json/PromptToJson.java
+++ b/src/api/java/org/commcare/api/json/PromptToJson.java
@@ -145,10 +145,8 @@ public class PromptToJson {
                 obj.put("answer", answerValue.getDisplayText());
                 return;
             case Constants.DATATYPE_DATE_TIME:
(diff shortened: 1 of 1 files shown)


benrudolph:
@wpride

wpride:
What's this doing? Wasn't the calendar there for timezone handling?

benrudolph:
the current thing is taking a timezone and converting it to UTC. so when the answer is returned the strings don't match because one is in the normal timezone and the other is in utc. this just directly formats it in its current timezone so the answer matches.

current:
sent to server: `"2017-02-23T15:00:37+02:00"`
response: `"2017-02-23T13:00:37+00:00"`

now:
sent to server: `"2017-02-23T15:00:37+02:00"`
response: `"2017-02-23T15:00:37+02:00"`

essentially the only thing this does is validate that it's an actual date

PR #526, opened 2017-02-24
Move query planning classes into core from android

wpride:
Moving some classes into core for usage in formplayer

cross-request: https://github.com/dimagi/commcare-android/pull/1646


src/main/java/org/commcare/modern/engine/cases/IndexTable.java
```diff
@@ -0,0 +1,18 @@
+package org.commcare.modern.engine.cases;
+
+import org.commcare.cases.query.queryset.DualTableSingleMatchModelQuerySet;
+
+import java.util.*;
+
+/**
+ * Created by willpride on 2/21/17.
+ */
+public interface IndexTable {
```
ctsims: Only piece of feedback: I'd keep this as CaseIndexTable, and change the android side to AndroidCaseIndexTable or something similar.

PR #527, opened 2017-02-24
Printing from a detail

amstone326:
cross-request: https://github.com/dimagi/commcare-android/pull/1662

### DIFF
--- a/src/main/java/org/commcare/suite/model/Detail.java
+++ b/src/main/java/org/commcare/suite/model/Detail.java
@@ -477,7 +477,7 @@ public class Detail implements Externalizable {
 
     private boolean detailFieldsHaveIds() {
         for (DetailField field : this.fields) {
-            if (field.getFieldIdentifierRobust() == null) {
+            if (field.getPrintIdentifierRobust() == null) {
                 return false;
             }
         }
@@ -513,7 +513,7 @@ public class Detail implements Externalizable {
             for (int i = 0; i < fields.length; i++) {
                 if (entityForDetail.isValidField(i)) {
                     mapping.put(
-                            fields[i].getFieldIdentifierRobust(),
+                            fields[i].getPrintIdentifierRobust(),
                             new DetailFieldPrintInfo(fields[i], entityForDetail, i));
                 }
             }
--- a/src/main/java/org/commcare/suite/model/Detail.java
+++ b/src/main/java/org/commcare/suite/model/Detail.java
@@ -2,6 +2,7 @@ package org.commcare.suite.model;
 
 import org.commcare.cases.entity.Entity;
 import org.commcare.cases.entity.NodeEntityFactory;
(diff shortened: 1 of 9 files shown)


src/main/java/org/commcare/suite/model/Detail.java
```diff
@@ -443,4 +459,74 @@ public boolean evaluateFocusFunction(EvaluationContext ec) {
...
+                System.out.println("Invalid print template path provided for detail with id " + this.id);
+            }
+        }
+        return false;
+    }
+
+    private boolean detailFieldsHaveIds() {
+        for (DetailField field : this.fields) {
+            if (field.getPrintIdentifierRobust() == null) {
+                return false;
```
ctsims: This logic seems backwards to me. Shouldn't it be "if any field has an identifier, return true. Otherwise return false" rather than "If any field doesn't have an ID, return false"?

src/main/java/org/commcare/suite/model/DetailField.java
```diff
@@ -69,6 +70,16 @@
     public DetailField() {
     }
 
+    public String getPrintIdentifierRobust() {
+        // TODO: change this implementation once HQ work is done
```
ctsims: change the implementation to be what?

amstone326: Ah, sorry I forgot about this. So part of the work Jenny is doing to enable this on HQ is to make it so that if printing is enabled for a detail, every `<field>` comes down with a `print-id` attribute that is equivalent to the key the app builder will use in the print template for that field. For now, just to prevent us from having to manually add a `print-id` to the suite file before that work is done, I was using this hack to grab the template argument if the print-id isn't there, since that usually (but not always) corresponds to what the `print-id` will be. Ideally we'd leave this here until Jenny's work is done, because otherwise it's going to be a pain in the ass for us to do testing in the interim...

src/main/java/org/commcare/suite/model/Detail.java
amstone326: No actually, my intention in including this check was to make sure that we don't set `printEnabled` to be true unless every `DetailField` has a valid identifier that we'll be able to retreive in `Detail.populateMappingWithDetailFields()`.

ctsims: Wouldn't it be fine for some fields to not have an identifier if it wasn't actually needed by the template? I would have thought only field detail items included inside of the template would  *need* an ID.

I guess I was assuming that the presence of the Template was the "cue" for printing, and that if anything was missing to make it work we'd fail with an error (IE "Detail is missing required template field 'ID') rather than silently.

amstone326: Yea that's fair, I don't think I feel strongly about this. Will remove

PR #528, opened 2017-03-02
Allow Select questions to accept mismatched answers as empty strings

wpride:
Fixes https://manage.dimagi.com/default.asp?246509 for formplayer

cc @benrudolph 

@ctsims I'm not sure if there are implications I've not thought of to doing this. Currently in Android we handle this at the getAnswer() level; however, allowing the underlying model to handle mis-matched answers seems reasonable to me.

### DIFF
--- a/src/api/java/org/commcare/api/json/PromptToJson.java
+++ b/src/api/java/org/commcare/api/json/PromptToJson.java
@@ -152,15 +152,20 @@ public class PromptToJson {
             case Constants.DATATYPE_CHOICE:
                 Selection singleSelection = ((Selection) answerValue.getValue());
                 singleSelection.attachChoice(prompt.getQuestion());
-                obj.put("answer", ((Selection) answerValue.getValue()).getTouchformsIndex());
+                int singleOrdinal = singleSelection.getTouchformsIndex();
+                if (singleOrdinal > 0) {
+                    obj.put("answer", singleOrdinal);
+                }
                 return;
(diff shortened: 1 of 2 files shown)


amstone326:
Hey so I scanned through the ticket, but I'm not following why it was decided that we do want to be failing silently. Sounds like maybe you and Kriti came to that conclusion offline, can you explain a bit?

wpride:
It sounds like there was a bigger conversation about this previously involving Kriti/Clayton and possibly others where Kriti et al made the argument that failing hard on this is not tenable for FMs because valid select answers change too frequently. So rather than crashing they prefer the silent failure mode and null answer.

benrudolph:
seems fine to me, one downside is that it might be magically changing the value of a case. crashing hard is pretty painful though

wpride:
@benrudolph agreed, but I already argued this (and lost) with Kriti and @czue 

Android already does this (silent failing) behavior so we're just aligning with them

czue:
i'm very good at making this argument if you want to duke it out again @benrudolph ! :fist_oncoming:

benrudolph:
haha i'm actually not sure which side i'm on. was just noting both a pro and a con (magically changing the case and crashing hard). :upside_down_face:

benrudolph:
_if_ this were a debate about code, i would for sure value data consistency over ease of data migrations #couchvssql. but i think that is fairly divorced from app building

PR #536, opened 2017-03-14
Implement user-configurable 'blanks' preference for DetailField sorting

amstone326:
Implements Jen's request from http://manage.dimagi.com/default.asp?247874 -- When a detail has a sort property, users can now configure whether cases for which the sort property is blank show up at the top of the list or the bottom. This is specified via a "blanks" attribute on a <sort> element which can take values "first" or "last". This configuration is independent of the specified sort direction (i.e. if blanks="last", blanks will always show up at the end of the case list, whether the sort order is set to ascending or descending).

@orangejenny One important question for you -- You had said to implement this such that if the "blanks" property isn't provided, it defaults to putting blanks at the top, which is what I've done currently. However, I realized via testing that the current behavior is a little more complicated than that. Right now, if the sort order is "ascending", then blanks do default to the top, but if the sort order is "descending", then blanks default to the bottom. I think this was incidental to how the code was written rather than intentional in any way. Should I change my implementation to maintain this default behavior?

### DIFF
--- a/src/main/java/org/commcare/cases/entity/EntitySorter.java
+++ b/src/main/java/org/commcare/cases/entity/EntitySorter.java
@@ -60,7 +60,7 @@ public class EntitySorter implements Comparator<Entity<TreeReference>> {
 
         boolean showBlanksLast = detailFields[index].showBlanksLastInSort();
         // The user's 'blanks' preference is independent of the specified sort order, so don't
-        // worry about the 'reverse' parameter here
+        // factor in the 'reverse' parameter here
         if (a1.equals("")) {
             if (a2.equals("")) {
                 return 0;
--- a/src/main/java/org/commcare/cases/entity/EntitySorter.java
+++ b/src/main/java/org/commcare/cases/entity/EntitySorter.java
@@ -28,8 +28,8 @@ public class EntitySorter implements Comparator<Entity<TreeReference>> {
     @Override
     public int compare(Entity<TreeReference> object1, Entity<TreeReference> object2) {
         for (int aCurrentSort : currentSort) {
-            boolean reverseLocal = (detailFields[aCurrentSort].getSortDirection() == DetailField.DIRECTION_DESCENDING) ^ reverseSort;
-            int cmp = (reverseLocal ? -1 : 1) * getCmp(object1, object2, aCurrentSort);
+            boolean reverse = (detailFields[aCurrentSort].getSortDirection() == DetailField.DIRECTION_DESCENDING) ^ reverseSort;
+            int cmp = getCmp(object1, object2, aCurrentSort, reverse);
             if (cmp != 0) {
                 return cmp;
             }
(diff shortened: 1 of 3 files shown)


amstone326:
NOTE TO SELF: Update the suite spec to reflect this option once this is released.

orangejenny:
> Right now, if the sort order is "ascending", then blanks do default to the top, but if the sort order is "descending", then blanks default to the bottom.

I think that's fine, not worth spending effort to make that consistent. Users who care where blanks go can use the new option.

amstone326:
@orangejenny Sorry I think the way I worded this was confusing. "Right now" meant "the state of the world before this PR". The way I've currently implemented this PR will _change_ this so that if no "blanks" attribute is specified, blanks will just always show up on top, which I had done based on your original comment. But it probably makes more sense to make the default behavior the same as it was before this change; that isn't a lot of work, I just wanted to double check with you that that seemed right. (Also, I actually think that defaulting to blanks at the top is the least good of all options because it just intuitively seems like that's not usually what a user would want).

orangejenny:
Ah, I did misunderstand, although re-reading I still don't think it's worth spending much time on. But if it's low effort to keep 2.35 behavior consistent with pre-2.35, that sounds good.

orangejenny:
@amstone326 Is it fine if the `blanks` attribute is provided to pre-2.35 CommCare? The simplest way to implement this on the HQ side was to make the UI 2.35+, but always provide `blanks` when writing new suite files. For pre-2.35, it writes `blanks="top"` if direction is ascending, or `blanks="bottom"` if direction is descending, to match the expected pre-2.35 behavior. Does that sound right?

amstone326:
@orangejenny:

- It's definitely fine if the `blanks` attribute is provided to pre-2.35 CommCare, because it will just get ignored. So just to make sure I'm understanding-- on any suite files generated after your change is merged, there will be a `blanks` attribute present whether or not the user has specified one?

- The default behavior you described is correct and sounds like a good way to handle it. 

- Small point, but I implemented the 2 values that the `blanks` attribute can take to be `"first"` and `"last"`, as per what was originally spec'd in the ticket. I can easily change that, but just wanted to double check first that `top` and `bottom` is what you've current implemented (as opposed to just a typo or something).

orangejenny:
I misspoke: `first` and `last` are used in the suite file, as specced in the ticket. The ui says top & bottom, so I got confused.

orangejenny:
> on any suite files generated after your change is merged, there will be a blanks attribute present whether or not the user has specified one?

Yes, exactly.

amstone326:
@orangejenny Ok perfect, think we're all set then!

ctsims:
@amstone326 This looks great! Merging.

in
https://github.com/dimagi/commcare-core/wiki/Suite20

btw, we actually use a "SINCE" thing to flag when things are introduced, so feel free to write about the new behavior now.

amstone326:
@ctsims Nice, thanks for the tip

PR #537, opened 2017-03-14
Enable Bulk Processing Restore Data

ctsims:
cross-request: https://github.com/dimagi/commcare-android/pull/1666

### DIFF
--- a/src/main/java/org/commcare/xml/bulk/BulkElementParser.java
+++ b/src/main/java/org/commcare/xml/bulk/BulkElementParser.java
@@ -22,6 +22,8 @@ import java.util.TreeMap;
  * A bulk element parser reads multiple types of the same transaction together into a buffer of
  * TreeElements before performing a bulk processing step.
  *
+ * NOTE: This parser will currently lose debugging information v. using a streamed processor,
+ * since the XML locations of potential errors are buffered.
  *
  * An implementing class should organize its implementation into the following steps:
  *
@@ -53,6 +55,10 @@ public abstract class BulkElementParser<T> extends TransactionParser<TreeElement
         super(parser);
     }
 
+    /**
+     * Sets the count at which the parser should trigger processing the current buffer
+     * @param newTriggerCount
+     */
     protected void setBulkProcessTrigger(int newTriggerCount) {
         this.bulkTrigger = newTriggerCount;
     }
@@ -83,7 +89,6 @@ public abstract class BulkElementParser<T> extends TransactionParser<TreeElement
         performBulkWrite(writeLog);
         clearState();
     }
-
 
     protected void clearState() {
         currentBulkElementBacklog.clear();
--- a/src/main/java/org/commcare/xml/bulk/BulkElementParser.java
+++ b/src/main/java/org/commcare/xml/bulk/BulkElementParser.java
@@ -1,0 +1,107 @@
+package org.commcare.xml.bulk;
+
+import org.commcare.data.xml.TransactionParser;
+import org.javarosa.core.model.instance.TreeElement;
+import org.javarosa.xml.TreeElementParser;
+import org.javarosa.xml.util.InvalidStructureException;
+import org.javarosa.xml.util.UnfullfilledRequirementsException;
+import org.kxml2.io.KXmlParser;
+import org.xmlpull.v1.XmlPullParserException;
+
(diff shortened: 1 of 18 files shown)


src/main/java/org/commcare/xml/bulk/BulkProcessingCaseXmlParser.java
```diff
@@ -0,0 +1,308 @@
...
+        currentBulkReadSet.add(caseId);
+    }
+
+    @Override
+    protected void preParseValidate() throws InvalidStructureException {
+        checkNode("case");
+    }
+
+    @Override
+    protected void processBufferedElement(TreeElement bufferedTreeElement, Map<String, Case> currentOperatingSet, LinkedHashMap<String, Case> writeLog) throws InvalidStructureException {
```
amstone326: It looks like the entire contents of this method are identical to that of `CaseXmlParser.parse()`. Can you have both use a common helper instead?

amstone326: Actually it looks like this is true for almost all of the methods in this class. I think this would feel a lot cleaner if all of those implementations were merged in some way. I don't like the idea that someone could go update `CaseXmlParser` down the line and not realize that this other place exists that should get the same changes. It's unfortunate that this class can't just inherit from `CaseXmlParser` as well, but some shared static methods don't seem terrible, since I think you should only have to pass the `parser` into each one.

ctsims: Hey,

It is identical except for the fact that one parses a streamed XML element and the other parses our intermediate buffered tree element object, and that makes the parsing very different structurally.

I looked at how to try to merge the two and it was super, super ugly, since you "consume" elements in the other parser, and in this one you query them from an in-memory object. The other parser is also capable, for instance, of reporting errors based on the current location in the parse, which this can't do.

I'll make a ticket to try again to merge the implementations, but it is going to be super ugly no matter what due to those differences. It likely makes sense once we have confidence in this parser to make the "default" parser this parser with a "trigger" buffer of 1 rather than maintaining both.

amstone326: Ah ok, if that last strategy is our medium-term plan then I'm fine with this state of things for now. Is there somewhere it makes sense to put a reminder to ourselves to do look at doing that down the line?

src/main/java/org/commcare/xml/bulk/BulkElementParser.java
```diff
@@ -0,0 +1,173 @@
...
+        if (currentBulkReadCount > bulkTrigger) {
+            processCurrentBuffer();
+        }
+        return subElement;
+    }
+
+    public void processCurrentBuffer() throws IOException, XmlPullParserException,
+            InvalidStructureException {
+        currentOperatingSet = new HashMap<>();
+        performBulkRead(currentBulkReadSet, currentOperatingSet);
```
amstone326: @ctsims Just noting this is here too in case it got missed when you reviewed comments

src/main/java/org/commcare/xml/bulk/BulkProcessingCaseXmlParser.java
ctsims: Updated a Trello ticket for tech debt and left a warning about not doing additional work in CaseXML without moving parsing to the new parser.

src/main/java/org/commcare/xml/bulk/BulkElementParser.java
ctsims: What you're describing is intentional from the code as is. The breakdown is basically that the implementing class needs to provide mechanisms to process data in multiple different states (IE: Single XML element, bulk backlog, etc), and the goal of this Inversion of Control is to eliminate the need for the implementing class to understand the "state" of the parse, which adds a fair amount of complexity. 

Essentially I didn't want the implementing class to attempt to maintain any state on its side, so injecting the neccesary elements here prevents needing to do so. I only did this as a subclass for convenience for now, but I think in the future a better pattern will be for an implementing class to provide implementations for those 5 methods _without_ subclassing this class, and functioning statelessly, but doing that work drifted further from the current implementation of the CaseXML parser than I was comfortable doing in one go.

PR #538, opened 2017-03-20
Convert to UTC

benrudolph:
@wpride this converts the datetime passed in to UTC


wpride:
We should just move these classes into Formplayer repo, doesn't make sense to keep here IMO

PR #539, opened 2017-03-24
XPath ID Support functions

ctsims:
Functions which support serial ID generation schemes

### DIFF
--- a/src/main/java/org/javarosa/core/util/CompressingIdGenerator.java
+++ b/src/main/java/org/javarosa/core/util/CompressingIdGenerator.java
@@ -4,7 +4,7 @@ package org.javarosa.core.util;
  * Created by ctsims on 3/24/2017.
  */
 
-public class CompressingIdGenreator {
+public class CompressingIdGenerator {
 
 
     /**
@@ -38,15 +38,15 @@ public class CompressingIdGenreator {
                                                     String bodySymbols,
                                                     int bodyDigitCount) {
 
-        if(growthSymbols.length() == 0 || leadSymbols.length() == 0) {
+        if (growthSymbols.length() == 0 || leadSymbols.length() == 0) {
             throw new IllegalArgumentException(String.format(
                     "Invalid Symbol Space for ID Compression, growth and lead set must both" +
                             " contain at least one symbol" +
                             "\nG[%s] | L[%s] | B[%s]",growthSymbols, leadSymbols, bodySymbols));
         }
 
-        for(char c : growthSymbols.toCharArray()) {
-            if(leadSymbols.indexOf(c) != -1) {
+        for (char c : growthSymbols.toCharArray()) {
+            if (leadSymbols.indexOf(c) != -1) {
                 throw new IllegalArgumentException(String.format(
                         "Illegal growth/lead symbol space. The character %s was found in both" +
                                 " spaces.", c));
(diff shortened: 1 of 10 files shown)


src/main/java/org/javarosa/xpath/expr/XPathJoinChunkFunc.java
```diff
@@ -0,0 +1,66 @@
...
+import org.javarosa.core.model.condition.EvaluationContext;
+import org.javarosa.core.model.instance.DataInstance;
+import org.javarosa.xpath.XPathArityException;
+import org.javarosa.xpath.XPathNodeset;
+import org.javarosa.xpath.parser.XPathSyntaxException;
+
+public class XPathJoinChunkFunc extends XPathFuncExpr {
+    public static final String NAME = "join-chunked";
+    // one or more arguments
+    private static final int EXPECTED_ARG_COUNT = -1;
```
amstone326: this is inconsistent with the arg count check below

src/main/java/org/javarosa/xpath/expr/XPathJoinChunkFunc.java
```diff
@@ -0,0 +1,66 @@
+package org.javarosa.xpath.expr;
+
+import org.javarosa.core.model.condition.EvaluationContext;
+import org.javarosa.core.model.instance.DataInstance;
+import org.javarosa.xpath.XPathArityException;
+import org.javarosa.xpath.XPathNodeset;
+import org.javarosa.xpath.parser.XPathSyntaxException;
+
+public class XPathJoinChunkFunc extends XPathFuncExpr {
```
amstone326: can you add class-level comments to this and `XPathIdCompressFunc` describing their intended usage/purpose

src/main/java/org/javarosa/xpath/expr/XPathJoinChunkFunc.java
amstone326: ah now I see that it's because this class overrides `validateArgCount()`. Could you update the comment above this line to just explain that? I think the current one is stale

src/main/java/org/javarosa/core/util/CompressingIdGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+     * space.
+     *
+     * Note that there is *always* exactly one "Lead" digit. It is acceptable for the count of
+     * "Body" digits to be 0.
+     *
+     * @param input a number to be encoded by the scheme provided. Must be a positive integer
+     * @param bodyDigitCount the fixed number of "Body" digits that will be used to encode the
+     *                       input value
+     */
+    public static String generateCompressedIdString(long input,
```
amstone326: I'm not familiar with this encoding schema; curious whether this based on some established protocol, or just something you created?

src/test/java/org/javarosa/core/util/test/CompressingIdTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Test
+    public void overlapTests() {
+        HashSet<String> generatedIds = new HashSet<>();
+        for (int i = 0; i < 100; ++i) {
+            for (int j = 0; j < 100; ++j) {
+                for (int k = 0; k < 100; ++k) {
+                    String iPart = CompressingIdGenerator.generateCompressedIdString(i, GROWTH_LIMITED, LEAD_LIMITED, BODY_LIMITED, 1);
+                    String jPart = CompressingIdGenerator.generateCompressedIdString(j, GROWTH_LIMITED, LEAD_LIMITED, BODY_LIMITED, 1);
+                    String kPart = CompressingIdGenerator.generateCompressedIdString(k, GROWTH_LIMITED, LEAD_LIMITED, BODY_LIMITED, 1);
```
esoergel: You should be able to move `iPart` and `jPart` up in to their respective loops and avoid a bunch of extra calls.

esoergel: Also, I don't think I get what this test is checking, assuming that generateCompressedIdString behaves the same way each call, what does this add above testing a single loop from 1 to 100?

src/main/java/org/javarosa/xpath/expr/XPathJoinChunkFunc.java
ctsims: documented for good measure (I copied this code from join and that's how it documented this)

src/main/java/org/javarosa/core/util/CompressingIdGenerator.java
ctsims: I will send you the spec (it's project specific

src/main/java/org/javarosa/xpath/expr/XPathJoinChunkFunc.java
ctsims: We don't appear to document XPath functions that way anymore, I updated the json doc that appears to be where these docs go now.

ctsims: (updated after feedback that is, there was no documentation on the first pass.

src/test/java/org/javarosa/core/util/test/CompressingIdTest.java
ctsims: the lengths could be broken down better (IE: More specific to 1 or 2 growth digits). The test is to ensure that you can combine the requisite ID elements together without them conflicting when they expand from multiple sides, like in the 
```
STEVE-2-1001
STEVE-21-001
```
example from the spec

esoergel: Ah, right, thanks!

PR #544, opened 2017-04-05
throw exception when download fails

wpride:
More direct error when CCZ download fails

@benrudolph


src/cli/java/org/commcare/util/engine/CommCareConfigEngine.java
```diff
@@ -140,16 +140,30 @@ protected String downloadToTemp(String resource) {
...
+                try {
+                    if (fos != null) {
+                        fos.close();
+                    }
+                    if (bis != null) {
+                        bis.close();
+                    }
+                    conn.disconnect();
+                } catch (IOException ex) {
+                    // Log error writing file and bail out.
```
benrudolph: did you want to implement this?

PR #549, opened 2017-04-17
Allow both repeat flavors

wpride:
cross-request: https://github.com/dimagi/formplayer/pull/359

### DIFF
--- a/src/main/java/org/javarosa/form/api/FormEntryModel.java
+++ b/src/main/java/org/javarosa/form/api/FormEntryModel.java
@@ -46,12 +46,6 @@ public class FormEntryModel {
      */
     public static final int REPEAT_STRUCTURE_NON_LINEAR = 2;
 
-    /**
-     * Form can have both types of repeat groups
-     */
-    public static final int REPEAT_STRUCTURE_MIXED = 4;
-
-
     public FormEntryModel(FormDef form) {
         this(form, REPEAT_STRUCTURE_LINEAR);
     }


src/main/java/org/javarosa/form/api/FormEntryModel.java
```diff
@@ -466,10 +465,8 @@ private void incrementHelper(Vector<Integer> indexes, Vector<Integer> multiplici
                 multiplicities.setElementAt(0, i);
                 elements.setElementAt(parent.getChild(curIndex + 1), i);
 
-                if (repeatStructure == REPEAT_STRUCTURE_NON_LINEAR) {
-                    if (elements.lastElement() instanceof GroupDef && ((GroupDef)elements.lastElement()).getRepeat()) {
-                        multiplicities.setElementAt(TreeReference.INDEX_REPEAT_JUNCTURE, multiplicities.size() - 1);
-                    }
+                if (elements.lastElement() instanceof GroupDef && ((GroupDef)elements.lastElement()).getRepeat()) {
```
wpride: Here is the fix

@ctsims @benrudolph

ctsims: @wspride Can you explain why this is a needed change at the formentry model level? CommCare Android does this currently, so i'm surprised to see that a change is needed in core, rather than at the platform/presentation layer.

amstone326:
@wpride Can this be closed?

ctsims:
@amstone326 @wpride can be closed from the UATBC side. We worked around by defining custom XML to emulate the non-linear repeat functionality.

PR #550, opened 2017-04-18
Fix back for autoselect

amstone326:
Fix and regression tests for http://manage.dimagi.com/default.asp?252170.


src/test/java/org/commcare/backend/session/test/BasicSessionNavigationTests.java
```diff
@@ -29,119 +29,119 @@ public void setUp() throws Exception {
     @Test
```
amstone326: Changes in this file are just because I realized that we had the order of arguments to `assertEquals()` backwards; it should be `(expected, actual)`

PR #558, opened 2017-04-27
Return instead of crash when rootLookup is null

wpride:
@ctsims L10K's app was crashing on one case list where this value was null. Returning null (instead of NPE crashing) seems to have resolved the issue. Root cause might be that this case's owner was deleted. Not sure if this will break other things so Don't Pull until confirmed this change is safe.

### DIFF
--- a/src/main/java/org/commcare/cases/query/queryset/DerivedCaseQueryLookup.java
+++ b/src/main/java/org/commcare/cases/query/queryset/DerivedCaseQueryLookup.java
@@ -60,7 +60,7 @@ public abstract class DerivedCaseQueryLookup implements QuerySetLookup {
     public List<Integer> performSetLookup(TreeReference lookupIdKey, QueryContext queryContext) {
         List<Integer> rootLookup = root.performSetLookup(lookupIdKey, queryContext);
         ModelQuerySet set = getOrLoadCachedQuerySet(queryContext);
-        if(set == null) {
+        if (set == null || rootLookup == null) {
             return null;
         }
 


ctsims:
Hey.

@wpride So I'm not comfortable returning in these situations unless it's *very* clear what the reason is, and if the reason is what you are saying (the parent case doesn't exist), I think we should be very explicit to *only* be returning back out because we've identified that exact case. 

With optimizations like this it's very, very dangerous for us to not-crash for these kinds of failures, because if we just return out we can either miss data (if we return back up the chain with 'No Answer') that results in inconsistent state, or we can just miss the optimization silently.

Either output creates significantly more work for us later when either the code is running slowly but we assume it's optimized, or when we can't understand why we aren't seeing data, since the optimization is turned on. 

As an example, this was also the area of the code where this optimization was failing earlier on webapps and it caught a serious bug in the optimization. If we had just returned out Null in that case, we would have completely failed to understand why the code wasn't performing (or was missing cases).

As such, if we have PR's like this I would very much prefer that we:
A) Include tests that guarantee that the code performs the same way with the optimization enabled and disabled (which catches the issues where the optimization misses data)
B) Very deliberately tag in-code and preferably in a log somewhere that something we expect should work didn't. That way if we are looking over this optimization trace later we could see a message like "Unexpected Failure to Apply Model Set Transform" which tips us off as to why the code is executing the way it is.

wpride:
@ctsims Cool, sounds good to me. Wasn't sure the possible repercussions of returning null here.  I think I (and I assume @amstone326 ) could use an in-person overview of how the Query Planner optimizations work when we get back with a particular eye towards debugging these sorts of issues. When we get an NPE here I'm not sure how to figure out what's going on/where we are in the query from the context, or when returning null is acceptable (IE when `set` is null) vs. not.

wpride:
@ctsims Cool, sounds good to me. I think I (and I assume @amstone326 ) could use an in-person overview of how the Query Planner optimizations work when I get back with a particular eye towards debugging these sorts of issues. When we get an NPE here I'm not sure how to figure out what's going on/where we are in the query from the context. Also not clear why its acceptable to return null if `set` is null here, but not

ctsims:
@wpride The other thing we could do if we know that's specifically a bad state to get into taht will result in a null pointer is catch null here and rethrow as a deliberate runtime exception with context. That definitely would be a positive change to the code if there's no context where that code path *wouldn't* work anyway.

PR #561, opened 2017-05-09
Don't derive print template path until using it

amstone326:


src/main/java/org/commcare/suite/model/Detail.java
```diff
@@ -465,8 +465,8 @@ private boolean templatePathValid(String templatePathProvided) {
             return true;
         } else if (templatePathProvided != null) {
             try {
-                this.derivedPrintTemplatePath =
-                        ReferenceManager.instance().DeriveReference(templatePathProvided).getLocalURI();
+                ReferenceManager.instance().DeriveReference(templatePathProvided).getLocalURI();
```
orangejenny: Does this still need to be here? Is it to surface errors, if there are any at this point?

amstone326: Yes, definitely still want this here because it controls whether this detail is viewed as able to be printed. If this derivation fails, we won't give the user the option to print in the UI

PR #564, opened 2017-05-23
Fix persistent case tile in form linking

amstone326:
Fix for https://manage.dimagi.com/default.asp?254740. Similar to the solutions of some other tickets we've had over the last year or so, we just needed to be checking for the UNKNOWN_TYPE as well as the explicit desired type, and if it is unknown, compare against the result of guessUnknownType().


ctsims:
@amstone326 Woah, that was a quick turnaround! You the bomb.

PR #576, opened 2017-06-01
2.36 hotfix - Quarantine on read failure

amstone326:
Addresses http://manage.dimagi.com/default.asp?255101.

We think this should fix https://manage.dimagi.com/default.asp?254634, but I have thus far been unable to replicate the error they are getting, so not positive. Clayton is going to take a stab at it, but either way this is clearly an improved state, so I think we should go forward with it even if we can't verify it ourselves.

### DIFF
--- a/src/main/java/org/javarosa/core/io/StreamsUtil.java
+++ b/src/main/java/org/javarosa/core/io/StreamsUtil.java
@@ -93,8 +93,8 @@ public class StreamsUtil {
      * Writes input stream to output stream in a buffered fasion, but doesn't
      * close either stream.
      */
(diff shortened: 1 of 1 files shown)


src/main/java/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -95,11 +95,24 @@ public OutputIOException(IOException internal) {
      */
     public static void writeFromInputToOutputUnmanaged(InputStream is,
                                                        OutputStream os) throws IOException {
```
ctsims: Can you split this into "throws InputIOExcepton, OutputIOException"? It's helpful for us to signal that

PR #578, opened 2017-06-01
Formplayer parity

wpride:
@amstone326 removing some now-unused functions


src/main/java/org/commcare/core/parse/ParseUtils.java
```diff
@@ -6,8 +6,8 @@
 import org.javarosa.xml.util.UnfullfilledRequirementsException;
 import org.xmlpull.v1.XmlPullParserException;
 
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
+import java.nio.charset.StandardCharsets;
```
amstone326: the `StandardCharsets` import is unused, and I think we usually prefer to be implicit about imports instead of using `.*`, especially if it's only 2 classes from that package

src/main/java/org/commcare/core/parse/CommCareTransactionParserFactory.java
```diff
@@ -55,7 +55,7 @@
     private final Set<String> processedFixtures = new HashSet<>();
 
     protected final UserSandbox sandbox;
-    private boolean isBulkProcessingEnabled = false;
+    protected boolean isBulkProcessingEnabled = false;
```
amstone326: Why did this need to change? I don't see where it's being used

wpride: This is used in the Formplayer implementation of this ParserFactory

src/main/java/org/commcare/core/parse/ParseUtils.java
wpride: Optimized

PR #583, opened 2017-06-02
Use command's EvaluationContext

wpride:
fixes https://manage.dimagi.com/default.asp?254666

When evaluating a menu's displayables we should be getting the properly generated new EvaluationContext rather than using the same as the parents.


wpride:
Fixes bug with references to the session from form display conditions

PR #585, opened 2017-06-05
Add a few useful methods

benrudolph:
@wpride so this gets into core as well


wpride:
@benrudolph great call - I should be better about proactively merging `formplayer` changes into master
 Also FYI the mobile team uses the "Ready to Review" tag to signify the same so people usually won't merge until that tag is applied

benrudolph:
ah thanks, makes sense will do that in the future

PR #588, opened 2017-06-06
Test downgrading the model cache size

ctsims:
This is a quick test to see if it will fix

https://manage.dimagi.com/default.asp?255503

or https://manage.dimagi.com/default.asp?255507

Should never be pulled until it's adjusted to set the targets dynamically based on heap or something, or generally allows some tuning to retain the original functionality.


amstone326:
@ctsims Can this be closed?

PR #590, opened 2017-06-07
add support for relevancy conditions on details

amstone326:


src/main/java/org/commcare/suite/model/Detail.java
```diff
@@ -82,6 +84,8 @@
 
     private String printTemplatePath;
 
+    private XPathExpression parsedRelevancyExpression;
```
wpride: Does this need to be handled in `readExternal` / `writeExternal` ?

amstone326:
@wpride Really good catch, thank you. Also note this PR has a cross-request: https://github.com/dimagi/commcare-android/pull/1735

PR #591, opened 2017-06-15
Handle calcs that do not return nodeset

benrudolph:
@wpride dumb mistake, this allows for values that aren't xpath nodesets to work like `true()`. (tests in formplayer pr)


wpride:
@benrudolph should this go into master too?

benrudolph:
yeah

benrudolph:
will update

PR #593, opened 2017-06-21
Wsp/fpu

wpride:


wpride:
@benrudolph

src/main/java/org/commcare/suite/model/Detail.java
```diff
@@ -145,6 +145,15 @@ public Detail(String id, DisplayUnit title, String nodeset,
         } else {
             numEntitiesToDisplayPerRow = 1;
         }
+
+        if (relevancy != null) {
+            try {
+                this.parsedRelevancyExpression = XPathParseTool.parseXPath(relevancy);
+            } catch (XPathSyntaxException e) {
+                e.printStackTrace();
```
benrudolph: don't love these

amstone326: Don't love what?

wpride: I think the silent erroring, but I think its the correct thing to do here

benrudolph: no this being a print statement rather than logging. (also this does error hard, if you want it to error silently i think the below needs to update)

PR #594, opened 2017-06-21
Dev work for INDDEX

amstone326:
Changes implemented:
- Sorting of answer choices in itemset select questions
- Add a `sort()` function in XPath (documentation with detailed specifications added [here](https://confluence.dimagi.com/display/commcarepublic/CommCare+Functions#CommCareFunctions-sort))
- Make it possible to index the same column name in multiple fixtures


amstone326:
@ctsims fyi that I'm sending an .apk with this code in it to Nick and Codou today, so they can start testing out how things work with the changes. If code review results in functional changes, I can update them accordingly, but wanted to get them started with something for now.

amstone326:
Release Note / announce: 

- Introduces an XPath sort() function which can sort a space-separated list of strings (full documentation [here](https://confluence.dimagi.com/display/commcarepublic/CommCare+Functions#CommCareFunctions-sort)
- Adds the ability to sort answer choices in a Multiple Choice Lookup Table question

src/main/java/org/commcare/modern/database/DatabaseIndexingUtils.java
```diff
@@ -33,7 +33,7 @@ public static String indexOnTableCommand(String indexName,
     }
 
     private static String makeIndexingStatement(String tableName, String index) {
-        String indexName = index + "_index";
+        String indexName = "fixture_" + tableName + "_" + index + "_index";
```
amstone326: This is the fix for the indexing bug

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1333,28 +1330,42 @@ public void populateDynamicChoices(ItemsetBinding itemset, TreeReference curQRef
...
-            choice.setIndex(i);
-            if (itemset.copyMode)
-                choice.copyNode = copyNode;
+        SelectChoice choice = new SelectChoice(label, value != null ? value : "dynamic:" + index,
+                itemset.labelIsItext);
+        choice.setIndex(index);
+
+        if (itemset.copyMode) {
+            choice.copyNode = copyNode;
+        }
```
ctsims: Can you confirm that everything in this file up to here is just refactor? The diffs came out surprisingly hard to read...

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1465,13 +1417,74 @@ private void parseItemset(QuestionDef q, Element e) {
...
+        itemset.copyMode = false;
+    }
+
+    private void parseItemsetSortElement(Element child, ItemsetBinding itemset) {
+        String sortXpathString = child.getAttributeValue("", REF_ATTR);
+        if (sortXpathString == null) {
+            throw new XFormParseException("<sort> in <itemset> requires 'ref'");
+        }
+
+        XPathPathExpr sortPath = XPathReference.getPathExpr(sortXpathString);
```
ctsims: Is it easy enough to make this an XPath expression, rather than a reference? The other two are references because that's part of the XForms specification, but there's no specific reason that we have to be that restrictive.

If it adds a lot of work to make this an expression rather than a path references, no worries for now, just seemed like a convenient capacity to add if easy.

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    protected Object evalBody(DataInstance model, EvaluationContext evalContext, Object[] evaluatedArgs) {
+        List<String> sortedList;
+        if (evaluatedArgs.length == 1) {
+            sortedList = sortSingleList(FunctionUtils.toString(evaluatedArgs[0]), true);
+        } else if (evaluatedArgs.length == 2) {
+            Object lastArg = evaluatedArgs[1];
+            if (lastArg instanceof Boolean) {
```
ctsims: I think this behavior is actually potentially harmful and could lead to issues if an expression which is comparing something ends up casting to an int or something.

If someone wants to sort by the original list, they can always just include it twice in a row, but in a non-typed function spec language, doing method dispatch dynamically on type doesn't feel like a good practice.

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+        });
+    }
+
+    private List<String> sortListByOtherList(String s1, String s2, boolean ascending) {
+        List<String> targetListItems = stringToList(s1);
+        List<String> comparisonListItems = stringToList(s2);
+
+        if (targetListItems.size() != comparisonListItems.size()) {
+            throw new XPathTypeMismatchException("Length of lists passed to sort() must match");
```
ctsims: It would possibly be worth including the list contents here, since we already have the strings. Or describing the lengths. Generally providing any cue that would let someone track why things did/didn't map.

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
```diff
@@ -0,0 +1,146 @@
...
+            } else {
+                correspondingStrings = new ArrayList<>();
+                stringMapping.put(comparisonString, correspondingStrings);
+            }
+            correspondingStrings.add(targetString);
+        }
+        return stringMapping;
+    }
+
+    private static List<String> stringToList(String s) {
```
ctsims: There's already a utility method to do this and the method below that we use in other xpath expressions. WE should reuse them for consistency.

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
```diff
@@ -0,0 +1,146 @@
...
+
+        Map<String, List<String>> stringMapping =
+                createMappingFromComparisonToTarget(comparisonListItems, targetListItems);
+
+        List<String> sortedComparisonList = sortSingleList(s2, ascending);
+        List<String> sortedTargetList = new ArrayList<>();
+        String previousComparisonString = "";
+        for (int i = 0; i < sortedComparisonList.size(); i++) {
+            String stringInSortedList = sortedComparisonList.get(i);
+            if (stringInSortedList.equals(previousComparisonString)) {
```
ctsims: We may need to sort lists where there are multiple keys on either side of the sorting, and i'm pretty concerned with how this is going to deal with those.

It's pretty annoying to handle a lot of string manipulations, I think the safest option would be if, rather than making the map above, you made a list of tuples [targetListItem, comparisonListItem] (we can use Pair<> in the core code now) and you made a comparator that just sorted by the second member of the tuple. That should handle the case where there are string matching duplicates and I think will likely make the code a bit clearer.

src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
```diff
@@ -99,6 +99,24 @@ public void testTypeCoercion(){
...
 
     @Test
+    public void sortTests() {
+        // simple sort
+        testEval("sort('commcare is the best tool ever', false())", null, null, "tool the is ever commcare best");
+
+        // sort 2nd list by 1st
+        testEval("sort('2222 5555 9999 1111', 'd b c a', true())", null, null, "1111 5555 9999 2222");
+        testEval("sort('a b c d e f', '4 2 1 5 3 2', false())", null, null, "d a e f b c");
+        testEval("sort('c c z f z f', '4 2 1 5 3 2', true())", null, null, "z c f z c f");
```
ctsims: To test above, I'd include some examples where there are multiple items with matching sort orders int he second list, IE:

a b c d : 3 2 1 2 = c b d a OR c d b a

src/main/java/org/javarosa/core/model/FormDef.java
amstone326: confirmed, and sorry about that, I should have done it separately

src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
amstone326: Hey @ctsims sorry I'm a little confused, the right side lists of a lot of these tests already do have repeats. `'4 2 1 5 3 2'` has 2 in it twice, and the way it handles deciding on the order for the items on the left side which both match to `2` is to sort by those items themselves in the direction of the sort. So that's why:

in `sort('a b c d e f', '4 2 1 5 3 2')`, `b` comes before `f` in the output  ( `c b f e a d`)
but 
in `sort('a b c d e f', '4 2 1 5 3 2', false())`, `f` comes before `b` in the output ( `d a e f b c`)

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
amstone326: I'm pretty sure what I've written here already handles what you're talking about (see my other comment below about the test cases). If you think I'm missing something more here after you read that, let me know and we can talk in person

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
amstone326: I looked for these and didn't find them (including in either of our `StringUtils` classes or in `ArrayUtilities`). If you know where they are, let me know and I'll switch over

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
amstone326: (Although I agree that doing it with `Pair<>` would be cleaner regardless, so I can do that refactor if I have time, but first I want to confirm first that the we're on the same page conceptually)

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
amstone326: Hm, fair point on the dynamic dispatching, but I don't agree that it's preferable to make people include the original list twice in a row if they want to sort a single list; I think that's the sort of thing that seems fine to us, but would mean that the spec for the function would make it very unclear to average users how they would ever use it to sort a single list. I would rather solve this by making it so that the directional argument is always required. Is that an OK solution?

src/main/java/org/javarosa/xform/parse/XFormParser.java
amstone326: Hey yea I think I'm going to leave this for now, for a couple of reasons. Mostly I just don't think I'm going to have time to do this before tomorrow without displacing other work that needs to get finished. But also from the perspective of the HQ configuration for the user, if this is going to be a 3rd box below those for the "Value Field" and "Display Text Field", it seems confusing to make it so that the way in which the users specifies it is actually not the same. I think there are ways to resolve that if more thought was to go into the HQ UI, but maybe to be decided at a later date if it becomes needed.

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
ctsims: DataUtil is the one that has splitting, but it's by array (We should still try to use it/be consistent)

XPathJoinFunc has a join, didn't find another one, but will poke around a bit more.

amstone326:
@ctsims Feedback implemented

PR #595, opened 2017-06-22
Add XPath function checksum

shubham1g5:
Case url - https://manage.dimagi.com/default.asp?255290#1352684

Implements checksum by Verhoeff Algorithm. Code based on implementation here - https://en.wikibooks.org/wiki/Algorithm_Implementation/Checksums/Verhoeff_Algorithm#Java

### DIFF
--- a/src/main/java/org/commcare/cases/util/StringUtils.java
+++ b/src/main/java/org/commcare/cases/util/StringUtils.java
@@ -136,9 +136,9 @@ public class StringUtils {
         return cost[len0 - 1];
     }
 
-    /*
+    /**
      * Converts a string to a string array.
-	 */
+     */
     public static String[] toArray(String str) {
 
         String[] myArray = new String[str.length()];
--- a/src/main/java/org/commcare/cases/util/StringUtils.java
+++ b/src/main/java/org/commcare/cases/util/StringUtils.java
@@ -1,6 +1,7 @@
 package org.commcare.cases.util;
 
 import org.commcare.modern.util.Pair;
+import org.javarosa.core.util.ArrayUtilities;
 
 import java.text.Normalizer;
 import java.util.regex.Pattern;
@@ -45,9 +46,9 @@ public class StringUtils {
         //if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.GINGERBREAD) {
         normalized = Normalizer.normalize(input, Normalizer.Form.NFD);
         //} else {
-            //TODO: I doubt it's worth it, but in theory we could run
-            //some other normalization for the minority of pre-API9
-            //devices.
+        //TODO: I doubt it's worth it, but in theory we could run
+        //some other normalization for the minority of pre-API9
+        //devices.
         //}
 
         String output = diacritics.matcher(normalized).replaceAll("").toLowerCase();
@@ -133,5 +134,19 @@ public class StringUtils {
 
         // the distance is the cost for transforming all letters in both strings
         return cost[len0 - 1];
+    }
+
+    /*
+     * Converts a string to a string array.
+	 */
+    public static String[] toArray(String str) {
+
+        String[] myArray = new String[str.length()];
+
+        for (int i = 0; i < str.length(); i++) {
+            myArray[i] = str.substring(i, i + 1);
+        }
+
+        return myArray;
     }
 }
--- a/src/main/java/org/commcare/cases/util/StringUtils.java
+++ b/src/main/java/org/commcare/cases/util/StringUtils.java
@@ -4,6 +4,8 @@ import org.commcare.modern.util.Pair;
 import org.javarosa.core.util.ArrayUtilities;
 
 import java.text.Normalizer;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.regex.Pattern;
 
 /**
@@ -137,16 +139,16 @@ public class StringUtils {
     }
 
     /**
-     * Converts a string to a string array.
+     * Converts a string to a list of Characters.
      */
-    public static String[] toArray(String str) {
+    public static ArrayList<Character> toList(String str) {
 
-        String[] myArray = new String[str.length()];
+        ArrayList<Character> myArrayList = new ArrayList<>(str.length());
 
         for (int i = 0; i < str.length(); i++) {
-            myArray[i] = str.substring(i, i + 1);
+            myArrayList.add(i, str.charAt(i));
         }
 
-        return myArray;
+        return myArrayList;
     }
 }
--- a/src/main/java/org/javarosa/core/util/ArrayUtilities.java
+++ b/src/main/java/org/javarosa/core/util/ArrayUtilities.java
@@ -1,7 +1,5 @@
 package org.javarosa.core.util;
 
-import java.lang.reflect.Array;
-import java.util.Arrays;
 import java.util.Vector;
 
 /**
@@ -33,20 +31,6 @@ public class ArrayUtilities {
         }
         return true;
     }
-
-    /**
-     * Reverses an Array
-     */
-    public static <T> T[] reverse(Object[] myArray, Class<? extends T[]> tClass) {
-        Object[] reversed = new Object[myArray.length];
-
-        for (int i = 0; i < myArray.length; i++) {
-            reversed[i] = myArray[myArray.length - (i + 1)];
-        }
-
-        return Arrays.copyOf(reversed, reversed.length, tClass);
-    }
-
 
     /**
      * Find a single intersecting element common to two lists, or null if none
--- a/src/main/java/org/javarosa/core/util/ArrayUtilities.java
+++ b/src/main/java/org/javarosa/core/util/ArrayUtilities.java
@@ -34,9 +34,9 @@ public class ArrayUtilities {
         return true;
     }
 
-    /*
-    * Reverses an array
-    */
+    /**
+     * Reverses an Array
+     */
     public static <T> T[] reverse(Object[] myArray, Class<? extends T[]> tClass) {
         Object[] reversed = new Object[myArray.length];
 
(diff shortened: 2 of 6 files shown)


src/main/java/org/commcare/cases/util/StringUtils.java
```diff
@@ -134,4 +135,18 @@ private static int LevenshteinDistance(String s0, String s1) {
...
+
+    /**
+     * Converts a string to a string array.
+     */
+    public static String[] toArray(String str) {
+
+        String[] myArray = new String[str.length()];
+
+        for (int i = 0; i < str.length(); i++) {
+            myArray[i] = str.substring(i, i + 1);
```
wpride: This might be more succinct using the String `charAt()` function (and possibly this whole function should return a `char[]` rather than a `String[]`)

src/main/java/org/javarosa/core/util/ArrayUtilities.java
```diff
@@ -33,6 +35,20 @@ public static boolean arraysEqual(byte[] array1, byte[] array2) {
     }
 
     /**
+     * Reverses an Array
+     */
+    public static <T> T[] reverse(Object[] myArray, Class<? extends T[]> tClass) {
```
wpride: If you switch to using a `char[]` you could use Apache Commons `ArrayUtils.reverse` https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/ArrayUtils.html#reverse-char:A-

src/main/java/org/javarosa/xpath/expr/XPathChecksumFunc.java
```diff
@@ -0,0 +1,96 @@
...
+        name = NAME;
+        expectedArgCount = EXPECTED_ARG_COUNT;
+    }
+
+    public XPathChecksumFunc(XPathExpression[] args) throws XPathSyntaxException {
+        super(NAME, args, EXPECTED_ARG_COUNT, true);
+    }
+
+    @Override
+    public Object evalBody(DataInstance model, EvaluationContext evalContext, Object[] evaluatedArgs) {
```
wpride: can be static

src/main/java/org/javarosa/xpath/expr/XPathChecksumFunc.java
```diff
@@ -0,0 +1,96 @@
...
+        return checksum(evaluatedArgs[0], evaluatedArgs[1]);
+    }
+
+    /**
+     * @param o1 algorithm type used to calculate checksum. We only support 'verhoeff' for now.
+     * @param o2 input we are calculating checksum for
+     * @return checksum of {@code o2} calculated using  {@code o1} type algorithm
+     */
+    private String checksum(Object o1, Object o2) {
```
wpride: can be static

src/main/java/org/javarosa/xpath/expr/XPathChecksumFunc.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Calculates Verhoeff checksum for the given {@code input} string <p>
+     *
+     * @param input input string to calculate verhoeff checksum for
+     * @return Verhoeff checksum value for {@code input}
+     * @see <a href="https://en.wikibooks.org/wiki/Algorithm_Implementation/Checksums/Verhoeff_Algorithm#Java">Based on Verhoeff checksum implementation here</a>
+     */
+    private String verhoeffChecksum(String input) {
```
wpride: can be static

src/main/java/org/javarosa/xpath/expr/XPathChecksumFunc.java
```diff
@@ -0,0 +1,96 @@
+package org.javarosa.xpath.expr;
+
+import org.commcare.cases.util.StringUtils;
+import org.javarosa.core.model.condition.EvaluationContext;
+import org.javarosa.core.model.instance.DataInstance;
+import org.javarosa.core.util.ArrayUtilities;
+import org.javarosa.xpath.XPathUnsupportedException;
+import org.javarosa.xpath.parser.XPathSyntaxException;
+
+public class XPathChecksumFunc extends XPathFuncExpr {
```
wpride: Good looking class!

src/main/java/org/javarosa/core/util/ArrayUtilities.java
shubham1g5: @wpride It seems like currently we are not using 'commons-lang3' library in the project. Should I go ahead and add it.

src/main/java/org/javarosa/xpath/expr/XPathChecksumFunc.java
shubham1g5: 👍

src/main/java/org/commcare/cases/util/StringUtils.java
shubham1g5: 👍

src/main/java/org/javarosa/core/util/ArrayUtilities.java
wpride: Usually I prefer adding a dependency to re-implementing functionality but definitely your call

amstone326: Another option is if you start with splitting your `input` String to the `verhoeffChecksum` function into an `List` instead of an array, then you can use `Collections.reverse()`

src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
```diff
@@ -58,8 +58,8 @@ private void testEval(String expr, FormInstance model, EvaluationContext ec, Obj
             if (exceptionExpected) {
                 fail("Expected exception, expression : " + expr);
             } else if ((result instanceof Double && expected instanceof Double)) {
-                Double o = (Double)result;
-                Double t = (Double)expected;
+                Double o = (Double) result;
```
amstone326: So I cannot currently find the style guide of ours that I am pretty sure exists, but our practice with casting is actually to leave no space between the closing paren and the target of the cast. Sometimes Android Studio tries to change this for you automatically which is incredibly annoying, so I'm not even sure if you changed these explicitly, but would you mind switching back the lines in this file that were changed? (Also note that I am a big fan of all other linting changes you made in this PR 😄)

src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
```diff
@@ -523,6 +523,10 @@ public void doTests() {
         testEval("id-compress(0, 'CD','','ABCDE',1)", null, ec, new XPathException());
         testEval("id-compress(0, 'CD','CD','ABCDE',1)", null, ec, new XPathException());
 
+        testEval("checksum('verhoeff','41310785898')", null, null, "4");
```
amstone326: Can you remind me how you determined the correct values of these outputs prior to running them in the code here?

src/main/java/org/javarosa/xpath/expr/XPathChecksumFunc.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    private String checksum(Object o1, Object o2) {
+        String algorithmKey = FunctionUtils.toString(o1);
+        String input = FunctionUtils.toString(o2);
+
+        switch (algorithmKey) {
+            case ALGORITHM_KEY_VERHOEFF:
+                return verhoeffChecksum(input);
+            default:
+                throw new XPathUnsupportedException("Algorithm key " + algorithmKey + " is not supported for checksum");
```
shubham1g5: 👍

src/main/java/org/javarosa/xpath/expr/XPathChecksumFunc.java
```diff
@@ -0,0 +1,96 @@
...
+                {9, 4, 5, 3, 1, 2, 6, 8, 7, 0},
+                {4, 2, 8, 6, 5, 7, 3, 9, 0, 1},
+                {2, 7, 9, 3, 8, 0, 6, 4, 1, 5},
+                {7, 0, 4, 6, 9, 1, 3, 2, 5, 8}
+        };
+
+        // The inverse table
+        int[] inv = {0, 4, 3, 2, 1, 5, 6, 7, 8, 9};
+
+        int check = 0;
```
shubham1g5: 👍

src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
shubham1g5: Yeah. All the formatting changes are due to Android Studio auto format. I think we really need to have a standard style.xml that we can import in Android Studio and which Android Studio can then utilise to format/style the code to avoid these conflicts. Though for now will revert these changes.

src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
shubham1g5: I took some sample Aadhar ids and passed the first 11 digits to the checksum function while the last digit as the expected outcome. So for this case Aadhar Id I used would be  '413107858984".

src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
amstone326: Agreed. For now I found how to configure AS to stop adding a space after casts, so will show you that today

src/main/java/org/javarosa/core/util/ArrayUtilities.java
shubham1g5: 👍

shubham1g5:
Pushed all the suggested changes. @amstone326  Thanks for catching the error in FunctionUtils. Weird that it didn't fail on Jenkins build.

src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
amstone326: Ah perfect, I forgot that property of the Adhaar IDs is the whole point

PR #596, opened 2017-06-22
Memory Tuning for Optimizations

ctsims:
Fix for:

https://manage.dimagi.com/default.asp?255503
and
https://manage.dimagi.com/default.asp?255507

This effectively disables a lot of the optimizations for async factory restores, because the async cases end up loading quite differently in a way that tanks the memory.

Specifically this splits up the case group bulk fetch process to allow _reporting_ in an un-escalated context, but caching into only the bulk context, then allows the async process to prevent a carryforward of the escalated scope.

It also implements a process which allows platforms to define their own boundaries for the bulk fetches based on available memory. This should both prevent issues on small heap devices, and enable platforms like Touchforms to grow their heap

I tested the impact against a the UATBC list and confirmed that the downside is limited to needing to load the bulk query context twice (between the "search" and "Expand" phases of a case list) which is a small impact on performance.

cross-request: https://github.com/dimagi/commcare-android/pull/1746

### DIFF
--- a/src/main/java/org/commcare/modern/engine/cases/CaseGroupResultCache.java
+++ b/src/main/java/org/commcare/modern/engine/cases/CaseGroupResultCache.java
@@ -7,15 +7,21 @@ import java.util.HashMap;
 import java.util.LinkedHashSet;
 
 /**
+ * A case group result cache keeps track of different sets of "Bulk" cases which are
+ * likely to have data or operations tracked about them (IE: results of a common query which
+ * are likely to have further filtering applied.
+ *
+ * Since these results are often captured/reported before a context is escalated, this cache
+ * doesn't directly hold the resulting cached cases themselves. Rather a CaseObjectCache should
+ * be used to track the resulting cases. This will ensure that cache can be attached to the
+ * appropriate lifecycle
+ *
  * Created by ctsims on 1/25/2017.
  */
 
 public class CaseGroupResultCache implements QueryCache {
 
     private HashMap<String,LinkedHashSet<Integer>> bulkFetchBodies = new HashMap<>();
-
-    private HashMap<Integer, Case> cachedCases = new HashMap<>();
-
 
     public void reportBulkCaseBody(String key, LinkedHashSet<Integer> ids) {
         if(bulkFetchBodies.containsKey(key)) {
@@ -25,7 +31,7 @@ public class CaseGroupResultCache implements QueryCache {
     }
 
     public boolean hasMatchingCaseSet(int recordId) {
-        return isLoaded(recordId) || getTranche(recordId) != null;
+        return getTranche(recordId) != null;
     }
 
     public LinkedHashSet<Integer> getTranche(int recordId) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/commcare/modern/util/PerformanceTuningUtil.java
```diff
@@ -0,0 +1,66 @@
...
+
+        return identifyDefaultPrefetchBlockSize(maxMemory);
+    }
+
+    /**
+     * @return A heuristic for the largest safe value for block prefetch based on a provided
+     * amount of memory which should be available for optimizations.
+     */
+    public static int identifyDefaultPrefetchBlockSize(long availableMemoryInBytes) {
+        if (availableMemoryInBytes == 0 || availableMemoryInBytes == -1) {
```
amstone326: Is this just if the device doesn't support getting the size of memory? How did you decide what to return in this case?

src/main/java/org/commcare/modern/util/PerformanceTuningUtil.java
```diff
@@ -0,0 +1,66 @@
...
+        long maxMemory = rt.maxMemory();
+
+        return identifyDefaultPrefetchBlockSize(maxMemory);
+    }
+
+    /**
+     * @return A heuristic for the largest safe value for block prefetch based on a provided
+     * amount of memory which should be available for optimizations.
+     */
+    public static int identifyDefaultPrefetchBlockSize(long availableMemoryInBytes) {
```
amstone326: Can you add some indication of the rule you're using here to convert memory size to the max block size?

src/main/java/org/commcare/modern/engine/cases/CaseGroupResultCache.java
```diff
@@ -2,52 +2,46 @@
...
     public boolean hasMatchingCaseSet(int recordId) {
-        return isLoaded(recordId) || getTranche(recordId) != null;
+        return getTranche(recordId) != null;
     }
 
-    public LinkedHashSet<Integer> getTranche(int recordId) {
-        for(LinkedHashSet<Integer> tranche: bulkFetchBodies.values()) {
-            if(tranche.contains(recordId)){
-                return tranche;
+    public Pair<String, LinkedHashSet<Integer>> getTranche(int recordId) {
```
amstone326: I know this existed in the prior PR, but can you rename this method to something more meaningful? It's really hard to understand what this is doing

amstone326:
This also isn't directly related to these changes, but does the `contextScope` field in `QueryContext` represent the size of the case load that the current query is concerned with? If so, can you add a comment to that effect explicitly? The current comment on that field doesn't say this and it took me a while to trace what it actually represents

src/main/java/org/commcare/modern/util/PerformanceTuningUtil.java
ctsims: These are just common flags that I wanted to not change behavior for. 7500 was the previous default, so I kept it as the default in any situations where we didn't have a clear heuristic.

ctsims:
@amstone326 All PR Feedback Integrated

ctsims:
Pulling since PR feedback was integrated.

PR #597, opened 2017-06-23
Support for "%w" date format for day of week

kaapstorm:
Hey @amstone326, any chance of adding date format support for "%w" for day-of-week calculations? (Also, I don't have a mobile dev environment set up, so this code is untested. Please don't merge before that unit test runs!)

If you do decide to accept this, and you let me know what release it'll be in, I'll update the "CommCare Functions" Confluence page.


kaapstorm:
Thanks @amstone326. This is for calculations for appointments or reports that relate to a specific day of the week. e.g. "MW does visits on the first Monday of the month, how many days away is that?" It's possible to use the string value a "cond" clause, but that's not ideal.

PR #599, opened 2017-06-23
Formplayer

benrudolph:
@wpride should we periodically be doing this?


wpride:
@benrudolph This has been an ongoing goal for me but the goalposts keep moving. I just did a big pass merging a bunch of stuff a few weeks ago. The remaining differences are harder to rectify - for example doing a real implementation of case claim for the CLI - or just something that Android doesn't need - for example the `instance()` thread safe wrapper is used for three or four singletons but Android obviously never has to worry about that.

benrudolph:
that's a bummer 😞 this is going to be hard to maintain

> for example doing a real implementation of case claim for the CLI

can't the CLI just not have this functionality?

> or just something that Android doesn't need - for example the instance() thread safe wrapper is used for three or four singletons but Android obviously never has to worry about that.

for your specific example I think it's ok to merge code that makes commcare thread safe even though we know on the phone it'll only ever be accessed by one user at a time. it's just a simpler manifestation. in general, i'd hope we can keep stuff that android really doesn't need in the formplayer directory, but realize that's easier said then done

PR #603, opened 2017-06-27
Enable entity traces for menus

amstone326:
cross-request: https://github.com/dimagi/commcare-android/pull/1750


src/main/java/org/javarosa/core/model/trace/EvaluationTraceReporter.java
```diff
@@ -9,12 +9,28 @@
...
 
-    void reset();
+    public abstract void reportTrace(EvaluationTrace trace);
 
-    Vector<EvaluationTrace> getCollectedTraces();
+    public abstract void reset();
+
+    public abstract Vector<EvaluationTrace> getCollectedTraces();
+
+    public void printAndClearTraces(String description) {
```
ctsims: I think moving this functionality to this interface is a bit of a negative. It couples this class strongly with the stringevaluationtracreserializer and moves an interface to an abstract class (which introduces limitations on the upstream classes). The decision of how to output your traces is also not something that the implemented classes would want to change  based on their implementation.

As a practical reason: Will has mentioned that dumping stuff to sysout isn't helpful for formplayer, so he's likely to need to handle the outputs on formplayer differently. Moving this here will make him have to find a way to decouple the output at a class level rather than at the method call level.

Is there a specific reason for moving this here instead of a util method? One other huge advantage of the util method approach is that it lets us make internal decisions based on the type of trace reporter. The null check for the reporter now has to be repeated in a helper method, for instance.

amstone326: I don't think I'm following these reasons. Questions below:

1. "Moving this here will make him have to find a way to decouple the output at a class level rather than at the method call level." --> Doesn't having this be a method of an abstract class make it easier for Will to override its behavior, rather than harder? 

2. "One other huge advantage of the util method approach is that it lets us make internal decisions based on the type of trace reporter." --> The current approach makes it very straightforward to do that; subclasses would just have to override the base implementation of `printAndClearTraces()`

ctsims: 1. Will needs to overwrite the behavior as the result of the code being run on a different platform, and having to superclass the Reporter (whose behavior differs in how it collects and aggregates the incoming EvaluationTraces) based on the platform would "fork" the reason for the subclass. 

As an example: One of the current trace reporters aggregates inputs and reduces them to time profiles, the other reduces them to debugging traces. If Will wanted both of them to work on FormPlayer, he'd actually have to subclass both of those, and independently have each subclass re-implement this method to use the same behavior.

2. I was referring to this behavior:
https://github.com/dimagi/commcare-core/pull/603/files#diff-ea18a5fcfa3cd726df3085e864d41c2bR113

https://github.com/dimagi/commcare-core/pull/603/files#diff-177f6db6a65942ac830e218ab8fc34f1R111

Which is common and would be re-implemented each time hte profile is used. This call could be put into a static utility method like
https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/utils/InstrumentationUtils.java#L19

but only if the reporter is an argument to the method, it can't be done with polymorphism because the reporter itself is the thing that's null.

amstone326:
@ctsims Ok cool sorry for misunderstanding, am on board with all that now. Updated.

PR #604, opened 2017-06-28
XPath Parser Error Message Fix

ctsims:
Fixes an issue where an error message would always display "null" as part of its message, rather than a useful piece of context.

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -1614,9 +1614,10 @@ public class XFormParser {
             tref = TreeReference.selfRef(); //only happens for <group>s with no binding
         }
 
+        TreeReference refPreContextualization = tref;
         tref = tref.parent(parentRef);
         if (tref == null) {
-            throw new XFormParseException("Binding path [" + ref.getReference().toString(true) + "] not allowed with parent binding of [" + parentRef + "]");
+            throw new XFormParseException("Binding path [" + refPreContextualization.toString(true) + "] not allowed with parent binding of [" + parentRef + "]");
         }
 
         return new XPathReference(tref);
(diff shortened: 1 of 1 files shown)


src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1616,7 +1616,7 @@ public static XPathReference getAbsRef(XPathReference ref, TreeReference parentR
 
         tref = tref.parent(parentRef);
         if (tref == null) {
-            throw new XFormParseException("Binding path [" + tref + "] not allowed with parent binding of [" + parentRef + "]");
+            throw new XFormParseException("Binding path [" + ref.getReference().toString(true) + "] not allowed with parent binding of [" + parentRef + "]");
```
wpride: Looks from the conditional block above like `ref` can be null - should we be handling that here?

ctsims: Oh, I was reading that output wrong, good catch. Will address.

ctsims:
@wpride thanks for that, addressed.

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1614,9 +1614,10 @@ public static XPathReference getAbsRef(XPathReference ref, TreeReference parentR
             tref = TreeReference.selfRef(); //only happens for <group>s with no binding
         }
 
+        TreeReference refPreContextualization = tref;
```
amstone326: @ctsims Assuming I'm missing something, but doesn't this just make `refPreContextualization` have a pointer to `tref`? How will this value remain different?

ctsims: tref gets redefined below which is what ends up making its value null if the contextualization fails. This just retains it's "pre-contextualized" state, which is generally a relative reference  like ("../something") which didn't work

PR #605, opened 2017-07-03
Make module badge calculations asynchronous

shubham1g5:
Case url: https://manage.dimagi.com/default.asp?257029#1355558

-  Adds RxJava2 to the project.
 - Makes MenuDisplayable getTextForBadge() return rx Observable to allow for async calculation
- getTextForBadge() implementation uses Schedulers.computation() to offload the work from main thread.

### DIFF
--- a/src/main/java/org/commcare/suite/model/Entry.java
+++ b/src/main/java/org/commcare/suite/model/Entry.java
@@ -20,6 +20,9 @@ import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.Vector;
 
+import io.reactivex.Observable;
+import io.reactivex.schedulers.Schedulers;
+
 /**
  * Describes a user-initiated action, what information needs to be collected
  * before that action can begin, and what the UI should present to the user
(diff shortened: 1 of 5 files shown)


src/main/java/org/commcare/suite/model/Entry.java
```diff
@@ -140,11 +143,12 @@ public String getDisplayText() {
...
     @Override
-    public String getTextForBadge(EvaluationContext ec) {
+    public Observable<String> getTextForBadge(EvaluationContext ec) {
         if (display.getBadgeText() == null) {
-            return null;
+            return Observable.just("");
         }
-        return display.getBadgeText().evaluate(ec);
+        return Observable.just(display.getBadgeText().evaluate(ec))
+                .subscribeOn(Schedulers.computation());
```
amstone326: Why doesn't `Menu.java`'s implementation of `getTextForBadge()` also include this call to `.subscribeOn(Schedulers.computation())`?

shubham1g5: Great catch. It shouldn't be here as well as I am letting the caller decide(Android code) which Scheduler to use. Will change it.

PR #606, opened 2017-07-07
Set authority for profile install based on install mode

amstone326:


src/cli/java/org/commcare/util/engine/CommCareConfigEngine.java
```diff
@@ -214,9 +214,11 @@ private void init(String profileRef) {
         }
     }
 
-    public void installAppFromReference(String profileReference) throws UnresolvedResourceException,
+    public void installAppFromReference(String profileReference)
+            throws UnresolvedResourceException,
             UnfullfilledRequirementsException, InstallCancelledException {
-        ResourceManager.installAppResources(platform, profileReference, this.table, true);
+        ResourceManager.installAppResources(platform, profileReference, this.table, true,
+                Resource.RESOURCE_AUTHORITY_LOCAL);
```
wpride: Was surprised this was safe but appears to be - I guess the CLI never supported installing from the profile reference?

amstone326: Yea I initially thought that this was going to be more complicated too, but I traced the code path and it seemed like it's only ever offline install

ctsims: Yeah, currently the CLI only supports 3 modes:
1) Link to an online CCZ file that gets downloaded and unzipped locally
2) Link to a local CCZ file that gets the same treatment
3) Link to a _local_ profile file 

The CLI _should_ support online _update, but installs I don't think ever rely on a ccpr file from the network.

PR #607, opened 2017-07-07
Performance: Allow bulk lookups with indexed fixtures

ctsims:
Move bulk read optimizations into the core engine and uses the same pattern that was developed for "bulk reads" of case data with fixture data as well. Major improvements to lookups which need to operate on indexed fixtures with more than 100 elements

cross-request: https://github.com/dimagi/commcare-android/pull/1759


ctsims:
Release Note:
 * Performance improvements for filters and lookup tables using large numbers of locations.

wpride:
@ctsims awesome to see so much moved into core. Is that with an eye towards moving this into Formplayer?

src/main/java/org/commcare/modern/engine/cases/RecordObjectCache.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A straightforward cache object query cache. Stores objects by their record ID.
+ *
+ * Used by other optimizations to isolate doing bulk loads and ensure that they are relevant
+ * when they occur
+ *
+ * Created by ctsims on 6/22/2017.
+ */
+
+public class RecordObjectCache<T> implements QueryCache {
```
wpride: I'm a bit confused by this class's purpose. I don't see it ever being instantiated and I don't see where the `cachedRecordObjects` are ever written. Could you clarify?

src/main/java/org/javarosa/core/model/utils/InstrumentationUtils.java
```diff
@@ -0,0 +1,32 @@
...
+     */
+    public static void printAndClearTraces(EvaluationTraceReporter reporter, String description) {
+        if (reporter.wereTracesReported()) {
+            System.out.println(description);
+        }
+
+        StringEvaluationTraceSerializer serializer = new StringEvaluationTraceSerializer();
+
+        for (EvaluationTrace trace : reporter.getCollectedTraces()) {
+            System.out.println(trace.getExpression() + ": " + trace.getValue());
```
wpride: Might be nice to associate the print stream with the `EvaluationTraceReporter` so that we don't have Sys.outs all over the place (wouldn't want this logging to the system in Formplayer, for example)

wpride:
@ctsims Just a couple of questions and the conflict then GTM from me

src/main/java/org/commcare/modern/engine/cases/RecordObjectCache.java
ctsims: They get written the bulk storage layer when we do a bulk read

Basically this is a cache lifecycle object where we store something like 

Case ID 2 -> [fully loaded case object from storage]

[Note: This is just the cacheobjectcasche from before, just namechanged to be more generic]

We use the other cache (the record set cache) to keep around lists of cases which should be bulk loaded together, but that list persists across diferent events, we separate them so this cache can be cleared at the end of various queries.

src/main/java/org/javarosa/core/model/utils/InstrumentationUtils.java
ctsims: This will be a bit tricky, but we can coordinate on how to get that set up in a useful way.

ctsims:
@wpride 

The specific reason is mostly to merge the core optimizations so they can be striped across the common storage elements, but it should also make it so that our tests and formplayer will be able to use the core optimization schemes soon, which should make it way easier for us to have consistent behavior and confidence across platforms

PR #613, opened 2017-07-19
Date Formatting w/Other Calendar Schemes

ctsims:
Allows using date format strings in the `format-date-for-calendar` the same as they are used in the `format-date` function

The month behavior still depends on platform, but makes it possible to use alternative calendars in a more predictable way around the numeric components.

Needed by L10k

IE:

`format-date-for-calendar('2017-07-15', 'ethiopian', '%Y-%m-%d')` -> `2009-11-08`

Old behaviors for the function calls should be unaffected

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathFormatDateForCalendarFunc.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathFormatDateForCalendarFunc.java
@@ -3,6 +3,7 @@ package org.javarosa.xpath.expr;
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.xform.util.CalendarUtils;
+import org.javarosa.xpath.XPathArityException;
 import org.javarosa.xpath.XPathUnsupportedException;
 import org.javarosa.xpath.parser.XPathSyntaxException;
 
@@ -10,7 +11,7 @@ import java.util.Date;
 
 public class XPathFormatDateForCalendarFunc extends XPathFuncExpr {
     public static final String NAME = "format-date-for-calendar";
-    private static final int EXPECTED_ARG_COUNT = 2;
+    private static final int EXPECTED_ARG_COUNT = -1;
 
     public XPathFormatDateForCalendarFunc() {
         name = NAME;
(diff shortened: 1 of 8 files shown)


ctsims:
Release Note:
Extended 'format-date-for-calendar' function to translate the output form using format-date input strings.

src/test/java/org/javarosa/core/model/utils/test/LocalizerTest.java
```diff
@@ -680,10 +680,15 @@ public void run() {
     private void runAsync(Runnable test, String label) {
         Thread t = new Thread(test);
         t.start();
-        try {
-            t.join(50);
-        } catch (InterruptedException e) {
+        int attempts = 4;
```
ctsims: This just makes this test get a bit longer to run. Weird things were causing it to not finish in the 50ms for me on occasion (I think garbage collection on the second thread)

src/main/java/org/javarosa/xpath/expr/XPathFormatDateForCalendarFunc.java
```diff
@@ -22,29 +23,41 @@ public XPathFormatDateForCalendarFunc(XPathExpression[] args) throws XPathSyntax
...
+    protected void validateArgCount() throws XPathSyntaxException {
+        if (args.length < 2 || args.length > 3) {
+            throw new XPathArityException(name, "2 or 3 arguments", args.length);
+        }
+    }
+
+    @Override
     public Object evalBody(DataInstance model, EvaluationContext evalContext, Object[] evaluatedArgs) {
-        return formatDateForCalendar(evaluatedArgs[0], evaluatedArgs[1]);
+        String formatString = null;
```
wpride: Would it make sense to set this to `"%e %B %Y"` here (since it'll end up being that anyways if its not overwritten on the next line) or is that too cute?

wpride:
@ctsims thanks for doing this. one minor comment then gtm from me

src/main/java/org/javarosa/xpath/expr/XPathFormatDateForCalendarFunc.java
ctsims: I could see the value there. In theory if that was a bad 'default' format for some future calendar, that would prevent that calendar from being able to differentiate between a specific format request and the default without modifying this core method, though, so I think this is a slightly safer fallback.

amstone326:
announce: Allows using date format strings in the `format-date-for-calendar` function, in the same as they are used in the `format-date` function

PR #614, opened 2017-07-20
Fixes max calculation for negative numbers

shubham1g5:
Case: https://manage.dimagi.com/default.asp?258478#1366378


shubham1g5:
Not

PR #616, opened 2017-07-20
Merge Master into Formplayer

ctsims:
Fixes max() and adds format for calendar


shubham1g5:
@wpride What do we use this formplayer branch for ?

wpride:
@shubham1g5 this is the branch of `commcare-core` that Formplayer will always use. There are some core divergences between how Formplayer and Android use the library that we haven't resolved yet so Formplayer isn't on master. We generally merge in the code from master after each release or as necessary.

PR #617, opened 2017-07-24
Fix Duplication of exception source in NodeEntityFactory

shubham1g5:
Case url: http://manage.dimagi.com/default.asp?258799

In the final error message to user the exception source was coming twice from the outer and nested exception. There is no need to set the source the outer exception as the inner exception already has that set.

### DIFF
--- a/src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
+++ b/src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
@@ -67,7 +67,7 @@ public class NodeEntityFactory {
                     sortData[count] = sortText.evaluate(nodeContext);
                 }
                 relevancyData[count] = f.isRelevant(nodeContext);
-            } catch (XPathSyntaxException | XPathException e) {
+            } catch (XPathSyntaxException e) {
                 /**
                  * TODO: 25/06/17 remove catch blocks from here
                  * We are wrapping the original exception in a new XPathException to avoid
--- a/src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
+++ b/src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
@@ -73,10 +73,7 @@ public class NodeEntityFactory {
                  * We are wrapping the original exception in a new XPathException to avoid
                  * refactoring large number of functions caused by throwing XPathSyntaxException here.
                  */
-                XPathException xe = new XPathException(e.getMessage(), e);
-                if (e instanceof XPathException) {
-                    xe.setSource(((XPathException)e).getSource());
-                }
+                XPathException xe = new XPathException(e);
                 throw xe;
             }
             count++;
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
```diff
@@ -73,10 +73,7 @@ public Detail getDetail() {
                  * We are wrapping the original exception in a new XPathException to avoid
                  * refactoring large number of functions caused by throwing XPathSyntaxException here.
                  */
-                XPathException xe = new XPathException(e.getMessage(), e);
-                if (e instanceof XPathException) {
-                    xe.setSource(((XPathException)e).getSource());
-                }
+                XPathException xe = new XPathException(e);
```
amstone326: For the same reason as that of the change you made in the android part of this PR, don't you only want to create this wrapped exception if the original exception was an `XPathSyntaxException`? Otherwise you have the same issue of wrapping an `XPathSyntaxException` by another `XPathSyntaxException`

src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
```diff
@@ -73,10 +73,7 @@ public Detail getDetail() {
                  * We are wrapping the original exception in a new XPathException to avoid
                  * refactoring large number of functions caused by throwing XPathSyntaxException here.
                  */
-                XPathException xe = new XPathException(e.getMessage(), e);
```
amstone326: What will be the value of `xe.getMessage()` now that you are not explicitly passing a message to the constructor? We would want `xe.getMEssage()` to be equivalent to `e.getMessage()`; can you confirm that that is the case?

shubham1g5: We don't need to pass message explicitly. I have confirmed that xe.getMessage() is indeed qual to e.getMessage().

src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
shubham1g5: You are right that we only actually want to wrap in case of an XPathSyntaxException though for the sake of consistency I have wrapped both of them. Since we are now not passing message explicitly the new XPathExcpetion will in both cases give the same message as the original exception (Earlier it was duplicating for XPathException). If it's confusing we can throw the XPathException as it is and only wrap in case of a XPathSyntaxException.

amstone326: Gotcha. I think I would still prefer if we only wrapped the `XPathSyntaxException`, since the only reason we are wrapping at all is to avoid the messy-ness of throwing a non-runtime exception. By wrapping both types of exceptions, it obfuscates our real reason for doing the wrapping, which I don't like.

shubham1g5: Yeah. Now I think so as well. Made the change

PR #619, opened 2017-07-25
Implements ModernHTTPRequester using Retrofit

shubham1g5:
Case url: https://manage.dimagi.com/default.asp?257212

- Creates `CommcareNetworkService` as a Retrofit interface for all requests.
- `CommcareNetworkServiceGenerator` creates an instance of `CommcareNetworkService` using `okHTTP` client and `AuthenticationInterceptor`
- Mock Responses for testing are handled using `OkHTTPResponseMock` and `FakeResponseBody`
- Also Moves Log Types from AndroidLogger to a separate class `LogTypes` to allow for using them at commcare-core level.


src/main/java/org/commcare/core/network/OkHTTPResponseMock.java
```diff
@@ -0,0 +1,56 @@
...
+import okhttp3.Headers;
+import okhttp3.Protocol;
+import okhttp3.Request;
+import okhttp3.ResponseBody;
+import retrofit2.Response;
+
+/**
+ * Response Factory for OkHTTP Response
+ */
+public class OkHTTPResponseMock {
```
amstone326: Can you add `Generator` or `Factory` to this class name?

src/main/java/org/commcare/core/network/OkHTTPResponseMock.java
```diff
@@ -0,0 +1,56 @@
...
+    public static Response<ResponseBody> createResponse(Integer responseCode, Headers headers) {
+        return createResponse(responseCode, null, headers);
+    }
+
+    public static Response<ResponseBody> createResponse(Integer responseCode, InputStream inputStream) {
+        ResponseBody responseBody = new FakeResponseBody(inputStream);
+        return createResponse(responseCode, responseBody);
+    }
+
+    public static Response<ResponseBody> createResponse(Integer responseCode, String body) {
```
amstone326: The last 3 methods in this class can be `private`

src/main/java/org/commcare/core/network/OkHTTPResponseMock.java
```diff
@@ -0,0 +1,56 @@
...
+    public static Response<ResponseBody> createResponse(Integer responseCode, ResponseBody responseBody) {
+        return createResponse(responseCode, responseBody, null);
+    }
+
+    public static Response<ResponseBody> createResponse(Integer responseCode, ResponseBody responseBody, Headers headers) {
+
+        okhttp3.Response.Builder responseBuilder = new okhttp3.Response.Builder()
+                .code(responseCode)
+                .protocol(Protocol.HTTP_1_1)
+                .request(new Request.Builder().url("http://localhost/").build());
```
amstone326: Why does this actually make a request, even to localhost?

src/main/java/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -89,8 +92,12 @@ public OutputIOException(IOException internal) {
         return output.toByteArray();
     }
 
+    public static InputStream toInputStream(String s) {
```
amstone326: This looks like it's unused

src/main/java/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -1,9 +1,12 @@
 package org.javarosa.core.io;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
```
amstone326: unused import

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+                if (!isValidRedirect(request.url(), HttpUrl.parse(newUrl))) {
+                    Logger.log(LogTypes.TYPE_WARNING_NETWORK, "Invalid redirect from " + request.url().toString() + " to " + response.request().url().toString());
+                    throw new IOException("Invalid redirect from secure server to insecure server");
+                }
+            }
+            return response;
+        }
+    };
+
+    private static HttpLoggingInterceptor logging = new HttpLoggingInterceptor()
```
shubham1g5: Currently to adb only. But we can remove this since we are explicitly logging requests anyway.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+            .setLevel(HttpLoggingInterceptor.Level.BASIC);
+
+    private static OkHttpClient.Builder httpClient = new OkHttpClient.Builder()
+            .connectTimeout(ModernHttpRequester.CONNECTION_TIMEOUT, TimeUnit.MILLISECONDS)
+            .readTimeout(ModernHttpRequester.CONNECTION_SO_TIMEOUT, TimeUnit.MILLISECONDS)
+            .addNetworkInterceptor(redirectionInterceptor)
+            .addInterceptor(logging)
+            .followRedirects(true);
+
+    private static Retrofit retrofit = builder.build();
```
shubham1g5: So we need to retain to the Retrofit instance as static variable so that all the requests uses the same socket connection. Though I am going to change the implementation so that we don't re-instantiate it again.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+
+
+    public static CommCareNetworkService createCommCareNetworkService(final String credential) {
+        if (credential != null) {
```
shubham1g5: Yeah. You are right. The better approach would be to always set the Auth Interceptor and update the credentials there. Going to change this.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+            AuthenticationInterceptor authInterceptor =
+                    new AuthenticationInterceptor(credential);
+            if (!httpClient.interceptors().contains(authInterceptor)) {
```
shubham1g5: Good catch. Will remove all of this messed up code.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -41,149 +41,93 @@
...
+    /**
+     * Executes the HTTP Request. Can be called directly to bypass response processor.
+     *
+     * @return Response from the HTTP call
+     * @throws IOException
+     */
+    public Response<ResponseBody> makeRequest() throws IOException {
+        switch (method) {
+            case POST:
+                currentCall = commCareNetworkService.makePostRequest(url, params, addPostHeaders(requestBody), requestBody);
```
shubham1g5: I was just trying to retain the earlier behaviour although current post headers are adding content Type and Content Length for the `POST` body and for a `MULTIPART_POST` this info would be different for different parts of the request. I think we can rename `addPostHeaders` as `addContentHeaders' to avoid the confusion.

src/main/java/org/commcare/core/network/OkHTTPResponseMock.java
shubham1g5: It's just setting the request parameters on the response object we are building and not making an actual request. It's a required parameter for `okhttp response`

src/main/java/org/commcare/core/network/ModernHttpRequester.java
amstone326: Ok, that sounds good to me

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
amstone326: Ok cool yea, I don't think logging network stuff to adb adds a whole lot for us.

src/main/java/org/commcare/core/network/AuthenticationInterceptor.java
```diff
@@ -0,0 +1,52 @@
...
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class AuthenticationInterceptor implements Interceptor {
+
+    @Nullable
+    private String credential;
+
+    public AuthenticationInterceptor(@Nullable String credential) {
```
amstone326: This constructor is now only ever called with the parameter as `null`, so can we just make it an empty constructor?

src/main/java/org/commcare/core/network/AuthenticationInterceptor.java
```diff
@@ -0,0 +1,52 @@
...
+
+        Request.Builder builder = original.newBuilder()
+                .header("Authorization", credential);
+
+        Request request = builder.build();
+        return chain.proceed(request);
+    }
+
+    @Nullable
+    public String getCredential() {
```
amstone326: This is unused

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -0,0 +1,83 @@
...
+    private static HttpLoggingInterceptor logging = new HttpLoggingInterceptor()
+            .setLevel(HttpLoggingInterceptor.Level.BASIC);
+
+    private static AuthenticationInterceptor authenticationInterceptor = new AuthenticationInterceptor(null);
+
+    private static OkHttpClient.Builder httpClient = new OkHttpClient.Builder()
+            .connectTimeout(ModernHttpRequester.CONNECTION_TIMEOUT, TimeUnit.MILLISECONDS)
+            .readTimeout(ModernHttpRequester.CONNECTION_SO_TIMEOUT, TimeUnit.MILLISECONDS)
+            .addNetworkInterceptor(redirectionInterceptor)
+            .addInterceptor(authenticationInterceptor)
```
amstone326: what's the conceptual difference between `addNetworkInterceptor()` and `addInterceptor()`?

src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -208,31 +150,71 @@ public static void processResponse(HttpResponseProcessor responseProcessor,
...
+    public static RequestBody getPostBody(HashMap<String, String> inputs) {
+        FormBody.Builder formBodyBuilder = new FormBody.Builder();
+        for (Map.Entry<String, String> param : inputs.entrySet()) {
+            formBodyBuilder.add(param.getKey(), param.getValue());
+        }
+        return formBodyBuilder.build();
+    }
 
-        cache.initializeCache();
+    public static String getCredential(User user, String domain) {
```
amstone326: I think these got deleted from this class, but not yet added to `HttpUtils`, so there are compilation issues now

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: `NetworkInterceptors` gets called for every network request i.e. even for redirected requests   while redirection is encapsulated for general request interceptors.

src/main/java/org/commcare/core/network/AuthenticationInterceptor.java
```diff
@@ -0,0 +1,43 @@
...
+    public Response intercept(Chain chain) throws IOException {
+        Request original = chain.request();
+
+        // if credentials are null, proceed without authenticating
+        if (credential == null) {
+            return chain.proceed(original);
+        }
+
+        // Throw an exception if we are sending an authenticated request over HTTP
+        if (!original.isHttps() && credential != null) {
```
ctsims: Noting that this would actually be a change to our current behaviors. I think it's likely a good change, but we will need to work to understand if we're introducing limitations that affect meaningful behavior before pushing this.

One example: Internally people do deploys from their local CommCareHQ instance to test functionality, and their machine's don't have HTTPS turned on, so this would prevent that from working

src/main/java/org/commcare/core/network/AuthenticationInterceptor.java
```diff
@@ -0,0 +1,43 @@
...
+
+        Request request = builder.build();
+        return chain.proceed(request);
+    }
+
+    public void setCredential(@Nullable String credential) {
+        this.credential = credential;
+    }
+
+    public static class PlainTextPasswordException extends RuntimeException {
```
ctsims: Seems a bit odd for this to be a RuntimeException. I'd imagine that this should be an IOException base?

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Provides an instance of CommCareNetworkService.
+ * We have declared everything static in this class as we want to use the same objects (OkHttpClient, Retrofit, …) throughout the app
+ * to just open one socket connection that handles all the request and responses.
+ */
+
+public class CommCareNetworkServiceGenerator {
+
+    // Retrofit needs a base url to generate an instance but since our apis are fully dynamic it's not getting used.
+    private static final String BASE_URL = "https://www.commcarehq.org/";
```
ctsims: I'm not sure I understand that line, can you explain what you mean by that?

CommCare mobile is in no way tied to our server, so in general I'd push back on any part of the code containing explicit HTTP references to our current production environment

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+            .followRedirects(true);
+
+    private static Retrofit retrofit = builder.client(httpClient.build()).build();
+
+    public static CommCareNetworkService createCommCareNetworkService(final String credential) {
+        authenticationInterceptor.setCredential(credential);
+        return retrofit.create(CommCareNetworkService.class);
+    }
+
+    private static boolean isValidRedirect(HttpUrl url, HttpUrl newUrl) {
```
ctsims: Quick Note: We only implemented this by hand before because the Raw Java HTTP libraries didn't do it for us. If Retrofit will replicate the same behavior we don't need to keep this implemented internally.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -41,149 +40,93 @@
...
-        writer.flush();
-        writer.close();
-        os.close();
-    }
 
-    protected URL buildUrlWithParams() throws MalformedURLException {
-        UriBuilder b = UriBuilder.fromUri(url.toString());
-        for (Map.Entry<String, String> param : params.entrySet()) {
-            b.queryParam(param.getKey(), param.getValue());
+    private Map<String, String> addContentHeaders(RequestBody postBody) {
```
ctsims: This method name seems like it's confusingly generic. It's "addContentHeadersForXFormPost", right? Not a generic thing

src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -41,149 +40,93 @@
...
+        headers.put("Content-Type", "application/x-www-form-urlencoded");
+        try {
+            headers.put("Content-Length", postBody.contentLength() + "");
+        } catch (IOException e) {
+            // Do we care ?
+            e.printStackTrace();
```
ctsims: I'd be somewhat surprised if this wasn't called in a context where adding the IOException to the method signature wasn't better than swallowing this exception.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -41,149 +40,93 @@
...
-    }
 
-    protected URL buildUrlWithParams() throws MalformedURLException {
-        UriBuilder b = UriBuilder.fromUri(url.toString());
-        for (Map.Entry<String, String> param : params.entrySet()) {
-            b.queryParam(param.getKey(), param.getValue());
+    private Map<String, String> addContentHeaders(RequestBody postBody) {
+        headers.put("Content-Type", "application/x-www-form-urlencoded");
+        try {
+            headers.put("Content-Length", postBody.contentLength() + "");
```
ctsims: As a small point of style I'd replace this with "String.valueOf(postBody.contentLength())" to eliminate ambiguity about type coercion

src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -193,12 +136,10 @@ public static void processResponse(HttpResponseProcessor responseProcessor,
             try {
                 responseStream = streamAccessor.getResponseStream();
             } catch (IOException e) {
-                responseProcessor.handleIOException(e);
+                responseProcessor.handleException(e);
```
ctsims: Is there a specific reason why we eliminated the specific exception typing on the response processor?

I think in the abstract sense, blurring an IOException to an untyped Exception seems like a bad thing.

src/main/java/org/commcare/core/network/AuthenticationInterceptor.java
shubham1g5: So  our earlier `ModernHTTPRequester` was enforcing this. So I retained this in the merger http requestor. By making it controllable do you mean providing some sort of flag in developer options or remove it completely ?

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: I meant the same thing when quoting `since our apis are fully dynamic it's not getting used`.  If you see the Retrofit interface [here](https://github.com/dimagi/commcare-core/blob/retrofit/src/main/java/org/commcare/core/network/CommCareNetworkService.java) we are always passing the complete urls when making a request. So base_url doesn't end up getting used and acts just as a placeholder for building Retrofit instance.(Retofit needs a non null non empty properly formatted base url for creating an instance)

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: OkHTTP builder only allows us to set `followRedirects(boolean)` which allows for all redirects  and `followSslRedirects(boolean)` which can restrict HTTP -> HTTPS and HTTPS -> HTTP. So there is still no way to set only HTTP -> HTTPS and restrict HTTPS -> HTTP

src/main/java/org/commcare/core/network/ModernHttpRequester.java
shubham1g5: Yeah. Also the only Post request we make is from `PostRequestActivity` and I was not sure of use case for that. Since it's causing confusion I think it makes sense to move this method to `PostRequestActivity` itself and rename it to `addContentHeadersForXFormPost`

src/main/java/org/commcare/core/network/ModernHttpRequester.java
shubham1g5: Yeah. It's a bad JAVA habit I picked up from JS

src/main/java/org/commcare/core/network/ModernHttpRequester.java
shubham1g5: Again something I am retaining from earlier `ModernHTTPRequester` though I think it's due to the fact that not sending content-length as a header is not a deal breaker for a request to succeed.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
shubham1g5: Agree. Though it was done to incorporate `PlainTextPasswordException`. If we make it an `IOException` then I think we would be able to retain the earlier exception typing

src/main/java/org/commcare/core/network/AuthenticationInterceptor.java
shubham1g5: Hmm. May be it makes sense here because our request urls are dynamic and dependent on the ccz profile rather than the app structure. Though I just copied it from earlier implementation and is not sure of the original intent.

src/main/java/org/commcare/core/network/AuthenticationInterceptor.java
ctsims: Oh, funny! Sorry, I hadn't actually realized we moved over to doing that. If this replicates existing behavior, you can ignore that request.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
ctsims: If that's the case I think we should replace the `https://www.commcarehq.org/` reference with a reference to something which would never work, like `http://example.url`. If someone misuses or misconfigures the libraries with this base URL I'm worried about the possibility that they write code that *accidentally* works because the url was `www.commcarehq.org` as is their test data, since that will introduce incredibly hard to understand errors when someone moves to a non-production URL and the code is failing there.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
ctsims: Good to know.

Out of curiosity, does it specifically strike you as an odd decision for us to restrict https->http only? A lot of this code was written in a time when we had to re-implement lots of "core" functionality for most systems (like digest auth) by hand, so there are some spots where we wrote this that are quite specific which may be less important than standardizing on normal industry behavior now that we are in a world with standard libraries again.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
ctsims: I think either would be good, would just want it to be clear from some form of context that the content header added here is going to set a mime type to an XForm for posting.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
ctsims: Totally fair to say that it was copied over from prior code, although no reason to not address error handling now that we're looking into it.

I think the point is less that the content length is mandatory for the request, but moreso that an IOException here doesn't indicate "no content length could be found", but indicates "There's a fundamental problem with some IO Read/Write that occurred while trying to find or set content length". 

If the pipe is broken or something and we swallow this I/O exception here, there's no guarantee that the next I/O exception thrown in the "real" mandatory context is going to have the same value of information that this one does.

If this was a Long Parse failure or something that is relevant _only_ to the content length I think swallowing sounds fine. Given that the error would be with the request objects themselves, i don't think this method can take responsibility for the failure.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
ctsims: One other option would actually be to make the response processor have two different methods. 

Generally speaking in this context I'd expect an I/O failure to have a different semantic meaning than a generic exception, so there's some argument to be made that it makes sense for there to be different error handling paths. 

I think either of the two options laid out (Making PlainText an I/O Exception, or making there be two return paths in the error handling) feel like they should work to me, open to a third option as well.

src/main/java/org/commcare/core/network/AuthenticationInterceptor.java
ctsims: hm, the web team claims the current workflow _is_ working for them (testing with a local machine over http) which I think means that if the old code was trying this that it wasn't working. I'm going to have them test with the new retrofit code and see if it blocks their usage (which would actually be _good_) and then suggest that if that's true we accept authing as long as it is within a local network, IE: to one of

```
localhost
192.168.*
10.*
```
(We can skip the 172 range because it's annoying to encode and no one uses it). 

I actually wonder if any of these utils can just take a URL and answer from some method call whether it's part of a private IP range?

shubham1g5: So earlier only `ModernHTTPRequester` was enforcing it and not `HTTPRequestGenerator` which was doing majority of requests. `ModernHTTPRequester` was only getting utilized for `PostRequestActivity` and `QueryRequestActivity` which seems to me are rare use cases. This version is definitely going to block HTTP and hence I am going to try to bypass local IPs through this check.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: Well it doesn't sound odd to me rather a good security practice. Though I am guessing most people block both HTTPS -> HTTP and HTTP -> HTTPS. Though no one seems to raise questions on approach we are following.

src/main/java/org/commcare/core/network/AuthenticationInterceptor.java
ctsims: thanks. 

I think we may still want a configurable flag for this for non-dimagi domains ("Disable secure authentication requirement" or something), because even though this will be fine for Dimagi domains, any open source project that's using CommCare for something not-secure who don't have HTTPS will essentially have their apps brick after our next release.

Unfortunately I think the way we're going to have to roll this out to not have that problem is a bit complex if we don't want to brick those apps. I think we're going to need to need a feature flag where "Secure Endpoint Enforcement" is disabled by default, and have HQ start adding that flag (as automatically enabled) on all new builds. Alternatively, we can feature flag "Disable Secure Endpoint Enforcement", but set the feature flag to default to _on_ for now, communicate that the change is happening, and then pick a date/release where we set the feature flag to default to "off", which is probably the easiest option.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
shubham1g5: Makes sense. Different implementations of ContentBody can implement different length functions and so to be safe I am going to throw the exception from here.

PR #623, opened 2017-07-27
Returns NaN for max and min functions in case of an empty nodeset

shubham1g5:
Case: https://manage.dimagi.com/default.asp?259053#1370723


ctsims:
Solid, solid PR.

PR #624, opened 2017-08-04
Don't trigger any calculations in read only mode

amstone326:
Fixes https://manage.dimagi.com/default.asp?259554

cross-request: https://github.com/dimagi/commcare-android/pull/1782


ctsims:
Hey,

I'd be more comfortable moving this to 2.39, I think it's a bit more complex than it seems.

The calculate graph is _also_ used to mark nodes as visible/not based on their display conditions, so this will likely result in the app displaying additional questions/logic that wouldn't have been displayed initially. We will need to find a way to mark questions as visible/not from the restored saved form before this strategy will work.

amstone326:
Ah oops, makes sense. Will re-visit

PR #625, opened 2017-08-07
Persist sort info for an ItemsefBinding

amstone326:
Caught this while writing the UI test for this feature (sorting answer choices in a lookup table select Q)!

This was making it so that sorting only worked the first time you opened a form, which I hadn't noticed previously.


src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -134,5 +136,7 @@ public void writeExternal(DataOutputStream out) throws IOException {
         ExtUtil.write(out, new ExtWrapNullable(copyRef));
         ExtUtil.writeBool(out, labelIsItext);
         ExtUtil.writeBool(out, copyMode);
+        ExtUtil.write(out, sortRef);
```
wpride: Do these need to be wrapped in `ExtWrapNullable`?

amstone326: @wpride Yea I think you're right, fixed, thank you.

PR #626, opened 2017-08-09
Improve performance of DB model builds

ctsims:
Bootstrapping the database backed tree structures is super slow  at scale due to the initial walk. This allows them to initialize with a walk that only touches id's (pulled into core from the android/db backed code used in the CaseDB).

Still needs a platform-specific implementation for the storage backend.

Release Note: "Improvements to performance for apps that make use of large location trees"

### DIFF
--- a/src/main/java/org/commcare/cases/instance/StorageInstanceTreeElement.java
+++ b/src/main/java/org/commcare/cases/instance/StorageInstanceTreeElement.java
@@ -105,7 +105,7 @@ public abstract class StorageInstanceTreeElement<Model extends Externalizable, T
         }
         elements = new Vector<>();
         int mult = 0;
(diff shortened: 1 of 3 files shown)


ctsims:
@wpride the formplayer storage class boolean-using iterator should get an @override after this, the same way android did.

amstone326:
@ctsims This PR seems like a no-op to me, since nothing actually implements the new version of `iterate()` in a meaningful way. What am I missing?

ctsims:
@amstone326 Android implements the new version of iterate already, this just exposes the behavior to the core models as well as the android specific ones.

amstone326:
announce: "Improvements to performance for apps that make use of large location trees"

PR #628, opened 2017-08-11
Centralize/Unify Sync and Submit Logic 

ctsims:
A bunch of reworks of the testing infrastructure which centralize how syncs and submissions are working.

Submitting forms to the server now works from the CLI, as well as tests, and syncing is now capable of handling significantly more robust outcomes.

Did this to make it possible to test complex sync behaviors to replicate bugs, but long term we should be able to get more aligned on things like what processing paths exist.

Shouldn't pull until we figure out what to do with the replication in here. It shouldn't really be run with the unit tests.

Includes: https://github.com/dimagi/commcare-core/pull/627/files which should be reviewed first just because it's way easier.


src/test/java/org/commcare/mockapp/AsyncDataCollisionTests.java
```diff
@@ -0,0 +1,164 @@
...
+import org.javarosa.form.api.FormEntryPrompt;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.text.Normalizer;
+
+import static org.junit.Assert.assertEquals;
+
+/**
```
ctsims: note: This is not a proper unit test, but will be run with the unit tests right now. It should get fixed to run only during some sort of integration test cycle.

amstone326:
@ctsims Is this still relevant / something you want to get merged?

PR #629, opened 2017-08-14
Static analysis infrastructure + Use in relevancy condition evaluation

amstone326:
cross-request: https://github.com/dimagi/commcare-android/pull/1790


amstone326:
@ctsims Realized I completely forgot about the current() and relative ref stuff; don't review yet

src/main/java/org/javarosa/xpath/expr/XPathPathExpr.java
```diff
@@ -84,6 +86,7 @@ public TreeReference getReference() throws XPathUnsupportedException {
                 break;
             case XPathPathExpr.INIT_CONTEXT_RELATIVE:
                 ref.setRefLevel(0);
+                ref.setContextType(TreeReference.CONTEXT_INHERITED);
```
amstone326: this fixes a bug where `XPathPathExpr.getReference()` was not properly assigning the context type here. The reason this wasn't an issue for anything else is that there was previously no logic in the codebase which hinged on checking for the value `CONTEXT_INHERITED` (this PR changes that)

src/main/java/org/commcare/suite/model/Entry.java
```diff
@@ -86,18 +87,31 @@ public Text getText() {
...
-            if (cur instanceof ExternalDataInstance) {
-                //Copy the EDI so when it gets populated we don't keep it dependent on this object's lifecycle!!
-                copy.put(key, new ExternalDataInstance(((ExternalDataInstance)cur).getReference(), cur.getInstanceId()));
-            } else {
-                copy.put(key, cur);
+            if (limitingList == null || limitingList.contains(cur.getInstanceId())) {
+                // Make sure we either aren't using a limiting list, or the instanceid is in the list
+                if (cur instanceof ExternalDataInstance) {
+                    //Copy the EDI so when it gets populated we don't keep it dependent on this object's lifecycle!!
+                    copy.put(key, new ExternalDataInstance(((ExternalDataInstance)cur).getReference(), cur.getInstanceId()));
```
ctsims: I'm a bit confused by this statement. Is the "dependence" a memory leak (IE: The EDI ends up keeping a reference to this object)?

src/main/java/org/javarosa/xpath/analysis/XPathAccumulatingAnalyzer.java
```diff
@@ -0,0 +1,106 @@
...
+    public XPathAccumulatingAnalyzer() {
+        super();
+    }
+
+    public XPathAccumulatingAnalyzer(TreeReference contextRef) {
+        super(contextRef);
+    }
+
+    @Nullable
+    public Set<T> accumulate(XPathAnalyzable rootExpression) {
```
ctsims: Seems weird for us to aggregate a list and then put it into a set on response. Was there a specific reason for that?

src/main/java/org/javarosa/xpath/analysis/XPathAccumulatingAnalyzer.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // For all AccumulatingAnalyzers, it should be sufficient to handle a relative reference by
+    // applying the analyzer separately to the expression itself and to the context ref
+    @Override
+    public void doAnalysisForRelativeTreeRef(TreeReference expressionWithContextTypeRelative)
+            throws AnalysisInvalidException {
+
+        if (getContextRef() == null) {
+            throw new AnalysisInvalidException();
```
ctsims: I think it would be useful to add messages to these, since it will make it easier to track down false negatives (IE: We get back an invalid analysis: Since the process is recursive it'll be really tricky to figure out where and what ended up throwing the issue.

src/main/java/org/javarosa/xpath/analysis/XPathAnalyzer.java
```diff
@@ -0,0 +1,84 @@
...
+            throws AnalysisInvalidException {
+        // So that the default behavior is to do nothing
+    }
+
+    public XPathAnalyzer spawnSubAnalyzer(TreeReference subContext) {
+        XPathAnalyzer subAnalyzer = initSameTypeAnalyzer();
+        subAnalyzer.isSubAnalyzer = true;
+        subAnalyzer.originalContextRef = this.getOriginalContextRef();
+        subAnalyzer.contextRef = subContext;
+        this.subAnalyzers.add(subAnalyzer);
```
ctsims: The method of keeping these in a recursive list seems a bit odd to me. Given that we have control over the internal lifecycle of the subanalyser, the pattern of

sub = SpawnSub
for (subcontexts())
  -- sub.analyse()
this.joinSub(sub)


feels like it would be a bit cleaner (we only maintain the subcontexts for their lifecycle) and not require as much context to be retained in the root object.

I think it's a pretty small change to shift to that, is there a specific advantage oyu were looking for from the current design?

src/test/java/org/commcare/backend/session/test/EvalContextTests.java
```diff
@@ -0,0 +1,76 @@
...
+        List<String> instancesThatShouldBeIncluded = new ArrayList<>();
+        instancesThatShouldBeIncluded.add("casedb");
+        Assert.assertEquals(instancesThatShouldBeIncluded, ec.getInstanceIds());
+
+        // 2) Confirm the display condition was evaluated properly with the restricted context
+        Assert.assertTrue(FunctionUtils.toBoolean(relevancyCondition.eval(ec)));
+    }
+
+    @Test
+    public void testRestrictedEvalContextGeneration2() throws Exception {
```
ctsims: I think it would be helpful to also test here the pattern where we _can't_ analyse the substring for some reason, just to ensure we're tracking the codepaths.

src/main/java/org/commcare/suite/model/Entry.java
amstone326: That comment/code isn't from this PR at all; you can see it was the same a few lines above. All that's changed here is adding the `if` statement on line 107

src/main/java/org/javarosa/xpath/analysis/XPathAccumulatingAnalyzer.java
amstone326: See the class-level Javadoc for context on this. I wanted to emphasize in-code the point which I make there by not having the accumulation return a List, since that would give the impression that duplicate elements have a useful meaning. Since they don't, a Set seemed like the correct output.

src/main/java/org/javarosa/xpath/analysis/XPathAccumulatingAnalyzer.java
amstone326: Good point, added.

src/main/java/org/javarosa/xpath/analysis/XPathAnalyzer.java
amstone326: @ctsims So the problem I see with this suggestion is that the control flow really can't work the way that you're suggesting. Specifically, the `sub.analyze()` line can't be executed from within `XPathAnalyzer`, because this class has no knowledge of what the sub-analyzer is supposed to be analyzing or where; that information is known by `TreeReference.applyAndPropagateAnalyzer()` (which is where `spawnSubAnalyzer()` is called from). So in order to do what you're suggesting, `TreeReference.applyAndPropagateAnalyzer()` would have to have to make some call on the sub-analyzer _after_ the analysis was executed, telling it to propagate its results back up to its parent analyzer. That's definitely doable, but I don't think I see it as preferable, because then in the future, if other `XPathAnalyzable` classes have to generate sub-analyzers in the course of their analysis, they have to also remember to add this extra call. Whereas with the current setup, the aggregation happens automatically without any extra work.

src/test/java/org/commcare/backend/session/test/EvalContextTests.java
amstone326: @ctsims Sorry I'm not sure if I know what you mean. Do you mean an instance in which the analysis results in an `AnalysisInvalidException`? If so I can add that, but I would think that that sort of test belongs more in `StaticAnalysisTest.java` than here.

amstone326:
@ctsims Responded to the specific points of feedback. I'll do my best to get the profiling done tomorrow if you send the version of the app that you want me to use. If there are specific menu screens or even specific relevancy expressions that you know would be most useful to test, including that info would probably make things go faster.

src/main/java/org/commcare/suite/model/Entry.java
ctsims: aha! missed that entirely, sorry.

src/main/java/org/javarosa/xpath/analysis/XPathAccumulatingAnalyzer.java
ctsims: Hey,

That makes lots of sense, I think my question was more why we were gathering the intermediate results as a list in the first place, rather than starting with a set. One big advantage of a set is that it would let us add elements to process "in-place" rather than aggregating into a set of recursive lists and then aggregating those back down. For big expressions this approach feels like it'll generate a big tree of strings that then compress back down to 2 strings, when it would have worked to just keep track of the set of 2 strings during the process.

src/main/java/org/javarosa/xpath/analysis/XPathAnalyzer.java
ctsims: Yeah, I was assuming that the control flow logic for spawning and rejoining would need to be delegated back down to the same level as the current dispatch (IE: Inside of the xpath elements themselves currently). In the TreeReference implementation below the only difference would essentially be a line at the end of each iteration that says "originalAnalyzer.join(spawnedChildAnalyzer)" which knows how to put things back together.

I don't think it would change the future analyzer classes since all they'd be on the hook for would be implementing the "Join" method where they retrieve and process the results of their children. The valuable part to me of having that logic change is that it separates how we do the dispatch from what the analyzer is doing. 

Seems a bit cleaner to me, but also seems pretty easy to change in the future if there's a re-implementation that's needed, so no action needed if you don't think it's a valuable change at the moment.

src/test/java/org/commcare/backend/session/test/EvalContextTests.java
ctsims: That makes sense. Yeah, no strong feelings for where it goes. I think the advantage in my mind for testing here was to establish that it wouldn't break a live app (and wouldn't result in incorrect usage "in-flight", but probably actually tricky to trigger at this level currently.

PR #630, opened 2017-08-18
Refactoring for Side by Side Compact Numeric Entry

shubham1g5:
Case - https://manage.dimagi.com/default.asp?259573#1379573

- Generalise getFieldListHost(index) as getHostWithAppearance(index, appearance)

### DIFF
--- a/src/main/java/org/javarosa/form/api/FormController.java
+++ b/src/main/java/org/javarosa/form/api/FormController.java
@@ -143,13 +143,9 @@ public class FormController {
         return indexIsInFieldList(getFormIndex());
     }
 
-    private boolean indexIsInCompact(FormIndex index) {
+    public boolean indexIsInCompact(FormIndex index) {
         FormIndex compactHost = getHostWithAppearance(index, FormEntryController.COMPACT);
         return compactHost != null;
-    }
-
-    public boolean indexIsInCompact() {
-        return indexIsInCompact(getFormIndex());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/javarosa/form/api/FormController.java
```diff
@@ -261,31 +272,32 @@ public int stepToPreviousEvent() {
...
+    @Nullable
+    private FormIndex getHostWithAppearance(FormIndex child, String appearanceTag) {
         int event = mFormEntryController.getModel().getEvent(child);
 
         if (event == FormEntryController.EVENT_QUESTION || event == FormEntryController.EVENT_GROUP || event == FormEntryController.EVENT_REPEAT) {
             // caption[0..len-1]
             // caption[len-1] == the event itself
             // caption[len-2] == the groups containing this group
-            FormEntryCaption[] captions = mFormEntryController.getModel().getCaptionHierarchy();
+            FormEntryCaption[] captions = mFormEntryController.getModel().getCaptionHierarchy(child);
```
amstone326: What does this change represent?

src/main/java/org/javarosa/form/api/FormEntryController.java
```diff
@@ -29,6 +29,7 @@
     public static final int EVENT_REPEAT_JUNCTURE = 32;
 
     public final static String FIELD_LIST = "field-list";
```
amstone326: Can you add a comment above these 2 fields indicating that they are both appearance attributes for a group?

src/main/java/org/javarosa/form/api/FormController.java
shubham1g5: if we don't pass the `child` index it finds the captions for the current index in form. Though we want to get `captions` for the `child` index passed according to function def. It was working earlier as we were always passing the `child` as current index.

PR #636, opened 2017-08-30
Adds toString for ASTNode

shubham1g5:
This PR tries to provide a better error message for XPathException thrown [here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xpath/parser/Parser.java#L333) by adding a toString implementation for ASTNode. Currently we get an error like `Bad node: org.javarosa.xpath.parser.ast.ASTNodeAbstractExpr@44e69680` which doesn't provide any info around the contents of the Node which is hard to debug for the user in case he have a large overall expression.

### DIFF
--- a/src/main/java/org/javarosa/xpath/parser/ast/ASTNode.java
+++ b/src/main/java/org/javarosa/xpath/parser/ast/ASTNode.java
@@ -14,148 +14,152 @@ public abstract class ASTNode {
 
     private int indent;
 
-    private void printStr(String s) {
-        for (int i = 0; i < 2 * indent; i++)
-            System.out.print(" ");
-        System.out.println(s);
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        translate(this, sb);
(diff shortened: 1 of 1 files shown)


wpride:
@shubham1g5 is this still live? If so, couple questions:

1. Looks like this changes the method's behavior by not itself printing the output
2. Looks like we're losing some new lines that were being added by the calls to `System.out.println`
3. Is this being used anywhere?

amstone326:
@shubham1g5 Bumping Will's questions above ^^

shubham1g5:
Hey, I am closing this since in retrospect the value retunred by toString can be more confusing for errors. Though the `print` method right now doesn't get used anywhere while `toString` is currently getting used [here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xpath/parser/Parser.java#L333), and i was trying to display some node information to the user when we throw this expression.

PR #638, opened 2017-09-06
remove no-longer-helpful choices warning

dannyroberts:
when you call ItemsetBinding.setChoices and this.choices is not null


dannyroberts:
cc @gcapalbo

PR #639, opened 2017-09-11
Fixes timezone for date widget

shubham1g5:
Case: https://manage.dimagi.com/default.asp?261212

Reverting [this change ](https://github.com/dimagi/commcare-core/commit/a53cda2f99ef427b858c7415d6ec0761e3ef3bfe#diff-0abc238934c64586dac0e0aeadf616f7L319) seems to be fixing both `NepaliDateWidget` and `convertToNepaliString` for me which fails in case of applying `UTC` timezone. 

Also we are using default timezone here while converting a date string to milis [here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xform/util/CalendarUtils.java#L402) so the opposite conversion from mills to date string should  use the same.

### DIFF
--- a/src/main/java/org/javarosa/xform/util/CalendarUtils.java
+++ b/src/main/java/org/javarosa/xform/util/CalendarUtils.java
@@ -346,7 +346,7 @@ public class CalendarUtils {
     }
 
     public static UniversalDate fromMillis(long millisFromJavaEpoch) {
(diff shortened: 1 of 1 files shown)


ctsims:
@shubham1g5 I think this is going to result in failures, and that the reason this works disparately is that our timezones are on the opposite side of the line, but it makes the most sense for now to revert to the original behavior since it makes more sense for this to fail in Boston than to fail in Nepal.

PR #641, opened 2017-09-14
use maven regex lib instead of vendorizing the jar

dannyroberts:


build.gradle
```diff
@@ -59,7 +59,7 @@ sourceSets {
 dependencies {
       compile 'kxml2:kxml2:2.3.0'
       compile group: 'javax.ws.rs', name: 'javax.ws.rs-api', version: '2.0.1'
-      compile files('lib/regexp-me.jar')
+      compile group: 'regexp', name: 'regexp', version: '1.3'
```
ctsims: Is this the same jar? regexp-me was a different version back in the day (java 1.3 compatible). Probably fine either way since 1.3/javame compatibility isn't important anymore, but not sure whether we should QA anything specific if it's not an identical version

dannyroberts: I'm not sure how to tell if they're the same or not, but if I'm looking in the right place, the binary content of the jars does differ, and when unzipped the corresponding binary class files differ.

That seems pretty inconclusive, but let's assume they're not the same version. Do you know how we can find out what version regexp-me.jar is from so we can at least compare release notes or something?

dannyroberts: also, I assume there's some good reason we don't use java.utils.regexp?

ctsims: Hey,

We only use 2-3 methods from the regex code in reality, I think you'd just need to validate that they keep working with the new libs? 

There was no specific intention on formplayer or commcare android for using that particular library. JavaME (on nokia phones) didn't have `java.utils.regex`, and regex-me was the only thing that fit on-device without blowing up the Jar. Now the only important thing is whatever we replace it with follows the same regex semantics

dannyroberts: Fair enough. I would think the right answer is to either trust tests or if we don't, add more tests in the same vein.

Currently we have
```
formplayer/libs/commcare/src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
520:        testEval("regex('12345','[0-9]+')", null, ec, Boolean.TRUE);
521:        testEval("regex('12345','[')", null, ec, new XPathException());
539:        testEval("regex('aaaabfooaaabgarplyaaabwackyb', 'a*b')", null, null, Boolean.TRUE);
540:        testEval("regex('photo', 'a*b')", null, null, Boolean.FALSE);
```

That doesn't seem super duper comprehensive, are there any specific patterns you want to make sure behave the same way? In my mind the subtle differences often come out in terms of _which substring matches_ rather than whether or not there's a match (i.e. does ('ababc', 'a*b') match on 'ab' or 'abab'?). Since we're only doing yes/no matching (right?), I think picking a few more examples to add to the test that are substantially different from the ones already there would suffice for me.

ctsims: Agreed.

dannyroberts:
Oh, also this is only going into effect on formplayer at this point, not sure how much that factors in. It certainly makes it much easier to fix if it goes wrong

amstone326:
@dannyroberts Is this still relevant? If so, what's it waiting on? (trying to clear out PRs that have been sitting open for ages)

dannyroberts:
I've only waited this long to respond because I'm embarrassed by my answer. Basically this is good cleanup that was helpful to me as a workaround at the time, but which isn't pressing anymore, and which I'm not pushing forward on at the moment largely because I'm no longer personally working much with the code base. I guess that means you can close it.

amstone326:
@dannyroberts Gotcha, and it's not mergeable as-is?

dannyroberts:
Well, I think it pretty much is, it's really just a matter of being sure that this standard library is functionally equivalent to the vendorized jar it's replacing. I think the only thing left to do is add a few more tests as I suggested in https://github.com/dimagi/commcare-core/pull/641#discussion_r139227568. Looking back over this conversation, that actually seems pretty easy, so I think it's worth me just finding 30 minutes to add a few tests. If any of them fail, then i'll just close this

amstone326:
That sounds good, thanks!

amstone326:
@dannyroberts Sorry to pester but re-pinging about the status of this

dannyroberts:
@amstone326 I think this is good to go once you review https://github.com/dimagi/commcare-core/pull/776 and tests pass. (I assume jenkins is going to eventually report build status on this PR.)

dannyroberts:
@amstone326 I think this is good to go once you review https://github.com/dimagi/commcare-core/pull/776 and tests pass. (I assume jenkins is going to eventually report build status on this PR.)

dannyroberts:
@amstone326 I think this is good to go once you review https://github.com/dimagi/commcare-core/pull/776 and tests pass. (I assume jenkins is going to eventually report build status on this PR.)

amstone326:

PR #642, opened 2017-09-19
In Progress: Partial Load Optimizations

ctsims:
This code has two major sets of changes.

1) New Optimization: SubContext Evaluation
* Static analyzer which is capable of identifying all references which are relevant to a set of XPath expressions.
* Refactors the Static Analysis tool to make the second analysis implementation slightly easier. 
* Implementation using the analyzer to create a subcontext for Itemset evaluation which lets the instances know that during an itemset eval they only need to make certain information available.

2) SubTree Expansion for Storage Backed Trees

This optimization essentially allows a storage backed xml object to only deserialized a limited amount of each element as-needed by a subcontext. So if in a subcontext only "name" and "id" are relevant to the query context, the root can load those from metadata and avoid loading the full xml object from storage.


ctsims:
Adding @amstone326 for review since this uses (and adapts/makes some changes to) your new static analysis system.

amstone326:
@ctsims Can you add a PR description to explain what this is doing at a high-level?

ctsims:
@amstone326 Right, sorry, adding now

ctsims:
@amstone326 updated description

src/test/java/org/commcare/fixtures/test/IndexedFixtureTests.java
```diff
@@ -68,4 +75,48 @@ public void errorOnMaliciousSchemaTest() throws XPathSyntaxException, Unfullfill
             XmlPullParserException, IOException, InvalidStructureException {
         ParseUtils.parseIntoSandbox(getClass().getResourceAsStream("/malicious-indexed-fixture.xml"), sandbox, true);
     }
+
+    @Test
+    public void testPartialQueryLoads() throws XPathSyntaxException, UnfullfilledRequirementsException,
```
amstone326: These 2 tests are identical except for the file path passed into `getResourceAsStream()`. Can you refactor them to use a shared helper?

src/test/java/org/commcare/fixtures/test/IndexedFixtureTests.java
```diff
@@ -68,4 +75,48 @@ public void errorOnMaliciousSchemaTest() throws XPathSyntaxException, Unfullfill
...
+        String exprString = "instance('testfixture')/test/entry[@filter_attribute = 'pass'][true() and filter_one = 'pass']/name";
+        XPathExpression expr = XPathParseTool.parseXPath(exprString);
+
+        cache.addTreeReferencesToLimitedScope(new TreeReferenceAccumulatingAnalyzer(ecForTest).accumulate(expr));
+
+        CaseTestUtils.xpathEvalAndAssert(ecForTest, exprString, "succeed");
+    }
+
+    @Test
+    public void testPartialQueryLoadsInBulkMode() throws XPathSyntaxException, UnfullfilledRequirementsException,
```
amstone326: I'm gathering that this test is supposed to test a different code path from the first test. But I'm not exactly clear on what that disparate path is, and I also don't see anything that checks whether that code path was actually executed.

src/main/java/org/commcare/cases/query/ScopeLimitedReferenceRequestCache.java
```diff
@@ -0,0 +1,127 @@
...
+
+public class ScopeLimitedReferenceRequestCache implements QueryCache {
+
+    //TODO: Do we get anything from sealing the scope other than vague confidence?
+
+    private final String DEFAULT_INSTANCE_KEY = "/";
+
+    // Maps instance names to the set of tree references that are "in scope" (i.e. that we must be
+    // able to evaluate) for that instance in the current context
+    private HashMap<String, Set<TreeReference>> map = new HashMap<>();
```
amstone326: This variable could use a better name

src/main/java/org/commcare/cases/instance/IndexedFixtureChildElement.java
```diff
@@ -42,6 +60,171 @@ protected TreeElement cache(QueryContext context) {
...
+
+        //If we don't already have that list, build it (or detect that it's won't be possible and tell
+        //the cache to not try.
+
+        Set<TreeReference> referencesInScope =
+                cache.getInScopeReferences(this.getInstanceName());
+
+        StorageIndexedTreeElementModel model = parent.getModelTemplate();
+
+        //TODO: Make sure this doesn't recursively call cache...
```
amstone326: What does this mean and is it a concern?

src/main/java/org/commcare/cases/instance/IndexedFixtureChildElement.java
```diff
@@ -42,6 +60,171 @@ protected TreeElement cache(QueryContext context) {
...
+            return limitedScope;
+        }
+
+        //If we don't already have that list, build it (or detect that it's won't be possible and tell
+        //the cache to not try.
+
+        Set<TreeReference> referencesInScope =
+                cache.getInScopeReferences(this.getInstanceName());
+
+        StorageIndexedTreeElementModel model = parent.getModelTemplate();
```
amstone326: It would be nice if the code from here through line 191 was moved into a helper method named something like `checkIfRequiredReferencesAreAllIndexed()` or something to that effect. I'm envisioning that the method would take in the `referencesInScope` argument above, and either return the populated `columnNameCacheLoads` object if it was successful, or `null` if it failed.

src/test/java/org/javarosa/xpath/analysis/test/TreeReferenceAccumulatorTest.java
```diff
@@ -0,0 +1,157 @@
...
+        TreeReferenceAccumulatingAnalyzer analyzer;
+        if (context == null) {
+            analyzer = new TreeReferenceAccumulatingAnalyzer();
+        } else {
+            analyzer = new TreeReferenceAccumulatingAnalyzer(context);
+        }
+
+        Set<TreeReference> results = analyzer.accumulate(expression);
+        if (results == null) {
+            return null;
```
amstone326: Why does this just return null? If there is anything in `matches`, shouldn't the test fail at this point?

src/test/java/org/javarosa/xpath/analysis/test/TreeReferenceAccumulatorTest.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        Set<TreeReference> expressions = new HashSet<>();
+
+        for (String match : matches) {
+            expressions.add(XPathReference.getPathExpr(match).getReference());
+        }
+
+        Assert.assertEquals(text, expressions, results);
```
amstone326: Would be nice if the assertion message contained something more descriptive than just `text`

src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
```diff
@@ -227,4 +254,6 @@ protected TreeElement cache() {
     }
 
     protected abstract TreeElement cache(QueryContext context);
```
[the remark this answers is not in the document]
ctsims: It will eventually, it just involves a fair amount more testing, and the meta data semantics can be slightly more complex, so I'm holding off on that until I can test it out and see the impact/value.

src/main/java/org/commcare/cases/query/QuerySensitiveTreeElement.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Get a child element with the given name and occurence position (multiplicity)
+     *
+     * @param name         the name of the child element to select
+     * @param multiplicity is the n-th occurence of an element with a given name
+     */
+    T getChild(QueryContext context, String name, int multiplicity);
+
+    boolean hasChildren(QueryContext context);
```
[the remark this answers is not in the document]
ctsims: just consistency of not having to change the signature of this interface in the event that a future implementation would require the context to be around. It makes the most sense to me that implementing classes provide all the hooks up front, rather than pick and choose, which seems a bit error prone to me.

src/main/java/org/commcare/cases/query/ScopeLimitedReferenceRequestCache.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * A list of instance names for instances which have flagged that they cannot utilize
+     * the limited scope.
+     */
+    private Set<String> excludedInstances = new HashSet<>();
+
+    private HashMap<String, String[]> instanceScopeLimitCache = new HashMap<>();
+
+    public ScopeLimitedReferenceRequestCache() {
```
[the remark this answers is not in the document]
ctsims: All QueryCache's get used the same way as outlined here: https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/cases/query/QueryCache.java#L10

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1362,19 +1369,59 @@ public void populateDynamicChoices(ItemsetBinding itemset, TreeReference curQRef
...
         }
+        if(reporter != null) {
+            InstrumentationUtils.printAndClearTraces(reporter, "ItemSet Field Population");
+        }
+
         itemset.setChoices(choices);
     }
 
+    private EvaluationContext getPotentiallyLimitedScopeContext(EvaluationContext ec, ItemsetBinding itemset) {
+        TreeReferenceAccumulatingAnalyzer analyzer = new TreeReferenceAccumulatingAnalyzer(ec);
```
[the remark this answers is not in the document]
ctsims: Yeah, it should be pretty helpful in the future for a lot of different calculations. Case Lists, Mobile UCR's (which are implemented as entity lists), and actually a fair number of just "Big Internal" requests are likely to utilize this pattern eventually.

src/main/java/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -476,7 +476,8 @@ public TreeReference relativize(TreeReference parent) {
         if (parent.isParentOf(this, false)) {
             TreeReference relRef = selfRef();
             for (int i = parent.size(); i < this.size(); i++) {
-                relRef.add(this.getName(i), INDEX_UNBOUND);
+                int index = this.getMultiplicity(i) == INDEX_ATTRIBUTE ? INDEX_ATTRIBUTE : INDEX_UNBOUND;
```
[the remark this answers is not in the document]
ctsims: Yeah, this was just broken before (attributes don't need to start with an unbound multiplicity) and was causing confusion / weird behavior.

src/main/java/org/javarosa/xpath/analysis/TreeReferenceAccumulatingAnalyzer.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void doNormalTreeRefAnalysis(TreeReference treeRef) throws AnalysisInvalidException {
+        addResultToList(treeRef.removePredicates());
+    }
+
+    // For all AccumulatingAnalyzers, it should be sufficient to handle a current() reference by
+    // applying the analyzer separately to the expression itself and to the context ref
+    @Override
+    public void doAnalysisForTreeRefWithCurrent(TreeReference expressionWithContextTypeCurrent)
```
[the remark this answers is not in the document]
ctsims: Mostly I didn't want to change the semantics of the other analyzer and wasn't sure if we wanted to lock ourselves into this approach. I'm not sure whether it's possible that there are circumstances where the other approach is possible and this approach is not, for example, and didn't want to lock the other analyzer into this strategy if it's wasn't strictly necessary.

src/test/java/org/commcare/fixtures/test/IndexedFixtureTests.java
ctsims: There is no effective mechanism to accomplish that (ensuring that within the query framework that specific code paths were/were not triggered) currently, and it's a huge lift to build one, unfortunately. I think it's a pretty big chunk of tech debt to use the trace system to be able to detect what occurred during a query.

src/test/java/org/javarosa/xpath/analysis/test/TreeReferenceAccumulatorTest.java
ctsims: The way these errors are output would be something like

```
Fail: TreeReferenceAccumulatorTest .testParsesWithContext
"instance('expr')/ebase/element[@attr = current()/context_value]/value"
expected:
                "instance('expr')/ebase/element/value",
                "instance('expr')/ebase/element/@attr",
                "instance('baseinstance')/base/element/context_value");
actual:
...
```
which seemed like the most helpful information that is available. junit also cuts off text aggressively, so I wanted to optimize for giving people as much context on what case was failing as possible (which is the XPath expression provided).

src/test/java/org/javarosa/xpath/analysis/test/TreeReferenceAccumulatorTest.java
ctsims: hm, good point. fixing

ctsims:
@amstone326 feedback addressed, moving on to commcare-anrdoid pr

ctsims:
@amstone326 looks like there wasn't any on android, actually,  so this is g2g again for review

src/main/java/org/commcare/cases/instance/IndexedFixtureChildElement.java
ctsims: refactored this below to be more explict about what's actually taking care of what things

src/main/java/org/commcare/cases/query/QuerySensitiveTreeElement.java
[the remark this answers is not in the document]
amstone326: Hm, I don't think I agree with this. If this method isn't being used, it's adding the appearance of additional complexity that isn't there (to something that already appears really complex). I also don't see how leaving it out creates any more room for error.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -318,7 +320,7 @@ private void expandReferenceAccumulator(TreeReference sourceRef, DataInstance so
             predicates = predCopy;
         }
 
-        AbstractTreeElement node = sourceInstance.resolveReference(workingRef);
+        AbstractTreeElement node = sourceInstance.resolveReference(workingRef, this);
```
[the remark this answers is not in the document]
amstone326: @ctsims still not sure why this change was made here; it looks like the sort of change that would have implications beyond this PR

src/test/java/org/javarosa/xpath/analysis/test/TreeReferenceAccumulatorTest.java
amstone326: got it, makes sense

src/test/java/org/javarosa/xpath/analysis/test/TreeReferenceAccumulatorTest.java
amstone326: Cool. Could you also add a test that expects an analysis failure? similar to https://github.com/dimagi/commcare-core/blob/master/src/test/java/org/javarosa/xpath/analysis/test/StaticAnalysisTest.java#L86-L88

amstone326:
@ctsims Thanks for the changes. I left a few more follow-up questions/requests, and also committed a change that I explained in a comment above.

src/test/java/org/javarosa/xpath/analysis/test/TreeReferenceAccumulatorTest.java
ctsims: The tests in testInvalidReferences() on line 98 test for analysis failure.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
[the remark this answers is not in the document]
ctsims: This change needs to be made for the Context to be available when the limited scope cache is used.

"resolveReference" is where the actual data is retrieved from the Instance Child element, and before it didn't have access to the context. It needs that access to use the new limited scope oiptimization

src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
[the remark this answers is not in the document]
ctsims: I'm not sure if adding a note will be more helpful than confusing. The new changes to cache() added to IndexedFixtureChildElement are a new optimization pattern entirely that just isn't  implemented in the other data models because it isn't as useful for them yet (similar to the way that the CaseChildElement is able to optimize index/* queries differently than the other data types).

I don't think it's specific technical debt to re-implement for cases and ledgers (IE: Nothing about the way they work is deprecated now), so I'll add a card to the performance board which mentions that it's low hanging fruit to re-implement this new pattern if it would be useful for optimizing something about those models later.

src/main/java/org/commcare/cases/query/QuerySensitiveTreeElement.java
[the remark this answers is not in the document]
ctsims: Ah, I understand. Sorry, I was thinking about the wrapper.

You're right, this didn't make sense. I removed them and also went through and made sure everything was tagged with @Override to make sure that we wouldn't miss implementing/de-implementing

ctsims:
@amstone326 Made changes based on your last comments and re-tested with your merge, everything seems to be working well still.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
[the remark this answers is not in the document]
amstone326: ah nice makes sense, thanks

PR #643, opened 2017-09-20
Fix Performance for back-to-back queries with mixed contexts

ctsims:
When we loaded itemsets back-to-back we would end up using the default cacher to store the results of large query results _outside_ of the QueryContext, which would mean that when we loaded the second time the results would come from the cache and fail to make it into RecordSetResultCache, preventing additional bulk optimizations from firing.

Long story short, something that took ~20s the first time you load it can take 3-8 times as long the next time, and end up triggering OOM issues due to loading things that should be bulk in serial


ctsims:
I'm self-merging this because I don't think anyone else has sufficient context to give feedback at the moment.

PR #645, opened 2017-09-21
Fix for a bug where Lazy Nodesets all relied on the same JVM-Wide static lock

ctsims:
Since Boolean.FALSE is a static object only one un-evaluated XPathLazyNodeset object could be evaluating at any given time.

### DIFF
--- a/src/main/java/org/commcare/suite/model/Entry.java
+++ b/src/main/java/org/commcare/suite/model/Entry.java
@@ -22,7 +22,7 @@ import java.util.Set;
 import java.util.Vector;
 import java.util.concurrent.Callable;
 
-import io.reactivex.Observable;
+import io.reactivex.Single;
 
 /**
  * Describes a user-initiated action, what information needs to be collected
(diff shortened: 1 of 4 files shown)


shubham1g5:
Just converted Observable to Single to make it semantically correct that the returned Observable will only emit a single event.

ctsims:
Hey Shubham,

Good catch on the Single pattern, that makes a lot of sense. Pulling these now.

After getting a listing of the Android Lifecycle events that were firing, I noticed that the onStart of the new activity was firing but onDestroy of this original wasn't (as you identified earlier), which would imply that the new Activity wasn't getting through its onStart.

I triggered the behavior and entered the debugger and did a quick accounting of all of the threads that were executing and noticed that the main thread was waiting on the synchronous() lock to evaluate its expression and got _really_ confused that the two threads had the same lock boolean (since on face value it looks like it's a unique class member). 

This error may actually be fortuitous, since I realized that there are a few other places in code where we're relying on cross-JVM locks and I think they are responsible for some thread contention in the FormPlayer web service. I can't even imagine how hard it would have been for us to track it down there!

PR #648, opened 2017-09-22
Allow concurrent install/upgrade actions between sandboxes

ctsims:
Shift Resource Manager to use a platform level lock rather than a VM level lock.

I believe this change should be safe for FormPlayer and Android clients, although once formplayer is properly merged back into master we should validate.

### DIFF
--- a/src/main/java/org/commcare/resources/ResourceManager.java
+++ b/src/main/java/org/commcare/resources/ResourceManager.java
@@ -22,12 +22,6 @@ public class ResourceManager {
     protected final ResourceTable upgradeTable;
     protected final ResourceTable tempTable;
 
-    /**
-     * Lock to synchronize update logic so that multiple threads don't try to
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/resources/ResourceManager.java
```diff
@@ -22,12 +22,6 @@
     protected final ResourceTable upgradeTable;
     protected final ResourceTable tempTable;
 
-    /**
-     * Lock to synchronize update logic so that multiple threads don't try to
-     * manipulate tables simultaneously.
-     */
-    protected final static Object updateLock = new Object();
-
     public ResourceManager(CommCarePlatform platform,
```
dannyroberts: does formplayer make a new `CommCarePlatform` per thread?

dannyroberts: (offlined: yes)

PR #649, opened 2017-09-25
Refactor how detail styles are handled

wpride:
cross-request: https://github.com/dimagi/formplayer/pull/479

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityDetailSubscreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityDetailSubscreen.java
@@ -5,6 +5,7 @@ import org.commcare.core.graph.util.GraphException;
 import org.commcare.core.graph.util.GraphUtil;
 import org.commcare.suite.model.Detail;
 import org.commcare.suite.model.DetailField;
+import org.commcare.suite.model.Style;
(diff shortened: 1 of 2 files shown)


dannyroberts:
> Bug where Style fields remained despite the data being removed due to being null or blank
Can you show me which lines in this PR actually change something vs. being part of the refactor (or explain how moving the code changes things?

src/cli/java/org/commcare/util/screen/EntityDetailSubscreen.java
```diff
@@ -46,21 +49,28 @@ public EntityDetailSubscreen(int currentIndex, Detail detail, EvaluationContext
                 dataTemporary.add(data);
                 headersTemporary.add(createHeader(field, subContext));
                 rowTemporary.add(createRow(field, subContext, data));
+                stylesTemporary.add(createStyle(field));
```
wpride: @dannyroberts this is the important change - moving the Style[] accumulation inside this `if` statement that only adds when the value of the detail is non-empty. Previously we were using the `EntityScreen`'s built in `getStyles` function which didn't perform this check.

dannyroberts: Ah nice, makes sense!

PR #651, opened 2017-09-27
Fix failure mode for missing index lookups

ctsims:
Replicate 
https://github.com/dimagi/commcare-core/pull/558

onto 2.39. Fixes issues that come when a derived query set gets back a missing match from its context.


ctsims:
@wspride er, well, that would explain why my local testing didn't work. Thanks.

PR #653, opened 2017-09-27
Remove old unused <submission> implementation and add new handlers

ctsims:
Basic wiring needed for implementations of
https://github.com/dimagi/commcare-core/wiki/In-Form-HTTP-Requests

Platforms can implement the:
FormSendCalloutHandler 

interface and register with the FormDef before initialization to enable those requests.

Note: This PR strips out all of the old plumbing for <submission> elements that we very briefly used in JavaRosa for doing things like submitting over SMS. That code was unused in any modern implementation.


snopoke:
Pretty cool. Do we have projects planning to use this?

ctsims:
Hey Simon! We have one project which is planning to use this only on the webapps platform layer (where it would make sense to presume the existence of a persistent web connection for sure!).

Before we push it down to the mobile layer we'll almost certainly need to expand the robustness of the events/action subsystem implementation since right now there are no user triggerable events.

Due to a super, super, super annoying 10 year old decision we idiotically implemented the <trigger> control (which is supposed to be a button) as  a label with no user input control (don't even ask), so now we need to dig ourselves out of that technical debt to provide users with the ability to trigger the callout. 

It's also synchronous (literally in the UI thread) currently, which is fine on the web, but we'll need to make a way for it to be "Blocking, but not literally taking over the UI thread" on the mobile platform before we are comfortable rolling it out there. (For a long time on mobile we've needed a way to wrap interactions with the Form in a way that blocks interacting with the form, but doesn't block the UI).

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1842,11 +1846,11 @@ public void addSubmissionProfile(String submissionId, SubmissionProfile profile)
         submissionProfiles.put(submissionId, profile);
     }
 
-    public SubmissionProfile getSubmissionProfile() {
+    public SubmissionProfile getSubmissionProfile(String id) {
         //At some point these profiles will be set by the <submit> control in the form.
```
wpride: Comment is out of date

src/main/java/org/javarosa/core/model/SubmissionProfile.java
```diff
@@ -18,71 +20,53 @@
...
     }
 
     @Override
     public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
-        ref = (XPathReference)ExtUtil.read(in, new ExtWrapTagged(XPathReference.class), pf);
-        method = ExtUtil.readString(in);
-        action = ExtUtil.readString(in);
-        mediaType = ExtUtil.nullIfEmpty(ExtUtil.readString(in));
-        attributeMap = (Hashtable<String, String>)ExtUtil.read(in, new ExtWrapMap(String.class, String.class), pf);
+        this.resource = ExtUtil.readString(in);
```
wpride: Does changing these `read`- and `writeExternal` signatures not require a migration of some kind?

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -852,56 +852,67 @@ public void parseSetValueAction(ActionController source, Element e) {
...
+        //For Validation Only
         String method = submission.getAttributeValue(null, "method");
-        String action = submission.getAttributeValue(null, "action");
 
-        SubmissionParser parser = new SubmissionParser();
-        for (SubmissionParser p : submissionParsers) {
-            if (p.matchesCustomMethod(method)) {
-                parser = p;
-            }
+        if(!("get".equals(method))) {
```
wpride: OOC why are we only supporting `GET`

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -852,56 +852,67 @@ public void parseSetValueAction(ActionController source, Element e) {
...
         }
 
-        //These two might exist, but if neither do, we just assume you want the entire instance.
-        String ref = submission.getAttributeValue(null, REF_ATTR);
-        String bind = submission.getAttributeValue(null, BIND_ATTR);
+        String replace = submission.getAttributeValue(null, "replace");
 
-        XPathReference dataRef = null;
-        boolean refFromBind = false;
+        if(!("text".equals(replace))) {
```
wpride: Why do we have these params at all if we're enforcing one value?

src/test/java/org/javarosa/test_utils/MockFormSendCalloutHandler.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Map;
+
+/**
+ * Mocks the responses needed for form <submission/> responses in tests
+ *
+ * Created by ctsims on 9/27/2017.
+ */
+
+public class MockFormSendCalloutHandler implements FormSendCalloutHandler {
```
wpride: Great test coverage

src/main/java/org/javarosa/core/model/SubmissionProfile.java
ctsims: XForm models change too often, so instead of relying on the serialized form, we use the serialized version as a cached object. If deserialization fails, we just reparse the form from the original object, so no migrations should be needed.

src/main/java/org/javarosa/xform/parse/XFormParser.java
ctsims: POST just has much more complex semantics to implement, and we don't have an obvious use case for it.

src/main/java/org/javarosa/xform/parse/XFormParser.java
ctsims: This is an XForms specification, we're just implementing it. (Our XForms would be invalid if thy didn't define the parameter). This also protects us against accidentally parsing something the form would expect us to understand that we do not (IE: If we got a form from another platform with mode=POST, the validator should ensure that we fail and explain that we can't support it)

PR #654, opened 2017-09-28
Killable Queries 

ctsims:
Introduced a concept to the EC/QC stack of an 'abandonable' query object.

This lets us signal to a background thread that a really long running part of a query should be abandoned in-place in a way that is comfortable letting the entire Evaluation/Query context crash out.

Used to prevent badge calculations continue to process after they had already been disposed of.

We can expand on this a bit in master, didn't want to touch more code than I had to for the release proximity

### DIFF
--- a/src/main/java/org/commcare/suite/model/Entry.java
+++ b/src/main/java/org/commcare/suite/model/Entry.java
@@ -23,6 +23,7 @@ import java.util.Vector;
 import java.util.concurrent.Callable;
 
 import io.reactivex.Single;
+import io.reactivex.functions.Action;
 
 /**
  * Describes a user-initiated action, what information needs to be collected
@@ -157,14 +158,22 @@ public abstract class Entry implements Externalizable, MenuDisplayable {
     }
 
     @Override
-    public Single<String> getTextForBadge(final EvaluationContext ec) {
+    public Single<String> getTextForBadge(EvaluationContext ec) {
+        final EvaluationContext[] abandonableContext =
+                new EvaluationContext[] {ec.spawnWithCleanLifecycle()};
         if (display.getBadgeText() == null) {
             return Single.just("");
         }
         return Single.fromCallable(new Callable<String>() {
             @Override
             public String call() throws Exception {
-                return display.getBadgeText().evaluate(ec);
+                return display.getBadgeText().evaluate(abandonableContext[0]);
+            }
+        }).doOnDispose(new Action() {
+            @Override
+            public void run() throws Exception {
+                abandonableContext[0].signalAbandoned();
+                abandonableContext[0] = null;
             }
         });
     }
--- a/src/main/java/org/commcare/suite/model/Entry.java
+++ b/src/main/java/org/commcare/suite/model/Entry.java
@@ -159,23 +159,10 @@ public abstract class Entry implements Externalizable, MenuDisplayable {
 
     @Override
     public Single<String> getTextForBadge(EvaluationContext ec) {
-        final EvaluationContext[] abandonableContext =
-                new EvaluationContext[] {ec.spawnWithCleanLifecycle()};
         if (display.getBadgeText() == null) {
             return Single.just("");
         }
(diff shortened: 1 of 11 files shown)


ctsims:

src/main/java/org/javarosa/core/model/condition/LifecycleSignaler.java
```diff
@@ -0,0 +1,42 @@
...
+            throw new RequestAbandonedException();
+        }
+    }
+
+    @Override
+    public void signalAbandoned() {
+        this.isAbandoned = true;
+    }
+
+    public static void AssertNotAbandoned(Abandonable abandonable) {
```
amstone326: why is the first letter of the method name capitalized? here and below

src/main/java/org/commcare/suite/model/Entry.java
```diff
@@ -157,14 +158,22 @@ public String getDisplayText() {
     }
 
     @Override
-    public Single<String> getTextForBadge(final EvaluationContext ec) {
+    public Single<String> getTextForBadge(EvaluationContext ec) {
+        final EvaluationContext[] abandonableContext =
```
amstone326: What is the reason for making this an array of ECs?

src/main/java/org/commcare/suite/model/Entry.java
```diff
@@ -157,14 +158,22 @@ public String getDisplayText() {
...
         if (display.getBadgeText() == null) {
             return Single.just("");
         }
         return Single.fromCallable(new Callable<String>() {
             @Override
             public String call() throws Exception {
-                return display.getBadgeText().evaluate(ec);
+                return display.getBadgeText().evaluate(abandonableContext[0]);
+            }
+        }).doOnDispose(new Action() {
```
shubham1g5: Same needs to go into Menu.java implementation. Better to make a default implementation for both Menu and Entry class at this point.

src/main/java/org/javarosa/core/model/condition/LifecycleSignaler.java
ctsims: These are static methods, so the name is capitalized to differentiate it from the class method.

src/main/java/org/commcare/suite/model/Entry.java
ctsims: Good point. I'll unify a thread thing for "Evaluating an EC with these semantics."

src/main/java/org/commcare/suite/model/Entry.java
ctsims: This lets the the "dispose" action remove the reference to the abandonable context immediately. Since the Callable object is going into this whole event handling subsystem, I wanted to be super careful about leaving around references to the lifecycle objects and risking them being memory leaks.

src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java
```diff
@@ -200,11 +201,24 @@
...
+     *                                   throw a RequestAbandonedException. Callers should not
+     *                                   generally catch that exception unless they rethrow it
+     *                                   or another exception, but they should anticipate that
+     *                                   they may need to clean up if the bulk read doesn't complete
      */
-    void bulkRead(LinkedHashSet<Integer> ids, HashMap<Integer, E> recordMap);
-}
\ No newline at end of file
+
+    void bulkRead(LinkedHashSet<Integer> cuedCases, HashMap<Integer, E> recordMap, Abandonable abandonable) throws RequestAbandonedException;
```
amstone326: @ctsims The build is failing b/c this needs an import statement for `RequestAbandonedException`. And shouldn't the overriding methods have this added to their declaration as well?

ctsims: Hey, yeay, sorry, not quite done putting in those changes, sorry, pushed early.

PR #660, opened 2017-10-04
Add FormDef.initialize() path for reloading incomplete

wpride:
When re-initializing `FormDef`s in Formplayer we need to re-initializing all the external database hooks but often do **not** want to re-initialize the triggerable's. In general we only want to re-initialize the triggerables if we're loading an incomplete form as the user database might have changed between form entry sessions.

Currently this is a divergence between Android and Web Apps because Formplayer needs to re-initialize this object for each request (IE for each question answered) while Android can simply persist the FormDef object and so only calls `FormDef.initialize()` once. This adds a path to allow Formplayer to likewise only call the `initAllTriggerables` method once except in the situation outlined above.


amstone326:
@wpride Is this ready? no tags

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1458,38 +1458,51 @@ public void readExternal(DataInputStream dis, PrototypeFactory pf) throws IOExce
...
+     *                                changes to user databases
      */
-    public void initialize(boolean newInstance, boolean isCompletedInstance,
-                           InstanceInitializationFactory factory, String locale, boolean isReadOnly) {
+    public void initialize(boolean newInstance,
+                           boolean isCompletedInstance,
+                           InstanceInitializationFactory factory,
+                           String locale,
+                           boolean isReadOnly,
+                           boolean reloadingIncompleteForm) {
```
amstone326: If I'm reading this correctly, all of the previously-existing versions of `initialize()` now blindly pass in `true` for this argument. But how are you sure that all of those functions are always called only when we're reloading an incomplete form?

wpride: Yes, that's correct. In Android, this function is only called once when you're opening a new form or re-opening an incomplete form. In Formplayer, we call this function every time you answer a question or interact with the form because we have to serialize and de-serialize and then re-initialize the form each time the server responds. This boolean is a bit of a hack/sneaky way to tell CommCare not to re-initialize triggerables if we're sure the background data hasn't changed - in short, we don't want to call this function for each answer in Formplayer. From the Android perspective I can see this naming doesn't really make sense. I'll update.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1458,38 +1458,51 @@ public void readExternal(DataInputStream dis, PrototypeFactory pf) throws IOExce
...
+     *                                changes to user databases
      */
-    public void initialize(boolean newInstance, boolean isCompletedInstance,
-                           InstanceInitializationFactory factory, String locale, boolean isReadOnly) {
+    public void initialize(boolean newInstance,
+                           boolean isCompletedInstance,
+                           InstanceInitializationFactory factory,
+                           String locale,
+                           boolean isReadOnly,
+                           boolean initializeTriggerables) {
```
amstone326: @wpride Hey so this variable re-naming definitely makes more sense, but it made me realize that there's 1 more thing that should change here. The `isReadOnly` argument was added (by me) fairly recently to serve essentially the same role that `initializeTriggerables` is now playing. You'll notice that the only usage of `isReadOnly` is in that one line below to control initializing triggerables. So it would be cleaner and a more accurate representation of what's going on here if you now removed the `isReadOnly` arg entirely, and then passed in `false` for `initializeTriggerables` in all places where we were previously passing `true` for `isReadOnly`. (If that wasn't totally clear, I can also just do this refactor if you want me to; let me know).

amstone326:
@wpride What's the current status of this? it has some conflicts to be resolved but it's also been around for a while so not sure if it's waiting on something?

wpride:
closing, still working out the kinks on the other PR

PR #662, opened 2017-10-05
2.39.2 Stop-gap measure to prevent crash on relative refs with predicates

amstone326:
Will prevent [this crash](https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59d3bfe3be077a4dccaf5fec?time=last-seven-days), and instead result in the analysis just not returning results and therefore having to inflate the full set of instances. (which is totally fine for now since that's what would have happened before 2.39 anyway, and is preferable to crashing or us trying to make a bigger change in a hotfix).


amstone326:
Release Note: Fixes a bug where a menu load optimization could cause a crash on certain expressions

PR #668, opened 2017-11-01
CommCareConfigEngine in line with Formplayer

wpride:
Some refactors of `CommCareConfigEngine` to unify with how we use this class in Formplayer

1. Pull out version numbers so that we can access statically
2. Allow configuration of `StorageFactory` and `InstallerFactory`, with default constructors
3. Remove `output` param since this was only ever `System.out`


wpride:
Yeah, I think because we've overwritten a lot of the crucial parts in `FormplayerConfigEngine` we can log directly

PR #671, opened 2017-11-06
Modified detail sort order to include cache and index fields at end

shubham1g5:
case: https://manage.dimagi.com/default.asp?262590#1415995

This will make sure that cache and index fields are included at the end whenever we are doing a default sort. In case of all fields marked as cahce and indexed, we don't call sort functions at all, so this change will have no effect in that case. 

Product Note: Currently we don't include sort properties marked as cache and index in HQ while sorting by default, this PR makes a change to include all those properties in default sort with least preference (non cache and index properties will take preference over cache and index properties).

### DIFF
--- a/src/main/java/org/commcare/suite/model/Detail.java
+++ b/src/main/java/org/commcare/suite/model/Detail.java
@@ -305,9 +305,13 @@ public class Detail implements Externalizable {
      */
     public int[] getOrderedFieldIndicesForSorting() {
         Vector<Integer> indices = new Vector<>();
+        Vector<Integer> cacheAndIndexedIndices = new Vector<>();
         outer:
         for (int i = 0; i < fields.length; ++i) {
             int order = fields[i].getSortOrder();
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/suite/model/Detail.java
```diff
@@ -321,15 +325,18 @@ public void writeExternal(DataOutputStream out) throws IOException {
...
-        if (indices.size() == 0) {
-            return new int[]{};
-        } else {
-            int[] ret = new int[indices.size()];
-            for (int i = 0; i < ret.length; ++i) {
-                ret[i] = indices.elementAt(i);
-            }
-            return ret;
+
+        int resultLength = indices.size() + cacheAndIndexedIndices.size();
```
ctsims: If you have a minute, I'd refactor this into a method (mergeIntegerArrays) simply because it would make the code mildly easier to read.

PR #673, opened 2017-11-09
Better exception for null instance in resolveReference()

amstone326:
This is to help get to the bottom of https://fabric.io/dimagi/android/apps/org.commcare.dalvik/issues/59d22cacbe077a4dcc9eaa0e?time=last-seven-days.

Even after trying to reproduce with one of the apps/users that experienced this crash, I couldn't get it to happen. I also can't reason out what could make it such that `qualifiedRef.getInstanceName()` would be null in this method, though it seems that that is what must be happening to cause this crash (I'm pretty sure it's expected that `this.getMainInstance()` would be null here, since we're just in a case list, and thus `qualifiedRef.getInstanceName()` must be the failure point). 

Will or Clayton, if either of you have any insight just from looking at this, obviously let me know. But otherwise the soft assert seemed like the best middle step.

### DIFF
--- a/src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
+++ b/src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
@@ -4,6 +4,7 @@ import org.commcare.cases.query.QueryContext;
 import org.commcare.cases.query.QuerySensitiveTreeElementWrapper;
 import org.commcare.cases.query.queryset.CurrentModelQuerySet;
 import org.commcare.cases.util.QueryUtils;
+import org.commcare.util.LogTypes;
 import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.instance.AbstractTreeElement;
 import org.javarosa.core.model.instance.DataInstance;
@@ -13,6 +14,7 @@ import org.javarosa.core.model.trace.BulkEvaluationTrace;
 import org.javarosa.core.model.trace.EvaluationTrace;
 import org.javarosa.core.model.trace.EvaluationTraceReporter;
 import org.javarosa.core.model.utils.CacheHost;
+import org.javarosa.core.services.Logger;
 import org.javarosa.xpath.IExprDataType;
 import org.javarosa.xpath.XPathLazyNodeset;
 import org.javarosa.xpath.expr.FunctionUtils;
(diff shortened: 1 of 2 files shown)


ctsims:
I think the circumstance you're describing (no main instance available, and no named instance in the ref) would replicate if you made a form display condition (or other out-of-form) reference which was fully qualified but simply had no instance, IE:

Form Display Condition = /data/theresnoinstance = 'false'

I think it would be better to fail explicitly after the soft assert with an XPathException which could be caught and processed by a top level handler

amstone326:
@ctsims Sorry if I'm misunderstanding something, but this error in `resolveReference()` is being triggered exclusively when it is triggered by the call-chain coming from `EntitySelectActivity.onItemClick()`, so the `qualifiedRef` for which there is no reference has to be the ref that comes from the call to `adapter.getItem()` [here](https://github.com/dimagi/commcare-android/blob/f5d37841600809c2c4b414dcdd5c77bbe1abefd4/app/src/org/commcare/activities/EntitySelectActivity.java#L513-L513). So I don't think the example you're describing is a good parallel for what would be going on here?

ctsims:
@amstone326 got it, I didn't actually look super deeply into the source, so I figured it was the less confusing option.

I think that throwing a more concrete exception rather than a soft assert (or along with the soft assert) still makes the most sense. The call is definitely going to fail immediately after, so it would make sense for us to crash with the most context possible.

This is how this is handled in a different context where we are looking for the instance to use to evaluate a reference:

https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xpath/expr/XPathPathExpr.java#L190

which specifically results in an (XPathMissingInstanceException)

https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xpath/expr/XPathPathExpr.java#L201

There are two different ways to get to the null pointer in `resolveReference`

A) The tree reference is against a 'main' instance and one isn't available (that's the example I gave)
B) The tree reference references an instance, but it's not available in the current context (`getInstance()` can return null if the instance isn't found).

It's probably worth distinguishing between those two outcomes in whatever error we throw as well.

amstone326:
@ctsims So I see what you're saying about differentiating between the 2 scenarios, but from a code perspective, I don't know if that's really meaningful, becase the state that leads to both conditions is exactly the same (when `this.getMainInstance()` and `qualifiedRef.getInstanceName()` are _both_ null). I think all you're really talking about is examining the structure of the reference in order to determine whether the instance that _should_ have been non-null was the main instance or its own internally-referenced instance available for use. So I'm including the reference in the exception message so that we can determine this by examination of that. But I'm not seeing how there would be any code path difference between these 2 scenarios.

amstone326:
@ctsims Can you let me know if you agree with what I said above or are still looking for further changes?

PR #679, opened 2017-11-17
Hackathon! Short cycle detection algorithm

wpride:
Vastly improves the error message for cyclical references in forms

cross-request: https://github.com/dimagi/formplayer/pull/547/files

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -33,7 +33,7 @@ import org.javarosa.core.model.util.restorable.RestoreUtils;
 import org.javarosa.core.services.Logger;
 import org.javarosa.core.services.locale.Localizer;
 import org.javarosa.core.services.locale.TableLocaleSource;
-import org.javarosa.core.util.DFS;
+import org.javarosa.core.util.ShortestCycleAlgorithm;
 import org.javarosa.core.util.DataUtil;
 import org.javarosa.core.util.Interner;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
@@ -2895,7 +2895,7 @@ public class XFormParser {
         }
 
         if (!acyclic) {
-            ArrayList<String> cycleString = new DFS(edges).getCycle();
+            ArrayList<String> cycleString = new ShortestCycleAlgorithm(edges).getCycle();
             reporter.error("Logic is cyclical, referencing itself. Shortest Cycle " + cycleString);
             throw new RuntimeException("Logic is cyclical, referencing itself. Shortest Cycle " + cycleString);
         }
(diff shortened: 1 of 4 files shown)


src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2898,14 +2895,9 @@ private void checkDependencyCycles() {
...
-            String errorMessage = "Logic Dependency Cycle:\n";
-            for (int i = 0; i < edges.size(); i++) {
-                TreeReference[] edge = (TreeReference[])edges.elementAt(i);
-                errorMessage += edge[0].toString() + " references " + edge[1].toString() + "\n";
-            }
-            reporter.error(errorMessage);
-
-            throw new RuntimeException("Logic is cyclical, referencing itself.");
+            ArrayList<String> cycleString = new ShortestCycleAlgorithm(edges).getCycle();
+            reporter.error("Logic is cyclical, referencing itself. Shortest Cycle " + cycleString);
```
ctsims: Relying on the toString of ArrayList here doesn't seem safe to me. I think we need a deliberate representation of what this should look like

ctsims: also: this message is duplicated and should be put in an intermediate String and re-used.

ctsims:
@wpride One other thing: The cycle message shows up in two different contexts in the context Ticket from jessica. Does this message replace both of those places?

wpride:
@ctsims thanks, updated

amstone326:
@wpride Why does this have to target the formplayer branch instead of just master?

wpride:
@amstone326 Because Formplayer is the primary mechanism delivering these parse error messages these days and I find it easier to develop on my Formplayer IDE than Android. Will get into `master` after this is merged

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2057,7 +2054,7 @@ private void addMainInstanceToFormDef(Element e, FormInstance instanceModel) {
         //TreeElement root = buildInstanceStructure(e, null);
         loadInstanceData(e, instanceModel.getRoot());
 
-        checkDependencyCycles();
+        //checkDependencyCycles();
```
amstone326: this shouldn't stay commented

amstone326:
@wpride Sorry I'm still confused; what about this PR couldn't be targeted to master as-is?

wpride:
@amstone326 thanks

@ctsims that reminds me - there are two places where we check for cyclical references, in `finalizeTriggerables` in `FormDef` and in `addMainInstanceToFormDef` in `XFormParser`. `finalizeTriggerables` is only claled in two places - once in `XFormParser` immediately after we call `checkDependencyCycles` and again in `readExternal` of `FormDef`. So unless we saved a `FormDef` with cyclical references (which shouldn't be possible?) I'm confused how we could end up in the second call path

wpride:
@amstone326 It could be I just want to finish all of the changes/iteration in one PR rather than two

dannyroberts:
@wpride @ctsims Is this good to go?

wpride:
This is gtg from my perspective - @ctsims

wpride:
Merging since this is already merged in `master`

PR #687, opened 2017-11-30
Add comment for why id is sometimes null

amstone326:


ctsims:
@amstone326 as a heads up you can also self merge commit-only PR's.

PR #689, opened 2017-11-30
Remove 'abandonable' concept in favor of simple thread interruption

ctsims:
Also introduces a useful 'sleep' method to xpath which majorly simplifies testing handling behavior around long running ops.

Major reduction in the tech debt of these systems.

@wpride Note - this will require a formplayer change.


src/main/java/org/commcare/suite/model/Text.java
```diff
@@ -349,19 +349,22 @@ public void ensureCacheIsParsed() throws XPathSyntaxException {
...
-                return evaluate(abandonableContext[0]);
+                toCancel[0] = Thread.currentThread();
+                return evaluate(abandonableContext);
             }
         }).doOnDispose(new io.reactivex.functions.Action() {
             @Override
             public void run() throws Exception {
-                abandonableContext[0].signalAbandoned();
-                abandonableContext[0] = null;
+                if(toCancel[0] != null) {
```
shubham1g5: @ctsims  The new pattern looks much simpler now 👍  One question, have you been able to test this without calling interrupt explicitly in `doOnDispose()` . I do believe that Rx framework should be interrupting the thread whenever it makes a call to dispose.

ctsims: Hey Shubham!

This is a great question. I meant to make a note that I changed this in the most conservative way I could without validating whether the interrupts happen by default.

I can look into that. I agree that it would be _awesome_ to not need to manually trigger the interrupt.

PR #691, opened 2017-12-05
Throws error in case of space in select choices

shubham1g5:
Case: https://manage.dimagi.com/default.asp?266292#1427432

Product Note: Throw an error like 'Select answer option cannot contain spaces in question 'QuestionText' with option 'ChoiceWithSpaceText' ' in case spaces are present in lookup table options used for a lookup table select question's value field.

### DIFF
--- a/src/main/java/org/javarosa/form/api/FormEntryPrompt.java
+++ b/src/main/java/org/javarosa/form/api/FormEntryPrompt.java
@@ -203,16 +203,25 @@ public class FormEntryPrompt extends FormEntryCaption {
     public Vector<SelectChoice> getSelectChoices(boolean shouldAttemptDynamicPopulation) {
         QuestionDef q = getQuestion();
         ItemsetBinding itemset = q.getDynamicChoices();
+        Vector<SelectChoice> selectChoices;
         if (itemset != null) {
             if (populatedDynamicChoices == null && shouldAttemptDynamicPopulation) {
                 form.populateDynamicChoices(itemset, mTreeElement.getRef());
                 populatedDynamicChoices = itemset.getChoices();
             }
-            return populatedDynamicChoices;
+            selectChoices = populatedDynamicChoices;
         } else {
             // static choices
-            return q.getChoices();
+            selectChoices = q.getChoices();
         }
+        if (selectChoices != null) {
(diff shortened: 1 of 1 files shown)


amstone326:
@shubham1g5 Along the lines of what Will said, this should be tagged as product/all-users, not invisible

sheelio:
Hmm, I'm nervous about projects being auto upgraded to the next version of CommCare and then their application behaviour breaking.  What's the error that users will see?

shubham1g5:
@sheelio User will see this - Select answer option cannot contain spaces in question 'QuestionText' with option 'ChoiceWithSpaceText'

sheelio:
This doesn't seem like a change that is safe to release without providing some opt-out for users (since they will have applications that are built with spaces in the choices and their projects would auto-upgrade to the next version of CommCare).  I know that users also include spaces in Lookup Table question values as well.

shubham1g5:
@wpride Now I am curious what problems does having a space causes since it seems to be working fine with some existing projects.

wpride:
@shubham1g5 You could never re-use this value - for example, it would fail if you saved this value to a case property and then tried to re-load it in a follow-up form; or (as we first experienced this) if you tried to edit the form. Its also just a violation of the XPath spec.

shubham1g5:
@sheelio Alright. We can hold on to this for now and give users  a warning for Commcare 2.43 when we will stop allowing spacey lookups in select questions. Also even when we finally deploy this change we can implement a custom setting in Dev preferences say ALLOW_LOOKUP_SELECT_WITH_SPACE that defaults to no and can be changed to yes from HQ.

sheelio:
@shubham1g5 
I'll check with Clayton about this.  My recommendation would be:
- We notify all projects that we're making this change
- We default new apps so that they error if there have spaces
- For old apps, we allow them to continue but add some logging so we can identify which projects are affected by this
- After X time, we switch so that all apps have the error

ctsims:
Hi @sheelio 

Happy to implement this following this process: https://confluence.dimagi.com/pages/viewpage.action?pageId=41489640

sheelio:
@ctsims Thanks - I think that makes the most sense.

shubham1g5:
Closing this in favour of https://github.com/dimagi/commcare-android/pull/1921

PR #692, opened 2017-12-05
Demonstrating issue with today()

amstone326:
@wpride @shubham1g5 So I have most of an explanation for https://manage.dimagi.com/default.asp?266188, but with a small question mark, so want to run it by you guys and see what you think. 

The "hole" I identified is this: `XPathTodayFunc` computes its result by doing `DateUtils.roundDate(new Date())`, and `roundDate()` in turn calls `getFields()` with an optional timezone argument. If no timezone is passed, then `getFields()` uses the default timezone from the host. So if you look at the test I added to `DateDataTests`, the behavior that REC is seeing is replicable if you create a `Date` that is just past midnight in a timezone that is "earlier"/more-east than the current default, and then call `roundDate()` on it _without_ specifying a timezone. In the test I've written, `roundDate()` returns the day before the original date due to the timezone offset between EST and GMT being greater than the time between the original `Date` object's time and midnight. 

The inconsistency with the value of `now()` is explained by the fact that the implementation of `XPathNowFunc` simply returns `new Date()`, so it isn't working with timezones at all (java.util.Dates do not have any timezone associated with them). So the timezone that is showing up on HQ for the `timeStart` value must be getting inferred by HQ in some way.

The reason I put "hole" in quotes above is that the only way this could cause the bug that REC is seeing is if the devices in question are set to the incorrect timezone. I believe this is possible, but I don't think there's anything we can do about it from a code perspective? I don't know of any way to be smarter about getting the right timezone other than using the default; do you guys? If not, then I think we just need to tell them to check the device timezones and make sure they're set properly.

### DIFF
--- a/src/main/java/org/javarosa/core/model/utils/DateUtils.java
+++ b/src/main/java/org/javarosa/core/model/utils/DateUtils.java
@@ -113,15 +113,15 @@ public class DateUtils {
         return nonGregorian;
     }
 
-    public static DateFields getFields(Date d) {
+    public static DateFields getFieldsInDefaultTimezone(Date d) {
         return getFields(d, null);
     }
 
-    public static DateFields getFields(Date d, String timezone) {
+    public static DateFields getFields(Date d, TimeZone timezone) {
         Calendar cd = Calendar.getInstance();
         cd.setTime(d);
         if (timezone != null) {
-            cd.setTimeZone(TimeZone.getTimeZone(timezone));
+            cd.setTimeZone(timezone);
         }
 
         DateFields fields = new DateFields();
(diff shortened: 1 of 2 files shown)


wpride:
Very nice catch. Two thoughts when you get back:

1) pertaining to this specific ticket, the `timeEnd` is reported as `<n10:timeStart>2017-08-25T07:51:11.157Z</n10:timeStart>` - I believe this value usually contains the device's time zone which appears to be set correctly. So I'm not sure that this bug would explain the ticket?
2) Seems like we should align the behaviors of `today()` and `now()` to both or neither use the time zone

amstone326:
Closing this because Will and I determined that it's a red herring

PR #697, opened 2017-12-08
Merge Formplayer CLI code into master

wpride:
Moves over most of the CLI-level updates made for Formplayer. This mostly involves loading up a richer set of data than the CLI was previously, making this accessible via getters (rather than solely through the `prompt` call), and more lazily loading some information that might not be necessary for each screen usage.

This arguably loads more information for the CLI than would ever be necessary, but I don't see that as a huge problem (though open to being pushed back on)

### DIFF
--- a/src/main/java/org/javarosa/core/util/DataUtil.java
+++ b/src/main/java/org/javarosa/core/util/DataUtil.java
@@ -67,6 +67,15 @@ public class DataUtil {
         return s.split("[ ]+");
     }
 
+    public static boolean intArrayContains(int[] source, int target) {
+        for (int sort: source) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/javarosa/core/util/DataUtil.java
```diff
@@ -67,6 +67,15 @@ public static String listToString(List<String> list) {
         return s.split("[ ]+");
     }
 
+    public static boolean intArrayContains(int[] source, int target) {
+        for (int sort: source) {
```
amstone326: This variable should be called something more generic instead of `sort` since it could be any int

PR #698, opened 2017-12-08
Case Claim for CLI

wpride:
Implements Case Claim for the CLI

Major entry in the annals of not helpful dev tooling improvements, but good for Formplayer/core merging and parity. Not a ton of shared code between Formplayer and CLI here due to the platform specific network libraries but pretty good. Slightly awkward fit for the CLI paradigm since the ~~SyncScreen doesn't accept any input but functional.~~

Sync screen now gives prompts:

```
0)Batman
> 0
Syncing with url https://www.commcarehq.org/a/chadbrochill/phone/claim-case/ and parameters {case_id=cbbb7605-6f12-4718-9b39-f6c8407e4dc3}
Sync successful with response Response{protocol=http/1.1, code=200, message=OK, url=https://www.commcarehq.org/a/chadbrochill/phone/claim-case/}
Press 'enter' to continue.
```


build.gradle
```diff
@@ -39,8 +39,6 @@ sourceSets {
   }
     api {
         java {
-            srcDir 'src/cli/java/org/commcare/util/screen'
```
ctsims: Is this the result of a merge? I'm not sure I understand why removing these woudln't cause issues compiling the CLI jar

src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -84,7 +88,10 @@ public ApplicationHost(CommCareConfigEngine engine, PrototypeFactory prototypeFa
     }
 
     public void setRestoreToRemoteUser(String username, String password) {
-        this.mLocalUserCredentials = new String[]{username, password};
+        this.username = username;
+        this.password = password;
+        String domain = PropertyManager.instance().getSingularProperty("cc_user_domain");
```
ctsims: I really thought there was a more centralized location to do this addition in the code these days, is that not true? (I might have also only done that with a cli change that isn't merged yet)

src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -323,14 +330,24 @@ private void printErrorAndContinue(String error, Exception e) {
...
             //XFORM TIME!
             return null;
         } else if (next.equals(SessionFrame.STATE_COMMAND_ID)) {
             return new MenuScreen();
         } else if (next.equals(SessionFrame.STATE_DATUM_VAL)) {
             return new EntityScreen();
+        } else if (next.equals(SessionFrame.STATE_QUERY_REQUEST)) {
+            return new QueryScreen(qualifiedUsername, password, System.out);
+        } else if (next.equals(SessionFrame.STATE_SYNC_REQUEST)) {
+            new SyncScreen(qualifiedUsername, password);
```
ctsims: So the sync screen on the CLI has no user facing output? feels like it would be worth having it display something to keep it in parity with what webapps does, even if its synchronous.

IE: SyncScreen could show details about the connection, the # of results returned, any error that came up, etc.

currently I think it's the only screen that does what this does, which is process something raw, so I'd want to change what we call it if we don't

src/test/java/org/commcare/test/utilities/MockApp.java
```diff
@@ -40,13 +40,6 @@ public MockApp(String resourcePath) throws Exception {
         final LivePrototypeFactory mPrototypeFactory = setupStaticStorage();
         MockUserDataSandbox mSandbox = new MockUserDataSandbox(mPrototypeFactory);
         CommCareConfigEngine mEngine = new CommCareConfigEngine(mPrototypeFactory);
-        
-        mEngine.setStorageFactory(new IStorageIndexedFactory() {
```
ctsims: I think this is needed for tests to run properly, or is this also just the result of merging? i'm not sure I see from this PR anything that would prevent this from being needed

src/cli/java/org/commcare/util/cli/ApplicationHost.java
wpride: This is centralized as of https://github.com/dimagi/commcare-core/pull/718, updating

build.gradle
wpride: Hm yes looks like a merge issue, sorry about that. I merged the latest and this resolved itself.

wpride:
@ctsims Thanks for the feedback, updated

wpride:
@ctsims bump

src/cli/java/org/commcare/util/screen/SyncScreen.java
```diff
@@ -0,0 +1,120 @@
...
+                .url(url)
+                .method("POST", RequestBody.create(null, new byte[0]))
+                .header("Authorization", credential)
+                .post(postBody)
+                .build();
+        try {
+            OkHttpClient client = new OkHttpClient();
+            return client.newCall(request).execute();
+        } catch (IOException e) {
+            throw new CommCareSessionException("Exception while making sync request", e);
```
ctsims: So in theory this IOException could easily represent a number of different outcomes, right? A session exception's going to kill the session entirely. I'd imagine that it would make more sense for this to be a retry-able event which clearly expresses the error (in case it's an error that would appear on the phone and they are debugging)

wpride: Good call, thanks!

src/cli/java/org/commcare/util/screen/SyncScreen.java
```diff
@@ -0,0 +1,133 @@
...
+                printStream.println("Press 'enter' to retry.");
+                return;
+            }
+            syncSuccessful = true;
+            ApplicationHost.restoreUserToSandbox(sessionWrapper.getSandbox(), sessionWrapper, sessionWrapper.getPlatform(), username, password);
+            printStream.println(String.format("Sync successful with response %s", response));
+            printStream.println("Press 'enter' to continue.");
+        } else {
+            // expected a sync entry; clear session and show vague 'session error' message to user
+            throw new CommCareSessionException("Initialized sync request while not on sync screen");
```
ctsims: It looks like issues with the I/O process around the response are now not re-tryable again and kill your session in the event of network issues. 

Can your revisit making the I/O part of this not a binary process?

Also: The primary purpose of the CLI is Debugging, so swallowing issues with vague errors should be our lazt line of defense. We should be as intentional as possible here to fail at the right step to retain the roper context. It seems after your last commit we're actually less likely to output a useful error message than we were before...

wpride: @ctsims I wrote this to make it clearer but I believe that retrying for `IOException`s worked previously since `makeRequest` would return null. I've confirmed that retrying works by manually throwing an `IOException` from the method as well as turning off my internet. 

Good call on the error printing

src/cli/java/org/commcare/util/screen/SyncScreen.java
```diff
@@ -0,0 +1,140 @@
...
+                    sessionWrapper.getPlatform(),
+                    username,
+                    password);
+
+            printStream.println(String.format("Sync successful with response %s", response));
+            printStream.println("Press 'enter' to continue.");
+        } catch (IOException e) {
+            e.printStackTrace();
+            printStream.println(String.format("Sync failed with exception %s", e.getMessage()));
+            printStream.println("Press 'enter' to retry.");
```
ctsims: Hm, just realized this will leave the user in an infinite loop. Is it possible for us to take in an input which decides whether to retry? (It's possible that the Command infrastructure, ie - ":home" or whatever) allows this, already?

Sorry for all of the back and forth on this PR,I'm just thinking of mistakes I made in the CLI which ended up frustrating me a lot later and causing a lot of churn when I'd get deep into debugging something and the whole session would just die or become unsalvagable

src/cli/java/org/commcare/util/screen/SessionUtils.java
```diff
@@ -0,0 +1,113 @@
...
+        });
+
+        //Go get our sandbox!
+        try {
+            System.out.println("GET: " + otaSyncUrl);
+            URL url = new URL(otaSyncUrl);
+            HttpURLConnection conn = (HttpURLConnection)url.openConnection();
+
+            if (conn.getResponseCode() == 412) {
+                System.out.println("Server Response 412 - The user sandbox is not consistent with " +
```
ctsims: should these sysouts get migrated to printStream.outs?

src/cli/java/org/commcare/util/screen/SyncScreen.java
wpride: Sorry, I'm not seeing the infinite loop - where are you referring to? If I manually throw an IOException from `makeSyncRequest` then I am prompted to retry and can type `:home` in order to exit if need be.

src/cli/java/org/commcare/util/screen/SessionUtils.java
wpride: :+1: thanks

PR #702, opened 2017-12-11
Updates to ArchiveFileRoot

wpride:
Changes to ArchiveFileRoot used by Formplayer


wpride:
@ctsims No issue, but Formplayer relies on these changes to `ArchiveFileRoot` in order to cache the CCZ file per-application (so that you don't have to download the CCZ for each user of an app)

PR #710, opened 2018-01-05
Remove warning if choice length too long

wpride:
fixes: https://manage.dimagi.com/default.asp?267666

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -1324,11 +1324,6 @@ public class XFormParser {
                 }
 
                 if (value != null) {
-                    if (value.length() > MAX_VALUE_LEN) {
-                        reporter.warning(XFormParserReporter.TYPE_ERROR_PRONE,
-                                "choice value [" + value + "] is too long; max. suggested length " + MAX_VALUE_LEN + " chars",
-                                getVagueLocation(child));
-                    }
 
                     //validate
                     for (int k = 0; k < value.length(); k++) {
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -1279,8 +1279,6 @@ public class XFormParser {
     }
 
     private void parseItem(QuestionDef q, Element e) {
-        final int MAX_VALUE_LEN = 32;
-
         //catalogue of used attributes in this method/element
         Vector<String> usedAtts = new Vector<>();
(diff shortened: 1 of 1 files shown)


src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1324,11 +1324,6 @@ private void parseItem(QuestionDef q, Element e) {
                 }
 
                 if (value != null) {
-                    if (value.length() > MAX_VALUE_LEN) {
```
orangejenny: Can't you also remove the constant's declaration?

wpride: Jenny, you know so much stuff

ctsims:
@wpride @orangejenny 

Hey guys. Just had a quick thought while looking at something:

Is there a specific reason we shouldn't just make the constraint _larger_ rather than removing it entirely?

In general we get ourselves in to trouble way more often by being arbitrarily lenient rather than arbitrarily constrained, which is something we've discussed trying to be much better at in the future.

I'd be pro moving the constant length to 256 just so there's at least some "sanity check" on the boundaries of the data here. Otherwise I'm a bit concerned we will have someone do something nuts and then be confused when they weren't even warned that it was crazy.

orangejenny:
@ctsims 

Point about constraints being good makes sense, though if we were to add a higher limit, perhaps it should be a fatal error...otherwise we wouldn't be warning people they're doing something crazy unless they *also* have a fatal error in their app. The whole idea that we show warnings iff there's also an error is confusing in general.

Given the age of the code I don't feel strongly either way about having some sort of sanity check, will leave to Will to decide.

wpride:
@ctsims @orangejenny I agree with Jenny - if we think that a certain character limit will crash the application then we should change this to a failure. Currently the user will never be exposed to this warning unless there's an additional hard failure - and, confusingly, this warning will show along with the hard failure and you could fix the warnings without fixing the failure.

ctsims:
The warning/error situation is irrelevant to this specific warning, fwiw.
There are a number of warnings that will have the same issue (only show up
if there's a hard failure), and if we want to fix that we should fix that
systemically.

There's no specific wisdom behind how we do one v. the other, but I think
of warnings as being "If you're seeing an unexpected error/issue somewhere,
warnings provide you with potential leads." I agree in this case that we
don't really understand where this would cause issues, so I could see it
being a hard limit. that would require us to know whether anyone's current
violating that limit, though, so we don't tank any real apps in the field
right now.

-Clayton

wpride:
Closing in favor of https://github.com/dimagi/commcare-core/pull/712

PR #714, opened 2018-01-09
Wsp/app fixture fix

wpride:
cross-request: https://github.com/dimagi/formplayer/pull/583

### DIFF
--- a/src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
+++ b/src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
@@ -138,7 +138,13 @@ public class CommCareInstanceInitializer extends InstanceInitializationFactory {
 
             TreeElement root = fixtureBases.retrieve(key);
             if (root == null) {
-                FormInstance fixture = SandboxUtils.loadFixture(mSandbox, refId, userId);
+                FormInstance fixture = SandboxUtils.loadFixture(mSandbox,
+                        refId,
+                        userId);
+
+                if (fixture == null) {
+                    fixture = SandboxUtils.loadAppFixture(mPlatform.getFixtureStorage(), refId, userId);
+                }
 
                 if (fixture == null) {
                     throw new FixtureInitializationException(reference);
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/core/sandbox/SandboxUtils.java
```diff
@@ -65,9 +68,16 @@ public static FormInstance loadFixture(UserSandbox sandbox,
             }
         }
 
+        if (appFixtureStorage != null) {
+            return loadAppFixture(appFixtureStorage, refId, userId);
+        }
         return loadAppFixture(sandbox, refId, userId);
     }
 
+    public static FormInstance loadFixture(UserSandbox sandbox, String refId, String userId) {
```
amstone326: Why do the other places that call this method directly not also need to try to retrieve the app-level fixture storage first?

src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
```diff
@@ -138,7 +139,15 @@ protected TreeElement loadFixtureRoot(ExternalDataInstance instance,
 
             TreeElement root = fixtureBases.retrieve(key);
             if (root == null) {
-                FormInstance fixture = SandboxUtils.loadFixture(mSandbox, refId, userId);
+                IStorageUtilityIndexed<FormInstance> fixtureStorage = null;
+                if (mPlatform != null) {
+                    fixtureStorage = mPlatform.getFixtureStorage();
```
amstone326: If I'm understanding this change properly, this means that if there are app fixtures in both user storage and app storage, we default to the ones in app storage. Is that intended? If so that's fine, I just don't know which is correct or if there is a clear correct answer for all cases.

src/main/java/org/commcare/core/sandbox/SandboxUtils.java
wpride: Hmm, both of those usages have the comment `//TODO: Wow. This is.... kind of megasketch` so I'm guessing they're pretty sketch

src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
wpride: Updated to attempt both (falling back to user storage if it isn't in app storage)

PR #715, opened 2018-01-09
Use timezone offset in DateUtils calcs when present

amstone326:
cross-request: https://github.com/dimagi/formplayer/pull/585  
 
This PR creates the concept of a `TimezoneProvider` that all `DateUtils` operations consult for a potential `timezoneOffset` to use when parsing or serializing dates, as an alternative to using the default timezone of the current environment. This is so that formplayer can can create an implementation of `TimezoneProvider` that passes the timezone offset of the user's browser, so that all date calculations are done relative to that offset, rather than the timezone that the formplayer server lives in.


src/main/java/org/javarosa/core/model/utils/DateUtils.java
```diff
@@ -122,18 +130,30 @@ public static DateFields getFields(Date d, String timezone) {
         cd.setTime(d);
         if (timezone != null) {
             cd.setTimeZone(TimeZone.getTimeZone(timezone));
+        } else if (timezoneOffset() != -1) {
```
wpride: Ah I see, we check for `-1` here

src/main/java/org/javarosa/core/model/utils/DateUtils.java
```diff
@@ -440,13 +484,29 @@ private static boolean parseDateAndStore(String dateStr, DateFields df) {
...
+
         DateFields fields = new DateFields();
         if (!parseTimeAndStore(str, fields)) {
             return null;
         }
         return getDate(fields);
     }
 
+    private static String getOffsetInStandardFormat(int offsetInMillis) {
+        int hours = offsetInMillis / 1000 / 60 / 60;
```
wpride: How does this work for (ridiculous) timezones with `30` or `45` minute offsets form UTC?

amstone326: Ah wow good catch, will fix

src/main/java/org/javarosa/core/model/utils/DateUtils.java
amstone326: This sounds like there was a comment you made before it that didn't go through? Did you have a concern here?

wpride: Was in reference to https://github.com/dimagi/formplayer/pull/585/files#r160769302

PR #716, opened 2018-01-09
Caching of expressions in forms

amstone326:
cross-request: https://github.com/dimagi/commcare-android/pull/1931

Also includes a lot of changes/additions to the `EvaluationTrace` infrastructure, in order to support examining how the expression caching is working.


amstone326:
@ctsims This shouldn't be merged yet since I still want to do all of the further testing we discussed, but it is at a good point for you to do a first review pass for overall architecture and implementation.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -178,6 +178,18 @@ public TreeReference getOriginalContext() {
...
 
+    public void enableExpressionCaching() {
+        this.expressionCacher = new InFormExpressionCacher();
+    }
+
+    public boolean expressionCachingEnabled() {
+        return expressionCacher != null;
+    }
+
+    public InFormExpressionCacher expressionCacher() {
```
ctsims: Super nitpicky, but "InForm" feels like an awkward thing for this to be based on v. maybe "MainInstance" (IE: Anything rooted at '/'), just to keep the vocabulary within what is generally referred to in the EvaluationContext

src/main/java/org/javarosa/xpath/analysis/ReferencesMainInstanceAnalyzer.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public ReferencesMainInstanceAnalyzer(String instanceName) {
+        super();
+        this.mainInstanceRoot = instanceName;
+    }
+
+    @Override
+    public void doNormalTreeRefAnalysis(TreeReference treeRef) throws AnalysisInvalidException {
+        if (treeRef.getName(0).equals(mainInstanceRoot)) {
```
ctsims: The idea here is that we can provide different 'main' instances?

src/main/java/org/javarosa/xpath/analysis/XPathAnalyzer.java
```diff
@@ -51,17 +53,13 @@ public TreeReference getOriginalContextRef() {
 
     protected void requireOriginalContext(TreeReference forReference) throws AnalysisInvalidException{
         if (getOriginalContextRef() == null) {
-            throw new AnalysisInvalidException("No original context ref was available when " +
```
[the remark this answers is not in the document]
amstone326: @ctsims I agree, but the reason I made this change was that a while back when I was profiling this I discovered that the creation of the exception objects was actually eating into the performance of the analysis by a noticeable amount. So saving that time seemed more important that preserving the exception messages

src/main/java/org/javarosa/xpath/analysis/ReferencesMainInstanceAnalyzer.java
amstone326: That wasn't exactly what I was thinking, no. I just didn't want to hardcode "data" as the root ref name since that is technically not always the case, even within a form

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
amstone326: Are you referring just to what `InFormExpressionCacher` is called?

ctsims: Yeah. It's just a bit of an odd vocabulary coupling since the EC isn't inherently part of the form.

src/main/java/org/javarosa/xpath/analysis/ReferencesMainInstanceAnalyzer.java
ctsims: Ah. So I think we dont' need to check the name of the root node directly for these checks. 

I think you can detect this purely by comparing 
```
            t.getContextType() == TreeReference.CONTEXT_ABSOLUTE && t.getInstanceName() == null;
```

for checking "Does this reference the "main" instance (IE: `/`)".

In theory this actually might mis-detect the main instance because
```
instance('external')/data/path/to/node

and

/data/path/to/node
```

have the same `0` name at the treeref level.

src/main/java/org/javarosa/xpath/analysis/XPathAnalyzer.java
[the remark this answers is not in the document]
ctsims: Aha, good catch. 

Seems like a bit of a shame to lose the capacity all together. It would be possible to either pre-package static, canned error messages (like "no context ref was available why trying to analyse a relative reference") or to do a debug dispatch like

```
public static final bool ANALYSIS_DEBUG = false;

public static AnalysisInvalidException ProduceInvalidException(String message) {
    if(ANALYSIS_DEBUG) { return new AnalysisInvalidException(message);}
    else { return STATIC_INSTANCE; }
}
```

so that in tests or in debugging sessions people can dig into things a bit more, but agree that the performance considerations should be weighed ahead of that value in terms of priority.

src/main/java/org/javarosa/xpath/analysis/TopLevelContextTypesAnalyzer.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    public boolean shouldIncludePredicates() {
+        return false;
+    }
+
+    @Override
+    public boolean shortCircuit() {
+        // there are only 4 possibilities, so if we've gotten them all then no need to keep going
+        return size() == 4;
```
ctsims: For safety and future-proofing it might be worth including the set of potential answers as a list or enumeration in this class (or elsewhere) and using a count of that set. This seems like invisible/loose coupling to a concept that won't necessarily be clear when updated elsewhere.

src/main/java/org/javarosa/xpath/analysis/XPathAccumulatingAnalyzer.java
```diff
@@ -24,41 +25,58 @@
  */
 public abstract class XPathAccumulatingAnalyzer<T> extends XPathAnalyzer {
 
-    private List<T> accumulatedList = new ArrayList<>();
+    private Collection<T> accumulated;
```
ctsims: I think we rely on the order of this accumulated list to make sure that results are ordered reasonably in traces (IE: results are returned in document / evaluation order).

There are hashsets which retain order as well as having set characteristics, I think it'd be worth mandating by type that this is ordered so that we don't lose that value

src/main/java/org/javarosa/xpath/expr/ExpressionCacheKey.java
```diff
@@ -10,22 +16,64 @@
...
+    // These are each only relevant to caching in some cases, and will be null in a cache key for
+    // which they are not relevant
+    @Nullable
+    private TreeReference contextRef;
+    @Nullable
+    private TreeReference originalContextRef;
+
+    public ExpressionCacheKey(InFormCacheableExpr expr, EvaluationContext ec) {
         this.expr = expr;
+        for (int contextType : getRelevantContextTypes(expr)) {
```
ctsims: It seems a bit tenuous to me that there's no way to make sure that this evaluation (getRelevantContextTypes) is being cached (along the lifecycle of the expr object rather than the lifecycle of this key object) . Is it just too fast to matter?

src/main/java/org/javarosa/xpath/analysis/ReferencesMainInstanceAnalyzer.java
```diff
@@ -11,21 +11,19 @@
...
-    public ReferencesMainInstanceAnalyzer(String instanceName) {
+    public ReferencesMainInstanceAnalyzer() {
         super();
-        this.mainInstanceRoot = instanceName;
     }
 
     @Override
     public void doNormalTreeRefAnalysis(TreeReference treeRef) throws AnalysisInvalidException {
-        if (treeRef.getName(0).equals(mainInstanceRoot)) {
+        if (treeRef.getContextType() == TreeReference.CONTEXT_ABSOLUTE &&
```
ctsims: Small style point that this being in a method called 

"doesTreeReferenceReferToMainInstance()"

or something would be a good point of self-documentation for what this check means

src/main/java/org/javarosa/xpath/expr/XPathEqExpr.java
```diff
@@ -50,15 +50,21 @@ public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOExcep
             op = XPathEqExpr.NEQ;
         }
         computedCacheability = ExtUtil.readBool(in);
-        isCacheable = ExtUtil.readBool(in);
+        exprIsCacheable = ExtUtil.readBool(in);
```
ctsims: It feels like there's a lot of context  and state that's being managed here that  would be much cleaner to delegate to a composed cacheState object (which would also make it easier to change this later more safely without requiring safely updating 5-6 implementaitons

src/main/java/org/javarosa/xpath/expr/XPathBinaryOpExpr.java
```diff
@@ -48,6 +48,11 @@ public int hashCode() {
     public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
         op = ExtUtil.readInt(in);
         readExpressions(in, pf);
+        computedCacheability = ExtUtil.readBool(in);
```
ctsims: these are repeated frequently, which is pretty dangerous for making these updates later. It seems like storing these flags (and the relevant serialization code) in a composed member object would be safer and save "do not repeat effort"

src/main/java/org/javarosa/xpath/analysis/TopLevelContextTypesAnalyzer.java
amstone326: yea good call

src/main/java/org/javarosa/xpath/analysis/XPathAccumulatingAnalyzer.java
amstone326: That hasn't actually changed from the way I had it before; not sure if you realized that and were just making the comment anyway? before this refactor, this was still getting implemented as an unordered set the majority of the time.

src/main/java/org/javarosa/xpath/expr/ExpressionCacheKey.java
amstone326: The final version of this has changed, and it is cached there.

src/main/java/org/javarosa/xpath/analysis/ReferencesMainInstanceAnalyzer.java
amstone326: I think the name of the class basically achieves that already, but fine with adding this as well

src/main/java/org/javarosa/xpath/expr/XPathEqExpr.java
amstone326: yea that's a really good point, sorry I should have thought of that

src/main/java/org/javarosa/xpath/expr/XPathBinaryOpExpr.java
amstone326: @ctsims Do you mean something different here than by the similar comment right above?

amstone326:
@ctsims Doc on thoughts for future is here: https://docs.google.com/document/d/15GchwSjbvuvgZ9jPx1GCgQGcHs1y5GuXfO5ZDOJYkCA/edit?usp=sharing

PR #718, opened 2018-01-10
Refactor PropertyManager

wpride:
cross-request: https://github.com/dimagi/commcare-android/pull/1928

Remove the static singleton PropertyManager in favor of using `CommCarePlatform` as the vector for passing this manager. Remove check for `logenabled` property at logger level, in favor of checking this at the `CommCareAppliation` level (where we have access to the `HiddenPreferences` for this value)

### DIFF
--- a/src/main/java/org/commcare/resources/model/ResourceInstaller.java
+++ b/src/main/java/org/commcare/resources/model/ResourceInstaller.java
@@ -102,7 +102,7 @@ public interface ResourceInstaller<T extends CommCarePlatform> extends Externali
      * @return True if the upgrade step was completed successfully.
      * @throws UnresolvedResourceException If the local resource definition could not be found
      */
-    boolean upgrade(Resource r) throws UnresolvedResourceException;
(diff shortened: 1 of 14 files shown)


src/main/java/org/commcare/resources/model/ResourceInstaller.java
```diff
@@ -102,7 +102,7 @@ boolean install(Resource r, ResourceLocation location,
      * @return True if the upgrade step was completed successfully.
      * @throws UnresolvedResourceException If the local resource definition could not be found
      */
-    boolean upgrade(Resource r) throws UnresolvedResourceException;
+    boolean upgrade(Resource r, T instance) throws UnresolvedResourceException;
```
amstone326: Can you rename all the places where you call this variable `instance` to `platform` instead? `instance` is so generic (could be an instance of anything)

wpride: I knew you'd say that :|

PR #725, opened 2018-01-10
Update graphing files with changes from Formplayer

wpride:


amstone326:
@wpride I'm confused what the changes to `XYSeries.java` and `Configurable.java` are here

wpride:
@amstone326 I was too, I think its that the file contents haven't changed but the revisions have (likely because I changed and then un-changed these files in `formplayer`)

PR #726, opened 2018-01-11
Add EntityScreen getReferenceMap() helper

wpride:
Needed by Formplayer


amstone326:
@wpride I should have said this sooner in this process, but when you add back methods like this that are going to register in Android Studio as unused, it would probably be a good idea to add a comment that they're used by formplayer

PR #727, opened 2018-01-12
Formplayer property manager

wpride:
cross-request: https://github.com/dimagi/formplayer/pull/589

Re-implementation of https://github.com/dimagi/commcare-core/pull/718 for `formplayer` branch


src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1276,7 +1276,7 @@ private String parseOutput(Element e) {
     }
 
     private void parseItem(QuestionDef q, Element e) {
-        final int MAX_VALUE_LEN = 32;
+        final int MAX_VALUE_LEN = 256;
```
amstone326: I don't think this should be in this PR

wpride: :+1:

PR #733, opened 2018-01-18
Improve cyclic error message

wpride:
Fixes https://manage.dimagi.com/default.asp?267666#

If you have a reference structure where:

**/data/outer_question**  depends on **/data/group/inner_question**

**and /data/group** depends on **/data/outer_question**

Then you have a cyclical reference error because **/data/group/inner_question** implicitly depends on **/data/group** - if the parent is not relevant, the child is not either.

Our vertices detection algorithm was missing this because when you looked for triggerables dependent on **/data/group** you would find nothing (because there were none directly referencing this node) causing an NPE. This remedies that by including all child references when looking for dependent triggerables of a `TreeReference`.

### DIFF
--- a/src/test/java/org/javarosa/core/model/test/CyclicReferenceTests.java
+++ b/src/test/java/org/javarosa/core/model/test/CyclicReferenceTests.java
@@ -22,7 +22,9 @@ public class CyclicReferenceTests {
             new FormParseInit("/xform_tests/group_cyclic_reference.xml");
         } catch (XFormParseException e) {
             String detailMessage = e.getMessage();
+            // Assert that we're using the shortest cycle algorithm
             assert detailMessage.contains("Shortest Cycle");
+            // There should only be three newlines since only the three core cyclic references were included
             int newlineCount = detailMessage.length() - detailMessage.replace("\n", "").length();
             assert newlineCount == 3;
             return;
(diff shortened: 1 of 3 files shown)


amstone326:
@wpride Are there are any unit tests for the Shortest Cycles Algorithm, or would that be hugely difficult? I'm just concerned that from looking at this it vaguely makes sense to me, but realistically I have no idea if it's right

wpride:
@amstone326 Good call! In adding a unit test I caught a bug. Lesson learned (for a minute...)

src/test/java/org/javarosa/core/model/test/CyclicReferenceTests.java
```diff
@@ -0,0 +1,34 @@
...
+    public void testCyclicReferenceWithGroup() {
+        try {
+            new FormParseInit("/xform_tests/group_cyclic_reference.xml");
+        } catch (XFormParseException e) {
+            String detailMessage = e.getMessage();
+            // Assert that we're using the shortest cycle algorithm
+            assert detailMessage.contains("Shortest Cycle");
+            // There should only be three newlines since only the three core cyclic references were included
+            int newlineCount = detailMessage.length() - detailMessage.replace("\n", "").length();
+            assert newlineCount == 3;
```
amstone326: I've never seen us use this syntax in tests before; does it result in the same sort of test failure and error message as `Assert.assertEquals(expected, actual)`?

wpride: Hmm I hadn't thought about - I think `assert` is Java keyword rather than `Assert.assert*` which is JUnit specific. Not sure if they're behaviorally different, but changing to JUnit for consistency.

PR #735, opened 2018-01-19
CLI Fix - Entity Select

ctsims:
Entity select was attempting to match the numeric input [IE: '3'] to the new reference map, causing a crash when anything was selected

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
@@ -233,7 +233,9 @@ public class EntityListSubscreen extends Subscreen<EntityScreen> {
 
 
         try {
-            host.setHighlightedEntity(input);
+            int chosenIndex = Integer.valueOf(input);
+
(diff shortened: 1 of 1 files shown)


amstone326:
@ctsims Was this caused by something recent?

ctsims:
I presume it was caused by merging formplayer in, since formplayer tracks inputs based on GUID ID's and not rows (since it's hard for it to keep consistent track of row ids)

amstone326:
Got it. FYI @wpride -- this looks like a place where you're going to have to do something more clever in order to get `formplayer` and `master` on the same page, then.

ctsims:
I think it should only affect the CLI, but I could be wrong.

wpride:
@ctsims Sorry about that

@amstone326 I thought I already was being clever - the `Integer.parseInt()` should throw a `NumberFormatException` if its a case ID selection and go to the catch statement where we use the Formplayer path. Looks like I dropped that `parseInt` call in my merge - sorry about that.

PR #738, opened 2018-01-23
use getClass()

kaapstorm:
Context: [FB 268583](https://manage.dimagi.com/default.asp?268583)

This is a guess: I don't have a local dev environment set up for FormPlayer to check this, so please test before merging.

@wpride cc @biyeun


kaapstorm:
nvm. I was looking at an outdated file.

PR #739, opened 2018-01-23
Tests for the CLI

wpride:
Small start with more framework than tests, but this runs through selecting a case and opening a form so should give us a base line to prevent any obvious regressions. I had to get a little bit creative to deal with assertions about the `ApplicationHost` and `XFormPlayer` input/output as well as with exiting (since in many situation the CLI would just loop indefinitely).

Could be that I missed some much more obvious way to deal with these things, let me know if so!


src/test/java/org/cli/CliTests.java
```diff
@@ -0,0 +1,178 @@
...
+        // Perform case selection
+        new CliTestRun<>("basic_app/basic.ccz",
+                "basic_app/restore.xml",
+                CaseTestReader.class,
+                "2 1 5 \n");
+    }
+
+    static abstract class CliTestReader extends BufferedReader {
+
+        private String[] args;
```
amstone326: `args` is the `steps` you provided, right? If so, can you name it that instead?

src/test/java/org/cli/CliTests.java
```diff
@@ -0,0 +1,178 @@
...
+            this.outStream = outStream;
+        }
+
+        @Override
+        public String readLine() throws IOException {
+            String output = new String(outStream.toByteArray(), StandardCharsets.UTF_8);
+            processLine(index, output);
+            String ret = args[index++];
+            outStream.reset();
+            return ret;
```
amstone326: I'm confused why this method returns 1 of your steps

src/test/java/org/cli/CliTests.java
```diff
@@ -0,0 +1,178 @@
...
+        new CliTestRun<>("basic_app/basic.ccz",
+                "basic_app/restore.xml",
+                CaseTestReader.class,
+                "2 1 5 \n");
+    }
+
+    static abstract class CliTestReader extends BufferedReader {
+
+        private String[] args;
+        private int index;
```
amstone326: Can you clarify what `index` represents with a more descriptive name and/or comment? It's tough to follow

src/translate/java/org/javarosa/engine/XFormPlayer.java
```diff
@@ -103,24 +108,12 @@ public void start(String formPath) throws FileNotFoundException {
         this.start(XFormUtils.getFormFromInputStream(new FileInputStream(formPath)));
     }
 
-    public void start(String formPath, Session session) throws FileNotFoundException {
```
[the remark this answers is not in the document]
wpride: yes

src/test/java/org/cli/CliTests.java
wpride: Added a comment explaining this class

wpride:
@amstone326 Thanks for the feedback! Updated, let me know if you want to offline about anything

ctsims:
This is awesome, Will! Thanks for making this part of normal test running.

No feedback beyond aliza's

build.gradle
```diff
@@ -84,6 +84,7 @@ dependencies {
       testCompile sourceSets.main.output
       testCompile sourceSets.api.output
       testCompile sourceSets.cli.output
+      testCompile sourceSets.translate.output
```
amstone326: Also not sure what this is for?

amstone326: @wpride Approving the changes but if this isn't needed then just remove it before you merge

wpride: This is necessary because the `translate` files (specifically `XFormPlayer`) are now under test

PR #744, opened 2018-01-31
Improve error message when sorting on an empty list

wpride:
re: field/dev

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
@@ -4,6 +4,7 @@ import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.core.util.DataUtil;
 import org.javarosa.xpath.XPathArityException;
+import org.javarosa.xpath.XPathException;
 import org.javarosa.xpath.XPathTypeMismatchException;
 import org.javarosa.xpath.parser.XPathSyntaxException;
 
@@ -51,6 +52,9 @@ public class XPathSortFunc extends XPathFuncExpr {
         } else {
             sortedList = sortSingleList(FunctionUtils.toString(evaluatedArgs[0]),
                     FunctionUtils.toBoolean(evaluatedArgs[1]));
+        }
+        if (sortedList.size() == 0) {
+            throw new XPathException(String.format("Called sort() on empty list with args %s", evaluatedArgs));
         }
         return DataUtil.listToString(sortedList);
     }
--- a/src/main/java/org/javarosa/xpath/expr/XPathSortByFunc.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathSortByFunc.java
@@ -5,6 +5,7 @@ import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.core.util.DataUtil;
 import org.javarosa.xpath.XPathArityException;
+import org.javarosa.xpath.XPathException;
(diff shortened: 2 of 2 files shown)


src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
```diff
@@ -52,6 +53,9 @@ protected Object evalBody(DataInstance model, EvaluationContext evalContext, Obj
             sortedList = sortSingleList(FunctionUtils.toString(evaluatedArgs[0]),
                     FunctionUtils.toBoolean(evaluatedArgs[1]));
         }
+        if (sortedList.size() == 0) {
```
shubham1g5: Wondering why can't we just return the same empty list instead of throwing an error ?

amstone326:
@wpride I wasn't paying that much attention to the convo on field/dev today, but I'm confused why checking whether the _sorted_ list is empty is the check that needs to be made here to prevent that error. It's hard to tell where the index out of range error is coming from exactly, but I would have thought it would happen _during_ the sorting process. At the point that you're checking for here, all that's left to do is `return DataUtil.listToString(sortedList)`, which shouldn't have any issue if `sortedList` is empty.

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
wpride: Good question, I hadn't thought about that. Given that we were crashing previously when this was called on an empty list I'm inclined to not change the behavior in this PR. I also think its pretty likely that if `sort()` is called with an empty list then that indicates a user error and we're better off crashing rather than hiding the issue.

wpride:
@amstone326 The crash (as seen on Sentry) was coming from here:

https://github.com/dimagi/commcare-core/blob/1faf5e184b3eb75fd6cd53522e76788fcc1bc6e3/src/main/java/org/javarosa/core/util/DataUtil.java#L56-L56

called *after* the sort rather than before. Where I added the code is simply the first time we have the args available as a list rather than as a String. We could parse the String arguments to check whether they'd result in an empty list before calling sort, but this was lower touch

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
shubham1g5: Hmm.. Though calling sort with a Xpath reference to a list which can sometimes evaluate to a empty list would also result in this error and won't always signify an error on user's part unless we always want that user to check the size of the list represented by xpath ref before calling sort with it.

amstone326:
@wpride Ah gotcha, I missed the substring line when I was looking to see how `listToString` could cause any issues. That makes sense, although going to join in the convo you're having with Shubham above.

src/main/java/org/javarosa/xpath/expr/XPathSortFunc.java
amstone326: Hey yea I definitely agree, good call Shubham-- it's completely plausible that an app builder would want to call `sort()` on something that will sometimes evaluate to empty and and sometimes not. Returning the empty list also seems like totally correct behavior to me from a theoretical standpoint.

wpride: I think being that strict about arguments is in line with how other our other XPath functions behave and we've had the onus be on users to account for null-conditions in their apps using `coalesce` when desired

shubham1g5: I agree with the Null-conditions part though don't think we should treat an empty list same as null. Like I would be pretty sad if Java Collections Sort throw an error for a var list that can be empty sometimes and I would have to check both for not null and not empty before calling sort.  

Though since we are already crashing for an empty list currently, I think throwing this error makes sense in context of this PR.

PR #746, opened 2018-02-01
Implementing app fixture handling for CommCareInstanceInitializer

wpride:
And other Formplayer parity cleanup

### DIFF
--- a/src/main/java/org/commcare/util/CommCarePlatform.java
+++ b/src/main/java/org/commcare/util/CommCarePlatform.java
@@ -2,13 +2,16 @@ package org.commcare.util;
 
 import org.commcare.resources.model.ResourceTable;
 import org.commcare.suite.model.Detail;
-import org.commcare.suite.model.FormEntry;
 import org.commcare.suite.model.Entry;
+import org.commcare.suite.model.FormEntry;
 import org.commcare.suite.model.Menu;
 import org.commcare.suite.model.OfflineUserRestore;
 import org.commcare.suite.model.Profile;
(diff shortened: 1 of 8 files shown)


wpride:
@amstone326 @shubham1g5 bump

src/main/java/org/commcare/util/CommCarePlatform.java
```diff
@@ -65,27 +71,30 @@ public int getMinorVersion() {
     }
 
     public Profile getCurrentProfile() {
+        if(cachedProfile != null) {
+            return cachedProfile;
+        }
         return (Profile)storageManager.getStorage(Profile.STORAGE_KEY).read(profile);
```
shubham1g5: Shouldn't we set the `cachedProfile` here before we return.

src/main/java/org/commcare/util/CommCarePlatform.java
```diff
@@ -102,9 +111,11 @@ public Entry getEntry(String entryId) {
 
     public void setProfile(Profile p) {
         this.profile = p.getID();
+        this.cachedProfile = p;
     }
 
     public void registerSuite(Suite s) {
+        installedSuites.add(s);
```
shubham1g5: In `getInstalledSuites()` we are using `installedSuites.addElement` instead of `installedSuites.add`. Is there a catch here due to which one needs to synchronized and other not!

wpride: No, just an oversight on my part, I don't think that synchronicity would be an issue here but changing this for consistency.

PR #748, opened 2018-02-13
Refactor ExtWrapList to be usable for any implementing class of List

amstone326:
Previously `ExtWrapList` could only be used to serialize/deserialize `Vector` objects. The tweaks in this PR allow it to be used for any class that implements the `List` interface (which is both more useful and more consistent with what the class name implies).

cross-request: https://github.com/dimagi/commcare-android/pull/1945

### DIFF
--- a/src/main/java/org/javarosa/core/util/externalizable/ExtWrapList.java
+++ b/src/main/java/org/javarosa/core/util/externalizable/ExtWrapList.java
@@ -3,27 +3,30 @@ package org.javarosa.core.util.externalizable;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.util.List;
 import java.util.Vector;
 
 
-//list of objects of single (non-polymorphic) type
+// List of objects of single (non-polymorphic) type
 public class ExtWrapList extends ExternalizableWrapper {
     public ExternalizableWrapper type;
     private boolean sealed;
+    private Class<? extends List> listImplementation;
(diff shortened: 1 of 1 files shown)


src/main/java/org/javarosa/core/util/externalizable/ExtWrapList.java
```diff
@@ -3,66 +3,83 @@
...
 
     @Override
     public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
         if (!sealed) {
             int size = (int)ExtUtil.readNumeric(in);
-            Vector<Object> v = new Vector<>(size);
-            for (int i = 0; i < size; i++) {
-                v.addElement(ExtUtil.read(in, type, pf));
+            try {
+                List<Object> l = listImplementation.newInstance();
```
ctsims: Initializing the list/vector with its size is actually a pretty serious performance win in the serialization layer, because it prevents alloc/gc's from needing to occur. 

It would even be fine (just for preventing performance regressions) to check if listImplemenation is a vector, and if so allocating it with new Vector(size), and using argumentless newInstance() otherwise. That won't let the other types of lists have that performance gain, but will prevent existing code from accidentally running slower, since they'll all be Vectors under the hood for now anyway.

amstone326: Ah cool makes sense, thanks for catching

PR #750, opened 2018-02-14
Include current language code in session metadata

amstone326:
Requested by ICDS in [this ticket](https://manage.dimagi.com/default.asp?270353) to make referencing the language code in a form more scalable. Just waiting for confirmation from Marissa that this solution is acceptable.

**Product Note**: Add the current language code to the commcaresession metadata, meaning that the expression `instance('commcaresession')/session/context/current-langcode` will return the language code for whatever language is currently active in the app


amstone326:
Closing this for now because Marissa seems to have changed her mind about wanting it (at least for now). Not going to delete the branch in case they come back and want it.

PR #751, opened 2018-02-14
Add "Single Load" event for Performance Traces

ctsims:
When doing performance traces, it used to be invisible when the app loaded metadata for a single record, which would make a lot of traces really confusing.

This change reveals single metadata loads in the trace results, making it clear when DB hits are occurring.

### DIFF
--- a/src/main/java/org/commcare/cases/instance/IndexedFixtureChildElement.java
+++ b/src/main/java/org/commcare/cases/instance/IndexedFixtureChildElement.java
@@ -103,7 +103,7 @@ public class IndexedFixtureChildElement extends StorageBackedChildElement<Storag
 
     protected String[] getElementMetadata(int recordId, String[] metaFields, QueryContext context) {
         if (context == null || this.parent.getStorageCacheName() == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/cases/instance/IndexedFixtureChildElement.java
```diff
@@ -156,8 +156,18 @@ private TreeElement detectAndProcessLimitedScopeResponse(int recordId, QueryCont
         return recordObjectCache.getLoadedRecordObject(recordObjectKey, recordId);
     }
 
-    private String[] readSingleRecordMetadataFromStorage(int recordId, String[] metaFields) {
-        return parent.storage.getMetaDataForRecord(recordId, metaFields);
+    private String[] readSingleRecordMetadataFromStorage(int recordId, String[] metaFields, QueryContext context) {
+        EvaluationTrace trace = new EvaluationTrace(String.format("Model [%s]: Single Metadata Load", this.parent.getStorageCacheName()));
```
shubham1g5: `this` is repetetive in `this.parent`

PR #753, opened 2018-02-15
Formplayer 2.42

wpride:
Backmerge of 2.42 into `formplayer`


src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -323,14 +341,17 @@ private void printErrorAndContinue(String error, Exception e) {
...
-
         if (next == null) {
             //XFORM TIME!
             return null;
         } else if (next.equals(SessionFrame.STATE_COMMAND_ID)) {
             return new MenuScreen();
         } else if (next.equals(SessionFrame.STATE_DATUM_VAL)) {
-            return new EntityScreen();
+            return new EntityScreen(true);
+        } else if (next.equals(SessionFrame.STATE_QUERY_REQUEST)) {
```
amstone326: Why is it that these are just getting added now when this functionality has existed for so long?

wpride: Web Apps doesn't use `ApplicationHost` as its runner, `MenuSession` has its own control logic

PR #757, opened 2018-02-16
Add LocalizerManager class

wpride:
This class allows Android and Web Apps to share the static `Localizer` code by introducing a manager to mediate between strategies. Taken from class comment:

(Yet another) Manager class, this one for determining which localization strategy to use.
The options are:

1. staticLocalizer: Static variable for platforms where the same Localizer can be safely shared across all threads on the JVM (Android)
2. threadLocalLocalizer: ThreadLocal variable for platforms where different threads are potentially running separate applications (Web Apps)

Defaults to the static Localizer. Web Apps should sets the strategy to useThreadLocal = `true` immediately on startup.

### DIFF
--- a/src/main/java/org/javarosa/core/services/locale/Localization.java
+++ b/src/main/java/org/javarosa/core/services/locale/Localization.java
@@ -7,31 +7,23 @@ import java.util.Hashtable;
 
 public class Localization {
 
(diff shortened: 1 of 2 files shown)


ctsims:
@wpride this looks totally right to me. Nice

PR #758, opened 2018-02-16
Update XPathExpression isLeafNode

wpride:
Some changes made by @benrudolph way back when he was working on debugger serialization (I think):
https://github.com/dimagi/commcare-core/pull/591
https://github.com/dimagi/commcare-core/commit/96169d4a07109d3ef459a72c5f77bd95f306b70e

The `XPathNodeset` check is just moved up and the `refs == null` check would prevent an NPE so I think this is pretty harmless

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathExpression.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathExpression.java
@@ -74,10 +74,6 @@ public abstract class XPathExpression implements Externalizable, XPathAnalyzable
         DataInstance instance = nodeset.getInstance();
         Vector<TreeReference> refs = nodeset.getReferences();
 
-        if (refs == null) {
-            return;
-        }
-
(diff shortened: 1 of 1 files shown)


src/main/java/org/javarosa/xpath/expr/XPathExpression.java
```diff
@@ -76,6 +73,11 @@ private static void serializeElements(XPathNodeset nodeset, OutputStream output)
 
         DataInstance instance = nodeset.getInstance();
         Vector<TreeReference> refs = nodeset.getReferences();
+
+        if (refs == null) {
+            return;
```
ctsims: I don't think it's good practice for us to assume that the nodeset is well formed if refs is null here. What circumstance is that arising in? I think it's a stronger change for us to encode that situation deliberately rather than assuming that it was a valid state for the nodeset to have arrived in.

PR #761, opened 2018-02-20
ReferenceManager refactor

wpride:
Implements a similar pattern to in https://github.com/dimagi/commcare-core/pull/757 but for `ReferenceManager`. Unfortunately had to duplicate much of the code in that PR because generic objects cannot be static

The managed class was already named `Manager` so I had to choose between calling this new class `ReferenceManagerManager` or some other synonym, so I settled on `Handler`... lol. Considered renaming `ReferenceManager` to `ReferenceResolver` or `ReferenceDeriver` and then naming the new class `ReferenceManager` but decided to keep this lower touch... open to suggestions!

Most of the diff is just moving the `instance()` method from `ReferenceManger` into the new class

### DIFF
--- a/src/cli/java/org/commcare/util/engine/CommCareConfigEngine.java
+++ b/src/cli/java/org/commcare/util/engine/CommCareConfigEngine.java
@@ -27,7 +27,7 @@ import org.javarosa.core.io.StreamsUtil;
 import org.javarosa.core.model.FormDef;
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.FormInstance;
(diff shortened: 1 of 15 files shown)


ctsims:
Hey Will,

not a big deal, but I think you may have gotten my comment backwards. 

The intent of the change I was asking for was that ReferenceManager's interface *not change* as it is the primary interface to the functionality, which would have resulted in there being no changes to the interface seen by the rest of the code. It would mean that `ReferenceManger` now refers to a different class (which is kind of sketchy), but I think it's the less burdensome change.

Imagining, for a minute, that someone else was using commcare-core, this change now requires them to update all of their code.

PR #764, opened 2018-02-21
Split non-DB centric sorting into commcare-core utility

wpride:
Move the non-database layer related entity sorting code into `commcare-core`

cross-request: https://github.com/dimagi/commcare-android/pull/1950

### DIFF
--- a/src/main/java/org/commcare/util/EntitySortUtil.java
+++ b/src/main/java/org/commcare/util/EntitySortUtil.java
@@ -1,0 +1,83 @@
+package org.commcare.util;
+
+import org.commcare.cases.entity.Entity;
+import org.commcare.cases.util.StringUtils;
+import org.commcare.modern.util.Pair;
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/util/EntitySortUtil.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+import java.util.Locale;
+
+/**
+ * Created by willpride on 2/21/18.
+ */
+
+public class EntitySortUtil {
+
+    public abstract static class EntitySortCallbackListener {
```
amstone326: nit but can this be called something else? `EntitySortCallbackListener` seems overly complex and non-specific when I think it would make sense to call it something like `EntityProvider`? Let me know if you disagree. Also, is there a reason this can't be an interface?

wpride:
@amstone326 Good call, thanks

PR #774, opened 2018-02-26
Revert removing EntityDatum instanceof check from guessUnknown

wpride:
In commit https://github.com/dimagi/commcare-core/commit/6dacb1848a8a5edbc5bb0c16448e68c1ba009d2e I removed the check `datum instanceof EntityDatum`. Unfortunately I can't remember what this change was attempting to address (September 2016...) and can't find the PR containing this change (likely because this was snuck in with a submodule update rather than a proper PR into `formplayer` branch). This change seems illogical to me now and the tests are passing, so I'm reverting the change.

### DIFF
--- a/src/main/java/org/commcare/session/CommCareSession.java
+++ b/src/main/java/org/commcare/session/CommCareSession.java
@@ -366,7 +366,8 @@ public class CommCareSession {
                 }
                 Vector<SessionDatum> data = entry.getSessionDataReqs();
                 for (SessionDatum datum : data) {
-                    if (datum.getDataId().equals(poppedId)) {
+                    if (datum instanceof EntityDatum &&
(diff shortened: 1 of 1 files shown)


amstone326:
@wpride This seems dangerous to me; is there something specific that makes you think this should be reverted? Given all of the confusing trouble we've had with `guessUnknownType()` and its importance to proper functioning of behavior in the session (namely what happens when the user presses 'back'), I'm inclined to leave whatever has been in the code and working.

wpride:
@amstone326 I did know that; the `commcare-core` tests also run as part of the `formplayer` build process so failures would be caught when you tried to update the submodule reference. I can't think of a good reason not to run tests on non-master merges, so happy to make that change

PR #776, opened 2018-02-27
break out regex tests and add a few more

dannyroberts:
in XPathEvalTest

PRing this separately to prove it passes before https://github.com/dimagi/commcare-core/pull/641, which also includes this commit to prove it passes there as well.

### DIFF
--- a/src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
+++ b/src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
@@ -118,6 +118,21 @@ public class XPathEvalTest {
     }
 
     @Test
+    public void testRegex() {
+        EvaluationContext ec = getFunctionHandlers();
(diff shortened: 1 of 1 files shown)


dannyroberts:
Hmm. I'm confused about the build—when does jenkins report the build status?

amstone326:

amstone326:
@dannyroberts errr something is broken, looking into it. Thanks for flagging.

wpride:
Copying response from other PR for historical purposes "I did know that; the commcare-core tests also run as part of the formplayer build process so failures would be caught when you tried to update the submodule reference. I can't think of a good reason not to run tests on non-master merges, so happy to make that change"

amstone326:

amstone326:
@wpride @dannyroberts Sweet, that was easy. I just added the `formplayer` branch as 1 of the target branches in the jenkins job config, and the tests just ran/passed.

PR #778, opened 2018-03-02
Fixing getEntriesForCommand

amstone326:
cross-request: https://github.com/dimagi/commcare-android/pull/1961

### DIFF
--- a/src/main/java/org/commcare/session/CommCareSession.java
+++ b/src/main/java/org/commcare/session/CommCareSession.java
@@ -180,13 +180,14 @@ public class CommCareSession {
      *
      * @param entry
      * @param currentSessionData
-     * @return true if the datum IDs in currentSessionData are a subset of @entry's datum requirements
+     * @return true if @entry needs all of the datums that are already added to the current session
      */
     private static boolean entryRequiresAllDataInSession(Entry entry,
                                                          OrderedHashtable<String, String> currentSessionData) {
         Vector<SessionDatum> entryRequirements = entry.getSessionDataReqs();
 
         if (currentSessionData.size() > entryRequirements.size()) {
+            // currentSessionData needs to be a subset of entryRequirements, so we can short-circuit if this is the case
             return false;
         }
 
--- a/src/main/java/org/commcare/session/CommCareSession.java
+++ b/src/main/java/org/commcare/session/CommCareSession.java
@@ -120,7 +120,7 @@ public class CommCareSession {
     public Vector<Entry> getEntriesForCommand(String commandId) {
         return getEntriesForCommand(commandId, false);
     }
-    
+
     public Vector<Entry> getEntriesForCommand(String commandId, boolean includeNested) {
         return getEntriesForCommand(commandId, new OrderedHashtable<String, String>(), includeNested);
     }
--- a/src/main/java/org/commcare/session/CommCareSession.java
+++ b/src/main/java/org/commcare/session/CommCareSession.java
@@ -117,8 +117,12 @@ public class CommCareSession {
         }
     }
 
-    private Vector<Entry> getEntriesForCommand(String commandId) {
-        return getEntriesForCommand(commandId, new OrderedHashtable<String, String>(), false);
+    public Vector<Entry> getEntriesForCommand(String commandId) {
+        return getEntriesForCommand(commandId, false);
+    }
+    
+    public Vector<Entry> getEntriesForCommand(String commandId, boolean includeNested) {
+        return getEntriesForCommand(commandId, new OrderedHashtable<String, String>(), includeNested);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


amstone326:

src/test/java/org/commcare/backend/session/test/ChildModuleNavigationTests.java
```diff
@@ -24,12 +27,20 @@ public void testNeedsCommandFirst() throws Exception {
 
         // since there are two entries registered under 'parent-module' that
         // need different case data, we need to select the entry before the case
-        assertEquals(session.getNeededData(), SessionFrame.STATE_COMMAND_ID);
+        assertEquals(SessionFrame.STATE_COMMAND_ID, session.getNeededData());
```
amstone326: switching the order of these so that it's `(expected, actual)`, as the method specification dictates

src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -116,74 +117,92 @@ public CommCareSession(CommCareSession oldCommCareSession) {
...
         }
         for (Suite s : platform.getInstalledSuites()) {
             List<Menu> menusWithId = s.getMenusWithId(commandId);
             if (menusWithId != null) {
-                for (Menu menu : menusWithId) {
-                    entries.addAll(getEntriesFromMenu(menu, currentSessionData));
+                // make a copy so that we can use this as a queue
+                List<Menu> menusToExamine = new ArrayList<>(menusWithId);
+                while (!menusToExamine.isEmpty()) {
+                    Menu menu = menusToExamine.remove(0);
```
ctsims: Super minor point that if this was a Stack you could pop(), or a queue you could poll() and it would be a bit more clear how you're using the datatype.

src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -116,74 +117,92 @@ public CommCareSession(CommCareSession oldCommCareSession) {
...
+        Vector<Entry> stillValid = new Vector<>();
         for (String cmd : menu.getCommandIds()) {
-            Entry e = map.get(cmd);
+            Entry e = globalEntryMap.get(cmd);
             if (e == null) {
                 throw new RuntimeException("No entry found for menu command [" + cmd + "]");
             }
-            if (entryRequirementsSatsified(e, currentSessionData)) {
-                entries.addElement(e);
+            if (entryRequiresAllDataInSession(e, currentSessionData)) {
```
ctsims: The more I think about this, the less it makes sense to me.

Specifically in the end of end-of-form navigation or register from case list (entity list <action>'s) you could add the session frame
```
m_0
case_id_one
case_id_two
```
to the session, and have two menu declarations you are jumping to, each with an entry:
```
m_0
e_1
 - case_id_one
 - case_id_two

m_0
e_2
 - case_id_two
 - case_id_one
```
and e_1 and e_2 should both meet this check. If you visited the `m_0` form directly, you'd expect to need to choose e_1 or e_2 first, because their order of selecting id's is different (and they may have different filters), but once you've chosen both id's, semantically both forms should accept the two.

amstone326: Hey sorry, I'm not 100% sure I see the conclusion you're leading to with that, in terms of what you think is wrong with the current implementation. Are you objecting to the fact that `entryRequiresAllDataInSession()` insists that the matching datum IDs match _in order_? If so, I think I agree that that would be a problem in the edge case you gave. But is that all you meant, or is there something else/more than that?

src/main/java/org/commcare/session/CommCareSession.java
amstone326: good call

src/main/java/org/commcare/session/CommCareSession.java
amstone326: And just to make note of it now- we'd be taking a small performance hit if we stopped doing the in-order check and just did N `contains` checks instead. Given that N is never large in the context of # of datums on the session stack, I think this is totally fine, but I had been assuming that that was part of why it was done that way to begin with.

ctsims: Hey,

Yeah, the in-order check is the most relevant part. I'm also honestly not clear on whether

```
case_id_one
case_id_three
case_id_two
```

_shouldn't_ match something that needs
```
case_id_one
case_id_two
```

Based on how we're using this code, which is to decide whether an entry is relevant, right? I think the only thing that makes an entry irrelevant based on what happens in the session are commands, rather than datums.

amstone326: Hm so I think I'm coming to a similar conclusion, but with slightly different reasoning (or maybe we are saying the same thing in different ways). But my thinking is this:

One of the contracts we adhere to in the CommCareSession is that we never have a user select a datum (or have the engine perform the computation of a computed datum) that isn't going to be used/needed by whatever form the user ends up at in that session. This is guaranteed by [this line](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/session/CommCareSession.java#L207), where we only return a datum from `getNeededData()` if it's required by _all_ the remaining valid entries. So on the assumption that there are no flaws in that implementation, then I think we can just say that the check performed by `entryRequiresAllDataInSession()` is repetitive, and therefore not needed. Does that reasoning sound right to you?

PR #779, opened 2018-03-11
Add common hashtags to ignore-set

wpride:
Vellum adds this XML consistently so should suppress warning


wpride:
@amstone326 I think @orangejenny would be able to provide a better explanation, but my understanding is that form builder includes these `hashtagTransforms` in order to map the `#case/` and `#form/` syntax to the expected `/data` paths at build time

orangejenny:
Will's explanation is good.

Thanks for doing this, it always just adds a little extra distraction when people have a cyclical reference problem and also see this warning pop up.

PR #784, opened 2018-04-03
Remove readonly ODK attribute

wpride:
#kill unused/annoying `readonly` attribute inherited from ODK

### DIFF
--- a/src/main/java/org/commcare/xml/RootParser.java
+++ b/src/main/java/org/commcare/xml/RootParser.java
@@ -22,7 +22,6 @@ public class RootParser extends ElementParser<RootTranslator> {
         this.checkNode("root");
 
         String id = parser.getAttributeValue(null, "prefix");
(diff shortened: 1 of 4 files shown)


shubham1g5:
It looks like a useful flag to me. Is there a reason we don't want to support this ?

wpride:
@shubham1g5 Mostly because as far as I know we're not using it (though this might be good to double check now that I think about it) and YAGNI

ctsims:
Hey, so this isn't an ODK inherited thing, this is part of the xforms spec

https://www.w3.org/TR/2001/WD-xforms-20010828/slice5.html#model-prop-readOnly

I'm opposed to removing things that are spec compliant, even if we aren't specifically using them, unless they are causing serious issues

amstone326:
@wpride can this be killed based on the above comments?

PR #797, opened 2018-04-30
2.43 - Fix regression by changing strategy

amstone326:
Going to be sort of hard to explain this so bear with me.. 

This PR reverts the primary change made in https://github.com/dimagi/commcare-core/pull/778, which was to make `getEntriesForCommand()` take in a boolean `includeNested` argument, in order to allow for the inclusion of entries belonging child modules in the list of entries. However, this change has previously unforeseen consequences on the behavior of existing apps that are too extreme to be acceptable. The best way to explain the problem is with an example: I discovered this while working with [this app](https://www.commcarehq.org/a/inddex-test/apps/view/5ee0251a637c4ff2ad21ddd17e140d8b/), which contains one menu (`m1`) that contains a lot of entries, as well as one child menu (`m9`). Prior to the change made in https://github.com/dimagi/commcare-core/pull/778, `getEntriesForCommand()` when called after the user selects `m1` would only return the entries that are direct children of `m1`, all of which require the same datum `case_id`. This resulted in the case list for selecting that datum always showing up before `m1` was displayed, which enabled the app builder to add display conditions to some of the entries within m1 that _reference_ the `case_id` datum within the session. However, once the entries within `m9` started getting included in the results of `getEntriesForCommand()`, the datum `case_id` was no longer needed by all entries in the result set, and so we were then trying to show the menu `m1` _before_ collecting `case_id`, at which point the display conditions on those entries throw an xpath error.

Given the potential for any number of apps out in the wild to have been relying on this behavior in the same way, I don't see how we can release this change, even though it is technically correct in terms of the CommCare session model.

So, what I've done here is to revert that change and instead make a different change that addresses the problem I had been trying to solve. I don't like it as much from a completeness standpoint, but it is much safer and gets the job done. I've highlighted and explained the new change inline in the PR.

### DIFF
--- a/src/main/java/org/commcare/session/CommCareSession.java
+++ b/src/main/java/org/commcare/session/CommCareSession.java
@@ -118,24 +118,12 @@ public class CommCareSession {
         }
     }
 
-    public Vector<Entry> getEntriesForCommand(String commandId) {
-        return getEntriesForCommand(commandId, false);
-    }
-
-    public Vector<Entry> getEntriesForCommand(String commandId, boolean includeNested) {
-        return getEntriesForCommand(commandId, new OrderedHashtable<String, String>(), includeNested);
-    }
-
     /**
      * @param commandId          the current command id
-     * @param currentSessionData all of the datums already on the stack
-     * @param includeNested     whether to include entries from child modules
      * @return A list of all of the form entry actions that are possible with the given commandId
      * and the given list of already-collected datums
      */
-    private Vector<Entry> getEntriesForCommand(String commandId,
-                                               OrderedHashtable<String, String> currentSessionData,
-                                               boolean includeNested) {
+    private Vector<Entry> getEntriesForCommand(String commandId) {
         Vector<Entry> entries = new Vector<>();
         if (commandId == null) {
             return entries;
@@ -143,14 +131,8 @@ public class CommCareSession {
         for (Suite s : platform.getInstalledSuites()) {
             List<Menu> menusWithId = s.getMenusWithId(commandId);
             if (menusWithId != null) {
-                // make a copy so that we can use this as a queue
-                Queue<Menu> menusToExamine = new LinkedList<>(menusWithId);
-                while (!menusToExamine.isEmpty()) {
-                    Menu menu = menusToExamine.poll();
-                    entries.addAll(getStillValidEntriesFromMenu(menu, currentSessionData));
-                    if (includeNested) {
-                        menusToExamine.addAll(s.getMenusWithRoot(menu.getId()));
-                    }
+                for (Menu menu : menusWithId) {
+                    entries.addAll(getStillValidEntriesFromMenu(menu));
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -118,39 +118,21 @@ public CommCareSession(CommCareSession oldCommCareSession) {
...
             if (menusWithId != null) {
-                // make a copy so that we can use this as a queue
-                Queue<Menu> menusToExamine = new LinkedList<>(menusWithId);
-                while (!menusToExamine.isEmpty()) {
-                    Menu menu = menusToExamine.poll();
-                    entries.addAll(getStillValidEntriesFromMenu(menu, currentSessionData));
-                    if (includeNested) {
-                        menusToExamine.addAll(s.getMenusWithRoot(menu.getId()));
-                    }
+                for (Menu menu : menusWithId) {
```
amstone326: Reverted back to this behavior and removed the logic surrounding `includeNested`

src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -161,8 +143,7 @@ public CommCareSession(CommCareSession oldCommCareSession) {
         return entries;
     }
 
-    private Vector<Entry> getStillValidEntriesFromMenu(Menu menu,
-                                                       OrderedHashtable<String, String> currentSessionData) {
```
amstone326: `currentSessionData` was unused after the changes made in https://github.com/dimagi/commcare-core/pull/778/ but got left in by accident, so just fixing that now

src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -195,19 +176,19 @@ public String getNeededData(EvaluationContext evalContext) {
...
         if (needDatum != null) {
             return needDatum;
-        } else if (remainingValidEntries.isEmpty()) {
-            throw new RuntimeException("Collected datums don't match required datums for entries at command " + currentCmd);
-        } else if (remainingValidEntries.size() == 1
-                && remainingValidEntries.elementAt(0) instanceof RemoteRequestEntry
-                && ((RemoteRequestEntry)remainingValidEntries.elementAt(0)).getPostRequest().isRelevant(evalContext)) {
+        } else if (entriesForCurrentCommand.isEmpty()) {
+            // No entries available directly within the current command, so we must need to select another menu
+            return SessionFrame.STATE_COMMAND_ID;
```
amstone326: This is the new change: If the current command has no entries, instead of throwing an exception, we assume that this is just because there no entries at this level but there are some nested in child modules

ctsims:
I won't have the headspace to review this super deeply, but I reviewed over your comments and questions and do agree that the nesting changes the meaning of having an empty entity set. It seems like something that was aligned with the old weird behavior (where things had to be in perfect order to be right) and is no longer part of the expectaiton

PR #803, opened 2018-05-10
Supporting changes for recovery measures

amstone326:
cross-request: https://github.com/dimagi/commcare-android/pull/1997


shubham1g5:

src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -872,7 +874,28 @@ public void clear(CommCarePlatform platform) {
         if (count > 0) {
             Logger.log(LogTypes.TYPE_RESOURCES, "Cleaned up " + count + " records from table");
         }
+        storage.removeAll();
+    }
 
+    /**
+     * Destroy this table, and remove any files installed by it.
+     */
+    public void clearAll(CommCarePlatform platform) {
```
ctsims: This is so close to the one above that I would merge them even if it makes the resulting method more complex

src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -872,7 +874,28 @@ public void clear(CommCarePlatform platform) {
         if (count > 0) {
             Logger.log(LogTypes.TYPE_RESOURCES, "Cleaned up " + count + " records from table");
         }
+        storage.removeAll();
```
ctsims: What's the reasoning behind this change?

src/main/java/org/commcare/resources/model/Resource.java
```diff
@@ -98,6 +98,9 @@
     // Resource is transitioning from being installed to being upgraded
     public static final int RESOURCE_STATUS_INSTALL_TO_UPGRADE = 21;
 
+    // Constant to deenote all resources in a table
```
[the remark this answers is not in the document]
shubham1g5: it's actually `denote`. This is just getting used in [clearByStatus](https://github.com/dimagi/commcare-core/blob/recovery-measures/src/main/java/org/commcare/resources/model/ResourceTable.java#L885)  to turn off the filtering by status. Realized that this should be part of `ResourceTable.java` so moved it there.

PR #806, opened 2018-05-18
Wsp/ethiopian date fix

wpride:
Naive but I think possibly correct fix for [this](https://manage.dimagi.com/default.asp?276244) bug

Currently when you create a new date without time such as in a date picker, the input to this function is IE `05-11-2018`. This then gets set to midnight of this day, which if there is a negative timezone adjustment will get shifted to the previous day. By setting the default time to be midday instead of midnight we can easily avoid this.

Whenever messing with dates/times/timezones I feel like every step forward is also a step backwards but every test is passing and this makes sense to me so

### DIFF
--- a/src/main/java/org/javarosa/core/model/utils/DateUtils.java
+++ b/src/main/java/org/javarosa/core/model/utils/DateUtils.java
@@ -221,6 +221,10 @@ public class DateUtils {
         cd.set(Calendar.SECOND, df.second);
         cd.set(Calendar.MILLISECOND, df.secTicks);
 
+        if (df.hour == 0 && df.minute == 0 && df.second == 0 && df.secTicks == 0) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/javarosa/core/model/utils/DateUtils.java
```diff
@@ -221,6 +221,10 @@ private static Date getDate(DateFields df, int timezoneOffset) {
         cd.set(Calendar.SECOND, df.second);
         cd.set(Calendar.MILLISECOND, df.secTicks);
 
+        if (df.hour == 0 && df.minute == 0 && df.second == 0 && df.secTicks == 0) {
```
amstone326: This feels like it can't always be the right thing to do. For instnace, what if the time happens to actually have been meant to be midnight?

PR #807, opened 2018-05-21
Adds Drift Interceptor

shubham1g5:
* Calculates Drift and saves it in Prefs
* Adds `CommCarePreferenceManagerFactory` and `ICommCarePreferenceManager` to wire Android prefs in core library.

### DIFF
--- a/src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
+++ b/src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
@@ -1,9 +1,14 @@
 package org.commcare.core.network;
 
+import org.commcare.core.services.CommCarePreferenceManagerFactory;
+import org.commcare.core.services.ICommCarePreferenceManager;
 import org.commcare.util.LogTypes;
 import org.javarosa.core.services.Logger;
 
 import java.io.IOException;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 import java.util.concurrent.TimeUnit;
 
 import okhttp3.HttpUrl;
@@ -13,6 +18,8 @@ import okhttp3.Request;
 import okhttp3.Response;
 import retrofit2.Retrofit;
 
+import static org.javarosa.core.model.utils.DateUtils.HOUR_IN_MS;
+
 /**
  * Provides an instance of CommCareNetworkService.
  * We have declared everything static in this class as we want to use the same objects (OkHttpClient, Retrofit, …) throughout the app
@@ -20,6 +27,9 @@ import retrofit2.Retrofit;
  */
 
 public class CommCareNetworkServiceGenerator {
+
+    public static final String CURRENT_DRIFT = "current_drift";
+    public static final String MAX_DRIFT_SINCE_LAST_HEARTBEAT = "max_drfit_since_last_heartbeat";
 
     // Retrofit needs a base url to generate an instance but since our apis are fully dynamic it's not getting used.
     private static final String BASE_URL = "http://example.url/";
@@ -42,6 +52,32 @@ public class CommCareNetworkServiceGenerator {
         }
     };
 
+    private static Interceptor driftInterceptor = new Interceptor() {
+        @Override
+        public Response intercept(Chain chain) throws IOException {
+            Request request = chain.request();
+            Response response = chain.proceed(request);
+            ICommCarePreferenceManager commCarePreferenceManager = CommCarePreferenceManagerFactory.getCommCarePreferenceManager();
+            if (commCarePreferenceManager != null) {
+                String serverDate = response.header("date");
+                try {
+                    long serverTimeInMillis = new SimpleDateFormat("EEE, dd MMM yyyy hh:mm:ss zzz").parse(serverDate).getTime();
+                    long now = new Date().getTime();
+                    long currentDrfit = (now - serverTimeInMillis) / HOUR_IN_MS;
+                    commCarePreferenceManager.putLong(CURRENT_DRIFT, currentDrfit);
+                    long maxDrfitSinceLastHeartbeat = commCarePreferenceManager.getLong(MAX_DRIFT_SINCE_LAST_HEARTBEAT, 0);
+                    currentDrfit *= currentDrfit < 0 ? -1 : 1; // make it positive to calculate max drift
+                    if (currentDrfit > maxDrfitSinceLastHeartbeat) {
+                        commCarePreferenceManager.putLong(MAX_DRIFT_SINCE_LAST_HEARTBEAT, currentDrfit);
+                    }
+                } catch (ParseException e) {
+                    e.printStackTrace();
+                }
+            }
+            return response;
+        }
+    };
+
     private static AuthenticationInterceptor authenticationInterceptor = new AuthenticationInterceptor();
 
     private static OkHttpClient.Builder httpClient = new OkHttpClient.Builder()
@@ -49,6 +85,7 @@ public class CommCareNetworkServiceGenerator {
             .readTimeout(ModernHttpRequester.CONNECTION_SO_TIMEOUT, TimeUnit.MILLISECONDS)
             .addNetworkInterceptor(redirectionInterceptor)
             .addInterceptor(authenticationInterceptor)
+            .addInterceptor(driftInterceptor)
             .followRedirects(true);
 
     private static Retrofit retrofit = builder.client(httpClient.build()).build();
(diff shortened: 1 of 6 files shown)


src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -39,13 +49,40 @@
...
+            ICommCarePreferenceManager commCarePreferenceManager = CommCarePreferenceManagerFactory.getCommCarePreferenceManager();
+            if (commCarePreferenceManager != null) {
+                String serverDate = response.header("date");
+                try {
+                    long serverTimeInMillis = new SimpleDateFormat("EEE, dd MMM yyyy hh:mm:ss zzz").parse(serverDate).getTime();
+                    long now = new Date().getTime();
+                    long currentDrfit = (now - serverTimeInMillis) / HOUR_IN_MS;
+                    commCarePreferenceManager.putLong(CURRENT_DRIFT, currentDrfit);
+                    long maxDrfitSinceLastHeartbeat = commCarePreferenceManager.getLong(MAX_DRIFT_SINCE_LAST_HEARTBEAT, 0);
+                    currentDrfit *= currentDrfit < 0 ? -1 : 1; // make it positive to calculate max drift
```
amstone326: It looks like you should also be doing this before you store `CURRENT_DRIFT ` in app prefs, because in `HomeScreenBaseActivity::checkForDrift`, you're checking `if (getCurrentDrift() > 0)`

amstone326: (Also I think it just makes sense conceptually so think of drift as being an absolute value)

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -39,13 +49,40 @@
         return response;
     };
 
+    private static Interceptor driftInterceptor = new Interceptor() {
+        @Override
+        public Response intercept(Chain chain) throws IOException {
+            Request request = chain.request();
+            Response response = chain.proceed(request);
+            ICommCarePreferenceManager commCarePreferenceManager = CommCarePreferenceManagerFactory.getCommCarePreferenceManager();
+            if (commCarePreferenceManager != null) {
```
amstone326: The way `CommCarePreferenceManagerFactory` is implemented right now, anyone who ever wants to use it will have to first do this check for if `CommCarePreferenceManagerFactory.getCommCarePreferenceManager()` is null. Would it maybe make more sense to have the default state of the factory be to have a dummy implementation of `ICommCarePreferenceManager` always initialized?

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -21,6 +28,9 @@
 
 public class CommCareNetworkServiceGenerator {
 
+    public static final String CURRENT_DRIFT = "current_drift";
+    public static final String MAX_DRIFT_SINCE_LAST_HEARTBEAT = "max_drfit_since_last_heartbeat";
```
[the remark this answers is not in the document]
shubham1g5: Shit. I really messed this up!

src/main/java/org/commcare/core/services/CommCarePreferenceManagerFactory.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Wiring to allow access to Android preferences in commcare-core, but can potentially be used for any temporary key value storage
+ */
+public class CommCarePreferenceManagerFactory {
+
+    private static ICommCarePreferenceManager sCommCarePreferenceManager;
+
+    public static void init(ICommCarePreferenceManager mCommCarePreferenceManager) {
+        sCommCarePreferenceManager = mCommCarePreferenceManager;
```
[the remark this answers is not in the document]
shubham1g5: s = static, removing m

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: Ahh I wanted to be able to tell whether device time is ahead or behind. I am going to change the  `getCurrentDrift() > 0` to `getCurrentDrift() != 0` instead ?

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: I am a bit conflicted in this. I will surely love not to do these null checks though adding a dummy class would just create more complexity in code base.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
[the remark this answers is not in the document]
amstone326: Hahah it's ok!

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
amstone326: That also works, but I don't see anywhere in the code where you're actually checking whether device time is ahead or behind. Do you mean that we want to be able to do that on HQ?

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
amstone326: Well what if instead of having the `ICommCarePreferenceManager` interface, you just turn that into a class, and then `AndroidPreferenceManager` can extend it?

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: Yup, I meant on HQ.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: I think that will result in PreferenceManager classes not to bind to any contract which can get messy since we might add a particular method in one preference manager class and not in other. So I would actually prefer adding a dummy class to removing the interface if we do wanna avoid null checks for this.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
```diff
@@ -39,13 +49,40 @@
...
+        public Response intercept(Chain chain) throws IOException {
+            Request request = chain.request();
+            Response response = chain.proceed(request);
+            ICommCarePreferenceManager commCarePreferenceManager = CommCarePreferenceManagerFactory.getCommCarePreferenceManager();
+            if (commCarePreferenceManager != null) {
+                String serverDate = response.header("date");
+                try {
+                    long serverTimeInMillis = new SimpleDateFormat("EEE, dd MMM yyyy hh:mm:ss zzz").parse(serverDate).getTime();
+                    long now = new Date().getTime();
+                    long currentDrift = (now - serverTimeInMillis) / HOUR_IN_MS;
```
amstone326: Just want to double check that this was an intended result- This means that we're counting a drift value of less than an hour as no drift. Is that what we want?

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
amstone326: I think we should add the dummy class then. Otherwise it's going to be very non-obvious to later users of this interface that the pref manager might be null, and could lead to unnecessary NPEs in the code.

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: Yeah. Drift is supposed to convey no. of hours and we are neglecting anything less than that.

amstone326: 👍

src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
shubham1g5: I feel like that will cause more checks in the code. For example lets say we decide to return -1 from `getLong` method of the `DummyCommCarePreferenceManager` class. Then the calling method will never know if it's a valid value or not and will again need to some sort of checks to determine this.  

Also today AS gives you a warning if you don't do a null check before accessing `ICommCarePreferenceManager` because of the `Nullable` annotation on `CommCarePreferenceManagerFactory::getCommCarePreferenceManager()`

amstone326: Hmm okay that's fair, I'll concede :)

PR #810, opened 2018-06-06
Fixes ClassCastException ViewEntry can't be cast to FormEntry

shubham1g5:
Correcting for assumption that all entries in a suite are of type `FormEntry`


shubham1g5:

PR #814, opened 2018-06-19
Add external_id as indexable case property

wpride:
cross-request: https://github.com/dimagi/commcare-android/pull/2010

### DIFF
--- a/src/test/java/org/commcare/cases/test/CaseExternalIdTest.java
+++ b/src/test/java/org/commcare/cases/test/CaseExternalIdTest.java
@@ -14,17 +14,16 @@ import org.junit.runners.Parameterized;
 import java.util.Collection;
 
 /**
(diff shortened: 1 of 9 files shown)


src/test/java/org/commcare/cases/test/CaseExternalIdTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.util.Collection;
+
+/**
+ * Test suite to verify end-to-end parsing of inbound case XML
+ * and reading values back from the casedb model
+ *
+ * @author ctsims
```
ctsims: author should be you.

Also: other test cases: 
1) [external_id][case_id] (dual indexes)
2) [external_id] (with no cases in the casedb)
3) [external_id and true()]

PR #817, opened 2018-07-03
Adds Resource Recovery Implementation in Resource Table

shubham1g5:
cross-request:https://github.com/dimagi/commcare-android/pull/2015


src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -60,8 +62,8 @@
     private boolean isResourceProgressStale = false;
     // Cache for profile and suite 'parent' resources which are used in
     // references resolution
-    private final Hashtable<String, Resource> compoundResourceCache =
-            new Hashtable<>();
+    private final Hashtable<String, Resource> compoundResourceCache = new Hashtable<>();
+    private SizeBoundUniqueVector<Resource> mMissingResources = new SizeBoundUniqueVector<>(0);
```
wpride: What's the advantage to using a `SizeBoundUniqueVector` with limit `0` here rather than, say, a `Vector`?

shubham1g5: Nothing really. 

I wanted to use `SizeBoundUniqueVector` more since I don't want to get resources duplicated in data structure we use to track `mMissingResources`. I didnt initialize it to `null` to avoid null checks. It gets reinitialized to an appropriate limit later in `initializeResources`.

PR #819, opened 2018-08-31
Soft Asserts when resource remain uninitialized after attempting initialization 

shubham1g5:
refactoring for
cross_request:https://github.com/dimagi/commcare-android/pull/2024

### DIFF
--- a/src/main/java/org/commcare/resources/model/ResourceTable.java
+++ b/src/main/java/org/commcare/resources/model/ResourceTable.java
@@ -919,6 +919,12 @@ public class ResourceTable {
                 | XmlPullParserException | UnfullfilledRequirementsException e) {
             throw new ResourceInitializationException(r, e);
         }
+
+        // If resource is still completely unitialized, treat this as a missing resource
+        // so that recovery process can try to recover it
+        if (r.getStatus() == Resource.RESOURCE_STATUS_UNINITIALIZED) {
+            missingResources.add(r);
+        }
     }
 
     /**
@@ -1136,7 +1142,7 @@ public class ResourceTable {
     }
 
 
-    private void setMissingResources(SizeBoundUniqueVector<Resource> missingResources) {
+    public void setMissingResources(SizeBoundUniqueVector<Resource> missingResources) {
         mMissingResources = missingResources;
     }
 
--- a/src/main/java/org/commcare/resources/model/ResourceTable.java
+++ b/src/main/java/org/commcare/resources/model/ResourceTable.java
@@ -920,10 +920,9 @@ public class ResourceTable {
             throw new ResourceInitializationException(r, e);
         }
 
-        // If resource is still completely unitialized, treat this as a missing resource
-        // so that recovery process can try to recover it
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -919,6 +919,12 @@ private void attemptResourceInitialization(CommCarePlatform platform, boolean is
                 | XmlPullParserException | UnfullfilledRequirementsException e) {
             throw new ResourceInitializationException(r, e);
         }
+
+        // If resource is still completely unitialized, treat this as a missing resource
+        // so that recovery process can try to recover it
+        if (r.getStatus() == Resource.RESOURCE_STATUS_UNINITIALIZED) {
```
ctsims: Hm, this worries me a bit, do we not have a positive verification for resources that are in a bad state (IE: Resource.RESOURCE_STATUS_CORRUPTED or similar)?

I don't think of the 'recovery' process as being 'safe' for resources that are in an acceptable state, so I'm wary of the ambiguity between uninitialized and 'broken'

shubham1g5: Hey so just above this [here](https://github.com/dimagi/commcare-core/blob/f096271e4c9f16fc2c3ad0c1b2d5cedb80ec3953/src/main/java/org/commcare/resources/model/ResourceTable.java#L915), we are trying to initialize this resource. So even though `unitialized` is generally an accepted state, if just after trying to initialize the resource, the resource status is still sent to `unitialized` that means something is wrong with the resource which is stopping this from initializing and we should try reinstalling it.  It's foolproofing against unrecognized errors during initialization which would not have necessary marked the resource as corrupted.

shubham1g5: So I have not seen this in practice and was just speculating it. We already throw an exception and error out in cases of known failures in initializing in the block -
```
try {
            r.getInstaller().initialize(platform, isUpgrade);
        } catch (FileNotFoundException e) {
            missingResources.add(r);
        } catch (IOException | InvalidStructureException | InvalidReferenceException
                | XmlPullParserException | UnfullfilledRequirementsException e) {
            throw new ResourceInitializationException(r, e);
        }
```

 I was more worried about unseen failures but if it's not likely I think it's ok to remove these lines or maybe just log them as Sofet Asserts.

ctsims: Hm, so by far the most common way this corruption presents itself is with the translation files, right? What is that exception?

I think it might actually be an IOException rather than a FileNotFound, but I could be wrong. 

I think my preference would still be to soft assert rather than fail here, and/or to set a more specific expectation of what it means to come out of initialize() without the resource being initialized.

shubham1g5: Hey, So I had changed it to throw FNF earlier [here](https://github.com/dimagi/commcare-android/blob/recovery-measures/app/src/org/commcare/android/resource/installers/FileSystemInstaller.java#L73) for all File backed resources. 

Going to change it to Soft Assert now with expectation that we might discover some unseen errors later on.

PR #820, opened 2018-09-06
Fixes callout when plain text assumption prefix is added

shubham1g5:
Earlier we were trying to get the extra value after removing the prefix `cc:xpath_key:` from key which will always give null value when `cc:xpath_key:` is added to the key.


shubham1g5:
Closing in favour of https://github.com/dimagi/commcare-core/pull/821

PR #821, opened 2018-09-14
Fixes Callout Migration Strategy for ICDS

shubham1g5:
FB: https://manage.dimagi.com/default.asp?280426#1527079

*Overrides plain text assumption for all keys if any key has the prefix


shubham1g5:
@ctsims The neceesity for prefix key is just as a flag that makes the new apk support the old ccz since it tells us that if the prefix is not present, we still need to treat the values as plain text and not as xpath.

ctsims:
@shubham1g5 Yeah, totally get that, was just saying that we could move it to its own flag value "use_xpath_for_parsing='yes'" or something that wouldn't need to be coupled with another property (or substring evaluated). the content value could then just be ignored entirely (on the old platform it could parse as the raw vlaue 'yes', on the new platform it would parse as the xpath `yes`), but not super important.

shubham1g5:
@ctsims  Got it. Though leaving it the same to save time. Thanks.

PR #830, opened 2018-11-16
Temporary Fix for Improper Case Sync Algorithm

ctsims:
The case sync algorithm currently fails to propagate tags across alternating applications of the rules, meaning that some topologies that are in occasional use don't match up with HQ's implementation (see the "alternating_propogation" example in the new tests).

This doesn't fix that directly, but is a quick / temporary fix for the most common layouts until we can apply a real fix.


ctsims:
@shubham1g5 Great, thanks! Will pull.

I know it's a bit of a lot to ask, but would it be possible for you to apply this changeset back to LTS?

PR #834, opened 2018-12-24
Pass UnResolvedResourceExceptions during install to calling function

shubham1g5:
Not sure if it was intentional though the original UnResolvedResourceException thrown from the individual installers `install` was never making up to the calling Activity and hence was not getting shown in error message. 

cross-request: https://github.com/dimagi/commcare-android/pull/2060

### DIFF
--- a/src/main/java/org/commcare/resources/model/ResourceTable.java
+++ b/src/main/java/org/commcare/resources/model/ResourceTable.java
@@ -391,6 +391,8 @@ public class ResourceTable {
                             invalidResourceException = e;
                         } catch (UnreliableSourceException use) {
                             unreliableSourceException = use;
+                        } catch (UnresolvedResourceException ure) {
+                            unresolvedResourceException = ure;
                         }
                         if (handled) {
                             recordSuccess(r);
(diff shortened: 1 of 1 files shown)


ctsims:
@shubham1g5 Odd, the exception should be being thrown normally even without this change, right? Nothing in that method is "swallowing" it (IE: Catching it and not re-throwing it) so it looks like all this change does is allow the install to proceed around it (IE: try other installers in case one of those works). 

Does this change result in the error making it upwards? If so my guess is that it's due to a *different* exception being the eventual throw, rather than because a URE made it up (since it would be thrown either way)

src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -413,13 +414,17 @@ private void findResourceLocationAndInstall(Resource r,
                     // Continue until no resources can be found.
                 } catch (UnreliableSourceException use) {
                     unreliableSourceException = use;
+                } catch (UnresolvedResourceException ure) {
```
ctsims: Note: I believe changes to this call stack should also generally be replicated to line 390 above

PR #838, opened 2019-01-31
Adds a clear api to Logger

shubham1g5:
cross_request: https://github.com/dimagi/commcare-android/pull/2064

### DIFF
--- a/src/main/java/org/javarosa/core/services/Logger.java
+++ b/src/main/java/org/javarosa/core/services/Logger.java
@@ -82,7 +82,7 @@ public class Logger {
         throw crashException;
     }
 
-    public static void clear() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/javarosa/core/services/Logger.java
```diff
@@ -81,5 +81,9 @@ public void run() {
         }
         throw crashException;
     }
+
+    public static void clear() {
+        logger = null;
```
ctsims: Just a quick note that "clear()" sounds to me like it clears the *log entries* themselves rather than the statically registered logger.

Might be worth renaming to something like "detachLogger" or something similar to make the outcome more clear

PR #839, opened 2019-02-05
Logs reference in error

shubham1g5:
Hoping to get more help with this - https://www.fabric.io/dimagi/android/apps/org.commcare.lts/issues/5c0cadccf8b88c29631def90/sessions/5C591DD202390001137C0D557EB00B7E_DNE_0_v2?

### DIFF
--- a/src/main/java/org/javarosa/form/api/FormEntryController.java
+++ b/src/main/java/org/javarosa/form/api/FormEntryController.java
@@ -106,7 +106,8 @@ public class FormEntryController {
                     model.getForm().copyItemsetAnswer(q, element, data);
                 } catch (InvalidReferenceException ire) {
                     ire.printStackTrace();
-                    throw new RuntimeException("Invalid reference while copying itemset answer: " + ire.getMessage() + " Reference: " + ire.getInvalidReference().toString());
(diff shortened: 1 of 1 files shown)


src/main/java/org/javarosa/form/api/FormEntryController.java
```diff
@@ -106,7 +106,7 @@ public int answerQuestion(FormIndex index, IAnswerData data) {
                     model.getForm().copyItemsetAnswer(q, element, data);
                 } catch (InvalidReferenceException ire) {
                     ire.printStackTrace();
-                    throw new RuntimeException("Invalid reference while copying itemset answer: " + ire.getMessage());
+                    throw new RuntimeException("Invalid reference while copying itemset answer: " + ire.getMessage() + " Reference: " + ire.getInvalidReference().toString());
```
ctsims: I'm not clear on whether the reference here can be null, which should likely be protected so we don't obfuscate the runtime expresson.

I think "toString()" is an implicit, here so it may not be necessary anyway.

shubham1g5:
@ctsims  Merging it after making that minor change without waiting for re-review.

PR #842, opened 2019-02-27
Last sync support for Indexed Fixtures

shubham1g5:
Refactoring to support root atrributes lookup for Android indexed fixture.
* Adds `IndexedFixtureIdentifier` to hold basic/meta properties for an indexed fixture. 
* Modifies `IndexedFixtureInstanceTreeElement` to have hold attributes in `TreeElement attrHolder`
* Refactors `T getAttribute(String namespace, String name)` to `AbstractTreeElement getAttribute(String namespace, String name) ` to be able to return `last_sync` as an instance of `AbstractTreeElement` rather than `IndexedFixtureChildElement`. It's possible that it's not the right soultion here but I could not think of any alternates. 

cross-request: https://github.com/dimagi/commcare-android/pull/2085


src/main/java/org/javarosa/core/model/IndexedFixtureIndex.kt
```diff
@@ -0,0 +1,7 @@
+package org.javarosa.core.model;
+
+import org.javarosa.core.model.instance.TreeElement
+
+data class IndexedFixtureIndex(val name: String, val base: String, val child: String, val attrs: TreeElement?)
```
ctsims: Can we call this "IndexedFixtureIdentifier" or "IndexedFixtureBase" I think using the same word in two different contexts in an object name ("Index" refers to two different concepts) makes it harder to understand.

ctsims: Also, this class should have a docstring to convey what the class does

build.gradle
```diff
@@ -8,13 +9,16 @@ repositories {
...
+    ext.kotlin_version = '1.2.31'
     repositories {
         maven {
             url "https://plugins.gradle.org/m2/"
         }
+        mavenCentral()
     }
     dependencies {
         classpath "me.champeau.gradle:jmh-gradle-plugin:0.3.1"
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
ctsims: I'm excited to integrate some Kotlin code into the Android half of the codbase, but I don't think it's worth the value tradeoff today to integrate kotlin code into the core library until we have a bit more experience and direction.

src/main/java/org/commcare/cases/instance/IndexedFixtureInstanceTreeElement.java
```diff
@@ -74,4 +78,37 @@ protected IndexedFixtureChildElement getChildTemplate() {
...
+
+    @Override
+    public String getAttributeValue(int index) {
+        return attrHolder.getAttributeValue(index);
+    }
+
+    @Override
+    public AbstractTreeElement getAttribute(String namespace, String name) {
+        TreeElement attr = attrHolder.getAttribute(namespace, name);
+        attr.setParent(this);
```
ctsims: Can you confirm that this method has no side effects that affect the current object (`this`)?

src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
```diff
@@ -228,7 +228,7 @@ public TreeElement getAttribute(QueryContext context, String namespace, String n
     }
 
     @Override
-    public TreeElement getAttribute(String namespace, String name) {
+    public AbstractTreeElement getAttribute(String namespace, String name) {
```
ctsims: Why was this change necessary? It looks like the downstream method returns a treeelement anyway.

src/main/java/org/commcare/modern/database/IndexedFixturePathsConstants.java
```diff
@@ -1,28 +0,0 @@
-package org.commcare.modern.database;
-
-/**
- * @author Phillip Mates (pmates@dimagi.com)
- */
-public class IndexedFixturePathsConstants {
```
ctsims: I think FormPlayer also uses this data, so it's probably not possible to move it to Android

src/main/java/org/commcare/cases/instance/IndexedFixtureInstanceTreeElement.java
shubham1g5: The only affect is that [invalidates](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/instance/TreeElement.java#L809) the reference cache for `this`. Is this something to be worried about ?

ctsims: No, invalidating the reference cache is fine, I just wanted to make sure that this wouldn't result in a "Backlink" being created where "this (the treelement part of the object)" now has an independent reference to "attr"

shubham1g5: got it. Doesn't look like that's happening.

PR #843, opened 2019-03-01
Fixes deletion of indexed fixtures in case of no fixture entries

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/REACH-102

Fix for: When there are no children in a fixture node, we were not deleting the already existing entries from the corresponding ficture table. 

cross-request: https://github.com/dimagi/commcare-android/pull/2089


src/main/java/org/commcare/xml/IndexedFixtureXmlParser.java
```diff
@@ -89,6 +89,11 @@ private void processRoot(TreeElement root) throws IOException {
             for (TreeElement entry : root.getChildrenWithName(entryName)) {
                 processEntry(entry, indices);
             }
+        } else {
```
shubham1g5: This is the actual fix. Rest is just nice to have.

src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java
```diff
@@ -119,6 +119,9 @@
      */
     boolean exists(int id);
 
+    // Returns whether the container holding the records exists
+    boolean exists();
```
ctsims: I think name of this method is liable to introduce errors given that  from a method signature perspective it looks like it's overloading the method name above. A programmer could accidentally omit the argument for exists(recordId) and still end up with a viable method that does something quite different.

We currently treat these StorageUtility objects as being somewhat imperative, so I think it's kind of odd to introduce a lifecycle method like this. 

I have a counter-proposal for the change.

What about skipping the `.exists()` call above and just always calling the removeAll method here:
https://github.com/dimagi/commcare-core/pull/843/files#diff-9d656a3648f2e2389a84ebee93e7b9f7R95

and then protecting the removeAll() method call directly here(https://github.com/dimagi/commcare-android/pull/2089/files#diff-00f1ecd4116e85d8b92f01ac76ac7405R556) with a check for whether the table exists before removing all of the records? 

It's a little more opaque, but it also doesn't require introducing a lifecycle management concept into the core utility, which I think might be odd to consider piecemeal (IE: If 'exists' is a method and returns false in commcare-core, it's kind of odd that there isn't a way to, say, create the storage or remove it)

shubham1g5: @ctsims Made the change. I was worried above it being opaque though in retrospect agree that better to not manage lifecycle for this.

PR #845, opened 2019-03-11
Formplayer Update

shubham1g5:
Updating the formplayer branch.


shubham1g5:
Self merging this. @ctsims  Let me know if that's not the right process.

PR #848, opened 2019-03-21
added error context for XPathConditional

rristovic:
Product note: Added more context to the error message when there is an issue with constraints.

Jira task: https://dimagi-dev.atlassian.net/browse/MOB-16

Screenshot.
expression error

### DIFF
--- a/src/main/java/org/javarosa/xpath/XPathConditional.java
+++ b/src/main/java/org/javarosa/xpath/XPathConditional.java
@@ -51,11 +51,10 @@ public class XPathConditional implements IConditionExpr {
             if (xpath != null) {
                 throw new XPathUnsupportedException(xpath);
             } else {
-                if (evalContext.isConstraint) {
-                    String contextMessage = String.format(" Question with id: \'%s\' has issue with condition expression: \'%s\'",
-                            evalContext.getContextRef().getNameLast(), expr);
-                    e.setMessagePrefix(contextMessage);
-                }
+                String contextMessage = String.format(" Question with id: \'%s\' has issue with condition expression: \'%s\'",
+                        evalContext.getContextRef().getNameLast(), expr);
+                e.setMessagePrefix(contextMessage);
+
                 throw e;
             }
 
--- a/src/main/java/org/javarosa/xpath/XPathConditional.java
+++ b/src/main/java/org/javarosa/xpath/XPathConditional.java
@@ -51,6 +51,11 @@ public class XPathConditional implements IConditionExpr {
             if (xpath != null) {
                 throw new XPathUnsupportedException(xpath);
             } else {
+                if (evalContext.isConstraint) {
+                    String contextMessage = String.format(" Question with id: \'%s\' has issue with condition expression: \'%s\'",
+                            evalContext.getContextRef().getNameLast(), expr);
+                    e.setMessagePrefix(contextMessage);
+                }
                 throw e;
             }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/javarosa/xpath/XPathConditional.java
```diff
@@ -51,6 +51,11 @@ public Object evalRaw(DataInstance model, EvaluationContext evalContext) {
             if (xpath != null) {
                 throw new XPathUnsupportedException(xpath);
             } else {
+                if (evalContext.isConstraint) {
```
shubham1g5: @rristovic Any specific reason we need to add this condition ?

src/main/java/org/javarosa/xpath/XPathConditional.java
```diff
@@ -51,6 +51,11 @@ public Object evalRaw(DataInstance model, EvaluationContext evalContext) {
             if (xpath != null) {
                 throw new XPathUnsupportedException(xpath);
             } else {
+                if (evalContext.isConstraint) {
+                    String contextMessage = String.format(" Question with id: \'%s\' has issue with condition expression: \'%s\'",
+                            evalContext.getContextRef().getNameLast(), expr);
```
shubham1g5: @ctsims Do you know if `evalContext.getContextRef().getNameLast()` will always give us the question id here ? If not, I think we should skip showing question id in the message.

src/main/java/org/javarosa/xpath/XPathConditional.java
rristovic: Hmm on the second thoughts, I there is no specific need. I will remove that check

src/main/java/org/javarosa/xpath/XPathConditional.java
```diff
@@ -51,6 +51,10 @@ public Object evalRaw(DataInstance model, EvaluationContext evalContext) {
             if (xpath != null) {
                 throw new XPathUnsupportedException(xpath);
             } else {
+                String contextMessage = String.format(" Question with id: \'%s\' has issue with condition expression: \'%s\'",
+                        evalContext.getContextRef().getNameLast(), expr);
```
ctsims: Couple of small notes:

1) It's preferable to use the full context ref for these, rather than drilling down to the last qname. /data/something/myquestion is the easiest way to track that. 

2) ContextRef can be a somewhat sketchy indicator of where an conditional is calculating, since we simplify the graph of triggerables, so I'd probably structure the message such that the XPathExpression itself is the first thing the user sees, since it'll be more canonically useful, and I might be a bit less concrete in the message, ie: ("Error calculating expression: "expression", being calculated for "context")

3) With user facing XPathExpressions, it's good to use toPrettyString() to format the output if the raw text isn't available.

4) "Conditional" is a useful concept for developer facing messages, but will be confusing for end users. If it's known whether an expression is a "calculate" or a "Display Condition", it's fine to use those two words, but otherwise best to leave out what type of expression was having an error

5) For tickets like this it's very helpful on review to add what the Replication Condition now shows as a result of the change, can you add that to the PR summary page?

rristovic: @ctsims  toPrettyString() can't be used because there could be an exception during parse boolean expression, this was the case with test expression in issue reproducing app.

ctsims:
Hi @rristovic 

I made a proposed update that will allow the engine to prettyprint out xpath exprs for common xpath errors that result in things we can't process at runtime (but are valid xpath constructs)

You can view by commit here:
https://github.com/dimagi/commcare-core/pull/848/commits/acb48110dac5b8d532c57fc6024c6a2cabe78e78

It should let you prettyprint out the expression string in a way that's productive for debugging.

I'm not super thrilled by adding another eval-ish path to the code, but these types of user errors are super common and incredibly time consuming to track down so I think it's worth the hit.

rristovic:
Thanks @ctsims,
Now it is work, I have updated the code to use toPrettyString() and screenshot in PR

ctsims:
Looks good! thanks for the second look

ctsims:

PR #850, opened 2019-04-07
Clears app installation files in case of a failure during installation

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-98

### DIFF
--- a/src/main/java/org/commcare/resources/ResourceManager.java
+++ b/src/main/java/org/commcare/resources/ResourceManager.java
@@ -52,25 +52,31 @@ public class ResourceManager {
             UnresolvedResourceException,
             InstallCancelledException {
         synchronized (platform) {
-            if (!global.isReady()) {
-                global.prepareResources(null, platform, forceInstall);
-            }
+            try {
+                if (!global.isReady()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/resources/ResourceManager.java
```diff
@@ -52,25 +52,31 @@ public static void installAppResources(CommCarePlatform platform, String profile
...
+                    Resource r = new Resource(Resource.RESOURCE_VERSION_UNKNOWN,
+                            CommCarePlatform.APP_PROFILE_RESOURCE_ID,
+                            locations, "Application Descriptor");
+
+                    global.addResource(r, global.getInstallers().getProfileInstaller(forceInstall), "");
+                    global.prepareResources(null, platform, forceInstall);
+                }
+            } catch (Exception e) {
+                // Clear all resources in case of a failure
+                global.clearAll(platform);
```
shubham1g5: This is the only change in PR.

shubham1g5:
Closing this PR since we need an alternative approach to be able to resume installations mid-way.

PR #851, opened 2019-04-18
Fix for Short Cycle Detection Algorithm

ctsims:
Ticket Reference:
https://dimagi-dev.atlassian.net/browse/HI-572

Will implemented an algorithm to find the shortest cycle in dependency evaluation, but the directional of the graph was implemented backwards

IE: For a calculation
velocity = distance / time

it would create the dependency edge set
distance -> velocity
time -> velocity

rather than
velocity -> time
velocity -> distance

which was making the algorithm mistakenly miss the cycle and error out.

Also ensured that cyclical references are reported specifically as a parse error, which was not unified on the current reporting path. We can end up detecting them in two places.

### DIFF
--- a/src/test/resources/xform_tests/real_form_with_cycle_errors.xml
+++ b/src/test/resources/xform_tests/real_form_with_cycle_errors.xml
@@ -1,0 +1,7060 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">
+	<h:head>
+		<h:title>Registo e Testagem da Família</h:title>
+		<model>
+			<instance>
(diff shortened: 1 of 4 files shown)


src/test/resources/xform_tests/real_form_with_cycle_errors.xml
```diff
@@ -0,0 +1,7060 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">
```
shubham1g5: I wonder if it's possible to use a smaller form for the test case that can repro the issue. Current one is 7k lines of xml which is not ideal.

ctsims: Hey, agreed that this seems like a big form. It also adds more time to tests than I'm comfortable with, but I couldn't think of a way to specifically clean out only these question elements and retain the "Regression." 

I'm not sure if it actually matters much to have the regression test for the future if it ends up causing issues. i'll mull it over  a bit

PR #855, opened 2019-06-11
Adds back removal of all upgrade table entries when clearing upgrade on failures

shubham1g5:
This was introduced in the [Recovery Measures PR](https://github.com/dimagi/commcare-core/pull/803). Before this PR, we had a `ResourceTable.clear` function which we were using to clear the upgrade table on failures or when an update gets staled. The Recovery measure PR refactored this `clear` to  be called `clearUpgrade` and utilize internal method `clearByStatus`. During this refactoring I mistakenly dropped `storage.removeAll` from the code path which will today result in update table not getting cleared by this function. 

This happened because of me getting confused by [the git diff](https://github.com/dimagi/commcare-core/pull/803/commits/892d642425519e31ee18f671078d05f7ec4df95c) which made it seem like my commit added the `storage.removeAll` statement when it was already there leading me to believe that I added that statement by mistake and hence removing it in a later commit in that PR. 

Also this bug has now been there in both 2.45 and 2.46 and I can't really comprehend any visible repercussions of this issue. We call `clear` after update failures or as a result of staged update getting stale. Though we also have another method `ResourceTable.destroy` that does clears the udpate table today and gets [called when an stages update has gone stale](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/engine/resource/AndroidResourceManager.java#L111) and also [when we have a new update available on top of the staged update](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/engine/resource/AndroidResourceManager.java#L141).  

I think we will need to make this fix available in next LTS release we do since LTS currenly also doesn't have this bug and risks are substantially higher with ICDS. Though I am not really sure if we should do a hotfix for it on 2.46 and wondering if you have any thoughts on that.

### DIFF
--- a/src/main/java/org/commcare/resources/model/ResourceTable.java
+++ b/src/main/java/org/commcare/resources/model/ResourceTable.java
@@ -872,14 +872,13 @@ public class ResourceTable {
     // Clears all resources in the table
     public void clearAll(CommCarePlatform platform) {
         clearByStatus(platform, RESOURCE_STATUS_ALL_RESOURCES);
-        storage.removeAll();
     }
 
     /**
      * Clears any resources with a given resource status and also try very hard to remove any files installed
      * by it. This is important for rolling back botched upgrades without leaving their files around.
      *
-     * @param platform CommCare platform
+     * @param platform       CommCare platform
      * @param resourceStatus Only resources with this status will get cleared
      */
     private void clearByStatus(CommCarePlatform platform, int resourceStatus) {
@@ -900,6 +899,8 @@ public class ResourceTable {
         if (count > 0) {
             Logger.log(LogTypes.TYPE_RESOURCES, "Cleaned up " + count + " records from table");
         }
+
+        storage.removeAll();
     }
 
     protected void cleanup() {
--- a/src/main/java/org/commcare/resources/model/ResourceTable.java
+++ b/src/main/java/org/commcare/resources/model/ResourceTable.java
@@ -864,24 +864,26 @@ public class ResourceTable {
     }
 
 
-    // Clears resources with status RESOURCE_STATUS_UPGRADE in the table
+    // Uninstalls resources with status RESOURCE_STATUS_UPGRADE in the table and wipe the complete table
     public void clearUpgrade(CommCarePlatform platform) {
-        clearByStatus(platform, Resource.RESOURCE_STATUS_UPGRADE);
+        uninstallResourcesForStatus(platform, Resource.RESOURCE_STATUS_UPGRADE);
+        storage.removeAll();
     }
 
-    // Clears all resources in the table
+    // Clears all resources in the table and wipe the complete table
     public void clearAll(CommCarePlatform platform) {
-        clearByStatus(platform, RESOURCE_STATUS_ALL_RESOURCES);
+        uninstallResourcesForStatus(platform, RESOURCE_STATUS_ALL_RESOURCES);
+        storage.removeAll();
     }
 
     /**
-     * Clears any resources with a given resource status and also try very hard to remove any files installed
+     * Uninstalls any resources with a given resource status and also try very hard to remove any files installed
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -872,14 +872,13 @@ public void clearUpgrade(CommCarePlatform platform) {
     // Clears all resources in the table
     public void clearAll(CommCarePlatform platform) {
         clearByStatus(platform, RESOURCE_STATUS_ALL_RESOURCES);
-        storage.removeAll();
```
shubham1g5: Moving this to `clearByStatus` since both `clearAll` and `clearUpgrade` calls `clearByStatus` and we need `storage.removeAll` in code paths for both `clearAll` and `clearUpgrade`.

src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -900,6 +899,8 @@ private void clearByStatus(CommCarePlatform platform, int resourceStatus) {
         if (count > 0) {
             Logger.log(LogTypes.TYPE_RESOURCES, "Cleaned up " + count + " records from table");
         }
+
+        storage.removeAll();
```
ctsims: This should only wipe removed resources here, right, not everything?

src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -864,25 +864,26 @@ public void destroy() {
     }
 
 
-    // Clears resources with status RESOURCE_STATUS_UPGRADE in the table
+    // Uninstalls resources with status RESOURCE_STATUS_UPGRADE in the table and wipe the complete table
     public void clearUpgrade(CommCarePlatform platform) {
-        clearByStatus(platform, Resource.RESOURCE_STATUS_UPGRADE);
+        uninstallResourcesForStatus(platform, Resource.RESOURCE_STATUS_UPGRADE);
+        storage.removeAll();
```
shubham1g5: Line 870 is the only meaningful change in PR.

shubham1g5:
@ctsims I tried out some scenarios to repro errors that may be caused by this. It appears that in cases when update gets staled, this issue doesn't surface since we are [calling `destory` ](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/engine/resource/AndroidResourceManager.java#L111) early in the update process when an update gets staled that takes care of removing the table entries. 

However I was able to repro the error using this flow - 
1. An update gets failed while download say due to sluggy network
2. User rechecks for the update and this time the download succeeds and the update becomes available to install
3. An error occurs while installing the update

In terms of why this happens, after update failure in step1 we call [clearUpgrade](https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/engine/resource/AndroidResourceManager.java#L208) which results in the respective update files for installed resources in the upgrade table to be cleared. This because of the bug doesn't clear resource entries itself from the table. At this point the upgrade table has some resource entries with status `UPGRADE` and some with "UNITIALIZED" status depending on at which stage the update failed. When the user rechecks for the update, update process checks for all uninitialized resources in the upgrade table and resumes downloading/staging them without validating the already installed resources which have been uninstalled by `clearUpgrade`. Therefore this update check goes on to be successful without any errors and later errors when we apply/install the update. 

It's hard to tell from  [Sumo with the corresponding failure message](https://service.us2.sumologic.com/ui/#/search/RUbVquoA7YUN4YKDvz1q9ike2E5uw2AWaIGyqbe9) whether this bug is causing errors frequently since any network error would also result in the same error message. 

Though I think there are high chances of this bug causing issues in update specially with network failures frequently resulting in update failures. Therefore think we should hotfix it.

ctsims:
Looks good after our conversation. Thanks for going through the refactor of the framing and names!

PR #857, opened 2019-07-04
Skips case deletion in case the case has already been deleted

shubham1g5:
Sentry: https://sentry.io/organizations/dimagi/issues/989338130/?project=142105&referrer=slack
Jira: https://dimagi-dev.atlassian.net/browse/MOB-116

It seems like when purging the cases, if somehow a case appears in the list of to be deleted cases twice, mobile tries to delete that case twice and encounters a Null Pointer exception in the process because of that case being already deleted.

In this particular case mobile is trying to delete the case https://www.commcarehq.org/a/broad-nigeria/reports/case_data/76f90a6d-8c97-4c39-8bf2-713ce2daa272/#related since the parent of this(caseid: 157127a9-9782-4cbe-a5ad-0daf494f4461) has been deleted. Though this case also has an extension case https://www.commcarehq.org/a/broad-nigeria/reports/case_data/8b9da4db-4808-4015-97b7-900b43b44d06/ with the common parent which causes "76f90a6d-8c97-4c39-8bf2-713ce2daa272" to appear twice in the list for to be deleted cases.

### DIFF
--- a/src/main/java/org/commcare/cases/util/CasePurgeFilter.java
+++ b/src/main/java/org/commcare/cases/util/CasePurgeFilter.java
@@ -111,10 +111,10 @@ public class CasePurgeFilter extends EntityFilter<Case> {
      *
      * The Index of each node is its case id guid
      * The Node element for the guide consists of an integer array [STATUS_FLAGS, storageid]
-     *         storageid is the id of the row of the case in the provided storage
-     *         STATUS_FLAGS is a bitmasked integer. It should be initialized with any of STATUS_OWNED,
-     *              STATUS_OPEN, and/or STATUS_RELEVANT as applies to the node (A node starts as relevant
-     *              if it is OWNED and it is OPEN)
+     * storageid is the id of the row of the case in the provided storage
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/cases/util/CasePurgeFilter.java
```diff
@@ -331,9 +331,11 @@ private void removeNodeAndPropagate(String indexOfRemovedNode,
 
         // Once all edges to/from this node have been removed, delete the node itself from the
         // DAG, and add it to the list of cases to be purged
-        int storageIdOfRemovedNode = internalCaseDAG.removeNode(indexOfRemovedNode)[1];
-        idsToRemove.addElement(new Integer(storageIdOfRemovedNode));
-        casesRemovedDueToMissingCases.addElement(indexOfRemovedNode);
+        if (casesRemovedDueToMissingCases.indexOf(indexOfRemovedNode) == -1) {
```
shubham1g5: This check is the only change in PR.

PR #861, opened 2019-07-16
Fixes bulk case parser to not overwrite same case blocks in a single Form

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/MOB-120

When there is a form containing the save to case question causing the form to have 2 case blocks, the later case block overwrites the earlier case block from the form.  

Bulk parser currently does do a lookup in `currentOperatingSet` to see if a case already exists but any cases we create during a form parsing doesn't really make it to `currentOperatingSet`. This PR changes that to write any cases we are creating or updating in `currentOperatingSet`. 

This bug is only there for BulkCaseXmlParser and not CaseXmlParser since CaseXmlParser directly looks up the case DB before creating a case.


ctsims:
Change looks good. I think after the hotfix it would be good to get a replicating test into core.

PR #862, opened 2019-07-16
[HOTFIX] Fixes bulk case parser to not overwrite same case blocks in a single Form

shubham1g5:
Corresponding master PR: https://github.com/dimagi/commcare-core/pull/861/


shubham1g5:
Self merging since the master PR is approved.

PR #871, opened 2019-11-12
Adds and Parses buildProfileId for Profile

shubham1g5:
Parses `buildProfileId` present in `profile` file as  - 

````
<profile version="226"
         update="http://192.168.20.109/a/mkangia-domain/apps/download/e0a3716cec574efdb3055e8391f071f2/media_profile.ccpr?latest=true&amp;profile=1bda6672420745a8837eaf03f6d0532d"
         requiredMajor="2"
         requiredMinor="44"
         uniqueid="2588b449b7eae879341d44f33b015389"
         name="MK-Application"
         descriptor="Profile File"
         buildProfileID="1bda6672420745a8837eaf03f6d0532d">
````

cross-request: https://github.com/dimagi/commcare-android/pull/2291

### DIFF
--- a/src/main/java/org/commcare/suite/model/Profile.java
+++ b/src/main/java/org/commcare/suite/model/Profile.java
@@ -30,7 +30,6 @@ public class Profile implements Persistable {
 
     public static final String STORAGE_KEY = "PROFILE";
     public static final String FEATURE_REVIEW = "checkoff";
-    public static final String FEATURE_USERS = "users";
 
     private int recordId = -1;
     private int version;
@@ -41,6 +40,7 @@ public class Profile implements Persistable {
 
     private String uniqueId;
     private String displayName;
+    private String buildProfileId;
 
     /**
      * Indicates if this was generated from an old version of the profile file, before fields
@@ -66,16 +66,16 @@ public class Profile implements Persistable {
      *                is obsoleted by it.
      */
     public Profile(int version, String authRef, String uniqueId, String displayName,
-                   boolean fromOld) {
+                   boolean fromOld, String buildProfileId) {
         this.version = version;
         this.authRef = authRef;
         this.uniqueId = uniqueId;
         this.displayName = displayName;
+        this.buildProfileId = buildProfileId;
         this.fromOld = fromOld;
         properties = new Vector<>();
         roots = new Vector<>();
         featureStatus = new Hashtable<>();
-
         //turn on default features
         featureStatus.put("users", true);
     }
@@ -106,6 +106,13 @@ public class Profile implements Persistable {
 	public String getDisplayName() {
 	    return this.displayName;
 	}
+
+    /**
+     * @return the buildProfileId for this particular app profile
+     */
+    public String getBuildProfileId() {
+        return buildProfileId;
+    }
 
     /**
      * @return if this object was generated from an old version of the profile.ccpr file
@@ -208,6 +215,7 @@ public class Profile implements Persistable {
         properties = (Vector<PropertySetter>)ExtUtil.read(in, new ExtWrapList(PropertySetter.class), pf);
         roots = (Vector<RootTranslator>)ExtUtil.read(in, new ExtWrapList(RootTranslator.class), pf);
         featureStatus = (Hashtable<String, Boolean>)ExtUtil.read(in, new ExtWrapMap(String.class, Boolean.class), pf);
+        buildProfileId = ExtUtil.readString(in);
     }
 
     @Override
@@ -222,5 +230,6 @@ public class Profile implements Persistable {
         ExtUtil.write(out, new ExtWrapList(properties));
         ExtUtil.write(out, new ExtWrapList(roots));
         ExtUtil.write(out, new ExtWrapMap(featureStatus));
+        ExtUtil.writeString(out, buildProfileId);
     }
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/org/commcare/suite/model/Profile.java [hunk 1]
```diff
@@ -208,6 +215,7 @@ public void readExternal(DataInputStream in, PrototypeFactory pf)
         properties = (Vector<PropertySetter>)ExtUtil.read(in, new ExtWrapList(PropertySetter.class), pf);
         roots = (Vector<RootTranslator>)ExtUtil.read(in, new ExtWrapList(RootTranslator.class), pf);
         featureStatus = (Hashtable<String, Boolean>)ExtUtil.read(in, new ExtWrapMap(String.class, Boolean.class), pf);
+        buildProfileId = ExtUtil.readString(in);
```
shubham1g5: @ctsims  I didn't find any particular code that serializes Profile so I think we don't need to worry about any sort of migration from old Profile to new Profile. Let me know if you think that's not right.

ctsims: I think that makes sense on commcare-android, since we read/write the profile from disk as the raw file. On J2ME I believe we stored the serialized version instead.

Can you check the `formplayer` process, though? I was surprised to see that a number of formplayer resources which I figured were stored and read as xml files were instead using the serialized versions. If so, all we need to do is bump the application database version on formplayer to force a full reinstall.

shubham1g5:
@ctsims This is driven from the [change request](https://dimagi-dev.atlassian.net/browse/ICDS-909) where we want to make `buildProfileId` visible in App Status report to know what profile each user's is using. AFAIK there is no spec doc for this feature unfortunately and the conversation is limited to the JIRA ticket (Tagging @mkangia to confirm)

Also ideally such kind of information should be part of heartbeat requests but the sense I got from the conversation on ticket is it depends on Warehouse implementation on HQ side which is not in proper shape right now. Though I guess it would still be good idea to include this info with Heartbeat as well in case we decide to make a switch to warehouse in future ?

ctsims:
@shubham1g5 

Maybe we can take to voice on Thursday, a bunch of stuff is coming to mind but not all related

1) I think we should create a spec for a change like this that can be shared between the teams to make sure we are aligned on what is being kept track of where. Once decided, the specs should also be update in HQ (in the Form Submission API) and on our side.
2) After the HTTP Request Params issue with user id being included in updates Simon's position (which I agree with) is that [HTTP Headers are a better fit](https://docs.google.com/document/d/1h3v0w8JvEuQaatSF_AtXrMyyxRIPAZwJHVJyG-U6Xdg/edit?ts=5d83b801#bookmark=id.uaxawd5fhui6) for any web request where the additional information does not affect **the request itself.** IE: The server won't handle this POST differently based on the information, it would just use it internally to update other info, so we should put it in a header rather than web params.
3) I think the HQ team is once again forgetting that the uptake curve before this change will be on real devices is Months and Months from now. Before this gets onto a phone there needs to be a full mobile release (months away), followed by a roll-out curve (which takes months to saturate), so the urgency on this can't be high and there's no need to rush
4) I'm finding it pretty concerning how much duplication we're getting across the heartbeat / form submission channels, and given my previous point (3) I don't think there's an excuse for the web team to insist that we put this data on one channel because it's having less challenges than another. It's already hard to keep track of what the 'canonical' versions are for the app/install, and I don't want to make it even more confusing.  

I think parsing out the profile version and making sure its sent to HQ is a good plan, but I'd strongly favor including it in the heartbeat over the form submissions if its purpose is to clarify what is on device, since that's the fundamental purpose of the heartbeat request.

shubham1g5:
@ctsims 
Hey, we did miss a step here in the process and this instance is a good reminder for me of why specs are important. Personally, I think I made the mistake of thinking of it as a very minor mobile change needed to support bigger HQ change and something which would not require any spec doc on mobile side. But in hindshight agree that having a spec and having it reviewed first would have been a much better call and thanks a lot for emphasizing the requirements of a spec here. 

On Points #3 and #4, Given the mobile release timelines, agree that it can't be a high priority item. Though the warehouse work on HQ that will allow HQ to process the parameter in Heartbeat is not really planned currently and there is no clarity whether it will ever be implemented on ICDS (@mkangia please correct me if I am wrong).  Given that, it can be a very long time before we have proper infrastructure on HQ to support this change further pushing out this feature aside from time taken for mobile release uptake. I do think it makes sense to delay on this untill the next LTS release candidate is in sight but if things remain same on HQ even then, I think including the paramenter in both Heartbeat and Form submission would be the right middle ground.

mkangia:
@ctsims @shubham1g5 
> I think the HQ team is once again forgetting that the uptake curve before this change will be on real devices is Months and Months from now

We know that the rollout of this would be gradual on the field and hence this information will start showing as and when the users start to update to the required CommCare version. HQ would just use the information if its present in the request header on form submission (considering we implement the changes for form submission right now). I agree this is not an "urgent" request and was given time to think and implement.

Simon did suggest to make this change for both heartbeat and form submissions.
are you suggesting we just use for heartbeat?
I don't mind working on the changes to heartbeat tracking if that is what is needed to be added or to replace tracking on form submissions.

shubham1g5:
Closing this since it's no longer required.

shubham1g5:
@ctsims  the need to parse the `buildProfileId` has come up again to be able to tag the analytics event with profile information and to get analytics data grouped by different profiles. I have reopened this PR again, it was already approved earlier though wanted to check if the design makes sense for the new use case.

src/main/java/org/commcare/suite/model/Profile.java
shubham1g5: confirming that this will require a formplayer app version dump.

PR #874, opened 2019-12-24
Remove catching all exceptions while localizing text

ShivamPokhriyal:
This PR will remove catching all the exception in the localizeText method so that we can handle it outside and show appropriate error message to user when some exception occurs.

cross-request: https://github.com/dimagi/commcare-android/pull/2159

### DIFF
--- a/src/main/java/org/javarosa/form/api/FormEntryPrompt.java
+++ b/src/main/java/org/javarosa/form/api/FormEntryPrompt.java
@@ -315,6 +315,7 @@ public class FormEntryPrompt extends FormEntryCaption {
     /**
      * Helper for getHintText, getHelpText, getConstraintText. Tries to localize text form textID,
      * falls back to innerText if not available.
+     * It may throw XPathException.
      */
     private String localizeText(QuestionString mQuestionString) {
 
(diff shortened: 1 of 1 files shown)


ShivamPokhriyal:
@damagatchi retest please

src/main/java/org/javarosa/form/api/FormEntryPrompt.java
```diff
@@ -315,6 +316,7 @@ public String getHelpText() {
     /**
      * Helper for getHintText, getHelpText, getConstraintText. Tries to localize text form textID,
      * falls back to innerText if not available.
+     * It may throw XPathTypeMismatchException.
```
shubham1g5: we should just say `XPathException`  here instead of the more specific `XPathTypeMismatchException` since there are cases when other kinds of XPath errors might result in different XPathExceptions from this method.

src/main/java/org/javarosa/form/api/FormEntryPrompt.java
```diff
@@ -21,6 +21,7 @@
 import org.javarosa.core.util.NoLocalizedTextException;
 import org.javarosa.core.util.UnregisteredLocaleException;
 import org.javarosa.xform.parse.XFormParser;
+import org.javarosa.xpath.XPathTypeMismatchException;
```
shubham1g5: is this getting used ?

PR #875, opened 2020-01-10
Add Locale Arg Parameters for <text> implementation

ctsims:
The `<text>` element [specification for the suite](https://github.com/dimagi/commcare-core/wiki/Suite20#text) and app config layer specifies the ability to insert `<argument>` parameters as usually supported by text (like "There are ${0} results of ${1}"), but apparently that was never implemented.

Additionally, there were almost no unit tests for the core functioning of the `<text>` system, so I added some which parse through a sample app to make sure they are maximally realistic.

### DIFF
--- a/src/main/java/org/commcare/suite/model/Text.java
+++ b/src/main/java/org/commcare/suite/model/Text.java
@@ -21,10 +21,13 @@ import org.javarosa.xpath.parser.XPathSyntaxException;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.Collections;
 import java.util.Date;
 import java.util.Enumeration;
 import java.util.Hashtable;
+import java.util.List;
 import java.util.Vector;
 import java.util.concurrent.Callable;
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/commcare/suite/model/Text.java
```diff
@@ -302,6 +334,46 @@ public boolean rawArgs() {
...
+        List<String> keys = new ArrayList<>();
+
+        for(String key : arguments.keySet()) {
+            if(key.equals("id")) {
+                continue;
+            }
+            keys.add(key);
+        }
+
+        //We presume all of the keys here are numbers and that they are in order (implicit in
```
shubham1g5: Looks like this comment was meant to be for L373 below ?

src/test/java/org/commcare/backend/model/AppConfiguredTextTests.java
```diff
@@ -0,0 +1,122 @@
...
+        Assert.assertEquals("ValueArgument", display.getDisplayText(evaluationContext));
+
+        Localization.setLocale("hin");
+
+        Assert.assertEquals("ArgumentValue", display.getDisplayText(evaluationContext));
+
+    }
+
+    @Test
+    public void testLocaliationIdParam() {
```
shubham1g5: spelling

src/main/java/org/commcare/suite/model/Text.java
ctsims: not quite, I'm explaining why we error out if there are more than 10 keys, but not clearly enough. Will update.

ctsims:
@shubham1g5 Thanks for the feedback! Updated based on your inputs.

PR #878, opened 2020-02-06
Use a custom KXmlSerializer to support non BMP characters

ShivamPokhriyal:
This adds support for writing surrogate pairs as a codepoint in KXmlSerializer. I've tested emojis and some other mathematical symbols which now work pretty well with this change. 

[This](https://github.com/dimagi/commcare-core/compare/support-non-bmp-characters?expand=1#diff-1c222c6933e09dcf545e6e6d24e6f305R128-R139) is the patch.

More details here: https://dimagi-dev.atlassian.net/browse/SAASP-10290


src/main/java/org/javarosa/xform/util/CommCareKXmlSerializer.java
```diff
@@ -0,0 +1,565 @@
...
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+
+/**
+ * @author $|-|!˅@M
+ */
+public class CommCareKXmlSerializer implements XmlSerializer {
```
shubham1g5: NitPick: Can we use the AOSP version as it is from [here](https://android.googlesource.com/platform/libcore/+/refs/heads/master/xml/src/main/java/com/android/org/kxml2/io/KXmlSerializer.java) and make a comment in the file that we are using the AOSP version form above link. This will just allow us to copy cat the changes time to time without worrying about merging different code parts.

ctsims:
Hey, 

So I'm quite uncomfortable with the copy+and+patch solution to this problem since it leaves the code in a really sketchy state vis-a-vis the dependency on the parser and lack of clarity on versioning.

I noticed that [kxml.org now redirects to this github repo](https://github.com/stefanhaustein/kxml2) which appears to be a new/forked version of the released libs with tests and proper version history from the original author. The Author makes explicit mention of using [a service called jitpack](https://jitpack.io/) for Gradle Dependency Management. Can we try to see wether building with the jitpack sourced 2.4.2 release of kxml also fixes the issue?

ctsims:
@ShivamPokhriyal - this looks great. Can we add a test which replicates the old failure so we don't end up with a regression?

Also, thanks for notifying that the 2.4.2 build was failing on the other repo and doing all of the other Open Source niceties. Those kinds of steps can take time to pay off, but often help put us in a better position in the future.

src/test/java/org/javarosa/xform/util/test/XmlSerializerTests.java
```diff
@@ -0,0 +1,50 @@
...
+            assertTrue(payload.getPayloadType() == IDataPayload.PAYLOAD_TYPE_XML);
+            ByteArrayInputStream is = (ByteArrayInputStream)payload.getPayloadStream();
+            try {
+                // Check if we are able to parse the serialized xml.
+                // If all the special characters are converted to their corresponding codepoint this would pass.
+                DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
+                DocumentBuilder dBuilder = dbFactory.newDocumentBuilder();
+                dBuilder.parse(is);
+            } catch (Exception e) {
+                fail(e.getMessage());
```
ctsims: One small nitpick: e.getMessage() in this context won't actually contain the stack trace, and since we don't know what kind of exception it is, it might be less helpful that the trace itself. I'd either dump the trace before this call and/or use a failure mode which makes sure it gets dumped.

PR #881, opened 2020-03-09
Library updates

shubham1g5:
Getting some lib updates in as it's start of a new release cycle and okhttp seems to have [released](https://github.com/square/okhttp/blob/master/CHANGELOG.md) some major performance changes in past releases.


ctsims:
Awesome, thanks!

PR #882, opened 2020-03-23
Throw exception when userid is null for a case

ShivamPokhriyal:
https://dimagi-dev.atlassian.net/browse/SAAS-10736
https://dimagi-dev.atlassian.net/browse/SAAS-10735

### DIFF
--- a/src/main/java/org/commcare/xml/CaseXmlParser.java
+++ b/src/main/java/org/commcare/xml/CaseXmlParser.java
@@ -73,6 +73,8 @@ public class CaseXmlParser extends TransactionParser<Case> {
         }
         Date modified = DateUtils.parseDateTime(dateModified);
 
+        String userId = parser.getAttributeValue(null, "user_id");
+
         Case caseForBlock = null;
         boolean isCreateOrUpdate = false;
 
@@ -80,7 +82,7 @@ public class CaseXmlParser extends TransactionParser<Case> {
             String action = parser.getName().toLowerCase();
             switch (action) {
                 case "create":
-                    caseForBlock = createCase(caseId, modified);
+                    caseForBlock = createCase(caseId, modified, userId);
                     isCreateOrUpdate = true;
                     break;
                 case "update":
@@ -122,7 +124,7 @@ public class CaseXmlParser extends TransactionParser<Case> {
         return null;
     }
 
-    private Case createCase(String caseId, Date modified) throws InvalidStructureException, IOException, XmlPullParserException {
+    private Case createCase(String caseId, Date modified, String userId) throws InvalidStructureException, IOException, XmlPullParserException {
         String[] data = new String[3];
         Case caseForBlock = null;
 
@@ -165,6 +167,8 @@ public class CaseXmlParser extends TransactionParser<Case> {
 
         if (data[1] != null) {
             caseForBlock.setUserId(data[1]);
+        } else {
+            caseForBlock.setUserId(userId);
         }
         return caseForBlock;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/xml/CaseXmlParser.java
```diff
@@ -187,6 +187,10 @@ private void updateCase(Case caseForBlock, String caseId) throws InvalidStructur
                 case "owner_id":
                     String oldUserId = caseForBlock.getUserId();
 
+                    if (oldUserId == null) {
+                        throw InvalidStructureException.readableInvalidStructureException("The case_id field of a <case> was found null", parser);
```
shubham1g5: In the error message `userid` should be null instead of case_id ? 

Also in the jira issue, seems like `owner_id` was not defined while from the error stack it seems like `userid` is getting set to null. So I am a bit confused what exactly is happening.

ShivamPokhriyal: Yeah, I'm also not sure how owner_id is different from userid. https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/cases/model/Case.java#L209

shubham1g5:
@ShivamPokhriyal  Would you be able to open this PR against 2.48 hotfix branch

src/main/java/org/commcare/xml/CaseXmlParser.java
shubham1g5: Yeah [seems like](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/xml/CaseXmlParser.java#L167) they are getting used interchangeably on phone with the assumption that case creator is always the owner of the case.  I think we should be doing this `null` check at the [case create](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/xml/CaseXmlParser.java#L167) part itself and throw the error from there.

ShivamPokhriyal: Cuz of [this condition](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/xml/CaseXmlParser.java#L146) It seems like `owner_id` is left out deliberately?

shubham1g5: Hmm, I am not really sure. Lets ask @ctsims

ctsims:
Apparently according to [our spec](https://github.com/dimagi/commcare-core/wiki/casexml20) `owner_id` is (or was) not actually required. Presumably we set it to the current user if it isn't defined.

I'm surprised but this, but presumably HQ has already been using this spec for a while, so I'm not sure we want to change it.

Did the behavior change on the mobile side such that we now crash if owner id is null?

ctsims:
Oh, I see, the issue isn't *no* owner Id, right? It's that the `owner_id` element exists, but it's blank? That is definitely a bit of a different issue (since instead of letting the owner id default to the transacting user, that would imply that the owner id is force set to nothing)

ctsims:
Based on my interpretation above, I think the right behavior for

`caseForBlock.getUserId()`

is for the parser to throw an error earlier (in the parse phase) if `<owner_id>` exists and is blank, and for `caseForBlock.getUserId()` to default to the user ID of the user in the `<case>` transaction block if no `<owner_id>` element exists.

ShivamPokhriyal:
@ctsims 
>> Did the behavior change on the mobile side such that we now crash if owner id is null?
No. It should've always crashed when there was no owner-id. 

>>Oh, I see, the issue isn't no owner Id, right? It's that the owner_id element exists, but it's blank?
The issue is owner-id element doesn't exist in `<create>` element.

src/main/java/org/commcare/xml/CaseXmlParser.java
```diff
@@ -187,6 +191,10 @@ private void updateCase(Case caseForBlock, String caseId) throws InvalidStructur
                 case "owner_id":
                     String oldUserId = caseForBlock.getUserId();
 
+                    if (oldUserId == null) {
```
shubham1g5: To fail as early as possible, we should move this error handling to create case block iself after we set userId

PR #888, opened 2020-04-03
Captive Portal Redirect Exception Handling

ctsims:
@ShivamPokhriyal 

This code provides a super simple entry point for adding code which detects a captive portal and will ensure that the code is executed at the most effective time in a request cycle based on our HTTP traffic. 

This will ensure that the ultimate downstream IO Exception from a request which fails due to a captive portal is semantically typed for easy handling

### DIFF
--- a/src/main/java/org/commcare/core/network/ModernHttpRequester.java
+++ b/src/main/java/org/commcare/core/network/ModernHttpRequester.java
@@ -15,6 +15,8 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
+import javax.net.ssl.SSLHandshakeException;
+import javax.net.ssl.SSLPeerUnverifiedException;
 
 import okhttp3.FormBody;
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -112,9 +114,30 @@ public void makeRequestAndProcess() {
...
+                throw new CaptivePortalRedirectException(e);
+            }
+
+            //Otherwise just rethrow the original exception. Probably a certificate issue
+            throw e;
+        }
     }
 
+    private boolean checkCurrentNetworkAsCaptivePortal() {
+        return false;
```
ctsims, line 137: @ShivamPokhriyal This is the entry point for adding however we want to do detection.

I tested this with some [sample code from the android source](https://stackoverflow.com/a/14030276/5290221) and everything worked, but wasn't sure exactly what mechanism we wanted to rely on here.

PR #889, opened 2020-04-06
Captive portal check

ShivamPokhriyal:
On top of #888
cross-request: https://github.com/dimagi/commcare-android/pull/2206

### DIFF
--- a/src/main/java/org/commcare/core/network/ModernHttpRequester.java
+++ b/src/main/java/org/commcare/core/network/ModernHttpRequester.java
@@ -4,7 +4,9 @@ import org.commcare.core.interfaces.HttpResponseProcessor;
 import org.commcare.core.interfaces.ResponseStreamAccessor;
 import org.commcare.core.network.bitcache.BitCache;
 import org.commcare.core.network.bitcache.BitCacheFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -112,9 +116,43 @@ public void makeRequestAndProcess() {
...
     }
 
+    private boolean checkCurrentNetworkAsCaptivePortal() {
+        String captivePortalURL = "http://www.commcarehq.org/serverup.txt";
+        CommCareNetworkService commCareNetworkService =
+                CommCareNetworkServiceGenerator.createNoAuthCommCareNetworkService();
+        try {
+            Response<ResponseBody> response =
+                    commCareNetworkService.makeGetRequest(captivePortalURL, new HashMap<>(), new HashMap<>()).execute();
+            Logger.log(LogTypes.TYPE_WARNING_NETWORK, "Detecting captive portal, Response Code: "
```
shubham1g5, line 145: I think we should remove the log from here and let the catchers of the `CaptivePortalRedirectException` log it in individual processes.

PR #895, opened 2020-05-17
Adds global -> geo-overlay block in Suite's Detail

shubham1g5:
Adds xml definition to define a map overlay for a case list  as follows - 

````
  <global>
      <geo-overlay>
        <coordinates>
          <text>
          </text>
        </coordinates>
        <title>
          <text>
          </text>
        </title>
      </geo-overlay>
    </global>
````
cross-request: https://github.com/dimagi/commcare-android/pull/2237

### CODE REVIEW

File: src/main/java/org/commcare/suite/model/Global.java [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
+        Vector<GeoOverlay> theGeoOverlays = (Vector<GeoOverlay>)ExtUtil.read(in, new ExtWrapList(GeoOverlay.class), pf);
+        geoOverlays = new GeoOverlay[theGeoOverlays.size()];
+        ArrayUtilities.copyIntoArray(theGeoOverlays, geoOverlays);
+    }
+
+    @Override
+    public void writeExternal(DataOutputStream out) throws IOException {
+        ExtUtil.write(out, new ExtWrapList(ArrayUtilities.toVector(geoOverlays)));
```
ctsims, line 42: Does this work if the list is empty/null?

I think there's a good chance we'll be using a global static namespace for detail info quite soon, and we'll want the model to be pretty flexible.

src/main/java/org/commcare/xml/DetailParser.java
```diff
@@ -58,6 +61,16 @@ public Detail parse() throws InvalidStructureException, IOException, XmlPullPars
...
 
         while (nextTagInBlock("detail")) {
+            if (GlobalParser.NAME_GLOBAL.equals(parser.getName().toLowerCase())) {
+                try {
+                    checkNode(GlobalParser.NAME_GLOBAL);
+                    global = new GlobalParser(parser).parse();
+                    parser.nextTag();
+
+                } catch (InvalidStructureException e) {
+                    System.out.println("Lookup block not found " + e);
```
ctsims, line 71: where does this error come from? This catch should almost certainly kill the parse and result in a global InvalidStructureException, so it would be better to not be caught at all

src/main/java/org/commcare/xml/GlobalParser.java
```diff
@@ -0,0 +1,35 @@
...
+
+    static final String NAME_GLOBAL = "global";
+
+    GlobalParser(KXmlParser parser) {
+        super(parser);
+    }
+
+    @Override
+    public Global parse() throws InvalidStructureException, IOException, XmlPullParserException {
+        Vector<GeoOverlay> geoOverlays = new Vector<>();
```
ctsims, line 26: How does this parse go when the global block is empty? It might be easiest for HQ to generate a bunch of empty `<global/>` blocks rather than only inject them when needed, so if we can plan around that it's really helpful.

src/main/java/org/commcare/xml/GeoOverlayParser.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public GeoOverlay parse() throws InvalidStructureException, IOException, XmlPullParserException {
+        DisplayUnit title = null;
+        DisplayUnit coordinates = null;
+        while (nextTagInBlock(NAME_GEO_OVERLAY)) {
+            String tagName = parser.getName().toLowerCase();
+            if (NAME_COORDINATES.contentEquals(tagName)) {
+                nextTagInBlock(NAME_COORDINATES);
+                coordinates = new DisplayUnit(new TextParser(parser).parse());
+            } else if (NAME_TITLE.contentEquals(tagName)) {
```
ctsims, line 34: Nitpick: I'd call this "label", not "title". Title is used elsewhere in the suite with a different meaning, and it might get confusing.

src/main/java/org/commcare/suite/model/Global.java
shubham1g5: `GlobalParser` today always [initialize this list](https://github.com/dimagi/commcare-core/blob/geoOverlay/src/main/java/org/commcare/xml/GlobalParser.java#L26) so that it's never null. Would we still want to provide null safety here ?

src/main/java/org/commcare/xml/DetailParser.java
shubham1g5: yeah makes sense, I kind of copied it from the parsing below for `lookup`. Another thing that doesn't make sense to me is we ae doing a `checkNode(GlobalParser.NAME_GLOBAL);` just after `if (GlobalParser.NAME_GLOBAL.equals(parser.getName().toLowerCase())) {` . Do you know why we are following that pattern ?

src/main/java/org/commcare/xml/GlobalParser.java
shubham1g5: In case of an empty global tag, `Global` will just get initialized with an empty  `geoOverlays` list.

src/main/java/org/commcare/suite/model/Global.java
ctsims: Nope, no need as long as it works (sometimes it's hard to tell with null boundary checks.

It did make me realize, though, that there aren't any sample app / parse tests for all of this. Can you write some up? Nothing complicated, just making sure the code paths get hit, and demonstrating the boundaries.

ctsims:
Thanks for answering questions. Everything looks good, but I do think we should have some basic tests for regression which make sure that our code coverage includes these new primitives.

shubham1g5:
@ctsims Thanks for your suggestion on tests, they helped in catching a bug already. Added couple tests testing the parsing logic, let me know if they seem sufficient.

shubham1g5:
@ctsims sure, implemented it. Ordering is maintained by default as we are using ordered collections everywhere.

shubham1g5:
@ctsims would you be able to re-approve to be able to merge this. Thanks.

PR #899, opened 2020-05-21
Adds lazy resource property

shubham1g5:
Multimedia Changes - 

1. Adds a lazy `Resource` property
2. Better classify `MissingMediaException` using `MissingMediaExceptionType`

cross-request: https://github.com/dimagi/commcare-android/pull/2243

### DIFF
--- a/src/main/java/org/commcare/resources/model/Resource.java
+++ b/src/main/java/org/commcare/resources/model/Resource.java
@@ -111,7 +111,8 @@ public class Resource implements Persistable, IMetaData {
     // Not sure if we want this persisted just yet...
     protected String parent;
 
-    private String descriptor;
+    protected String descriptor;
+    private boolean lazy;
 
     /**
      * For serialization only
@@ -124,19 +125,24 @@ public class Resource implements Persistable, IMetaData {
      * Creates a resource record identifying where a specific version of a resource
      * can be located.
      *
-     * @param version    The version of the resource being defined.
-     * @param id         A unique string identifying the abstract resource
-     * @param locations  A set of locations from which this resource's definition
-     *                   can be retrieved. Note that this vector is copied and should not be changed
-     *                   after being passed in here.
+     * @param version   The version of the resource being defined.
+     * @param id        A unique string identifying the abstract resource
+     * @param locations A set of locations from which this resource's definition
+     *                  can be retrieved. Note that this vector is copied and should not be changed
+     *                  after being passed in here.
      */
-    public Resource(int version, String id, Vector<ResourceLocation> locations, String descriptor) {
+    public Resource(int version, String id, Vector<ResourceLocation> locations, String descriptor, boolean lazy) {
         this.version = version;
         this.id = id;
         this.locations = locations;
         this.guid = PropertyUtils.genGUID(25);
         this.status = RESOURCE_STATUS_UNINITIALIZED;
         this.descriptor = descriptor;
+        this.lazy = lazy;
+    }
+
+    public Resource(int version, String id, Vector<ResourceLocation> locations, String descriptor) {
+        this(version, id, locations, descriptor, false);
     }
 
     /**
@@ -166,6 +172,10 @@ public class Resource implements Persistable, IMetaData {
      */
     public String getRecordGuid() {
         return guid;
+    }
+
+    public void setRecordGuid(String guid) {
+        this.guid = guid;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -1175,4 +1190,9 @@ public void setMissingResources(SizeBoundUniqueVector<Resource> missingResources
     public SizeBoundUniqueVector<Resource> getMissingResources() {
         return mMissingResources;
     }
+
+    public Vector<Resource> getLazyResources() {
+        // todo
```
ShivamPokhriyal, line 1195: What's this todo for?

src/main/java/org/commcare/xml/ResourceParser.java
```diff
@@ -46,6 +47,6 @@ public Resource parse() throws InvalidStructureException, IOException, XmlPullPa
             }
         }
 
-        return new Resource(version, id, locations, descriptor);
+        return new Resource(version, id, locations, descriptor, lazy != null && lazy.contentEquals("true"));
```
ShivamPokhriyal, line 50: `"true".contentEquals(lazy)` can be used to avoid null checks.

shubham1g5: "true".contentEquals(null) does throw a NPE

src/main/java/org/commcare/resources/model/ResourceTable.java
shubham1g5: sorry, left it by mistake

src/main/java/org/commcare/xml/ResourceParser.java
ShivamPokhriyal: Ohh, sorry for that! Just curious why are  we using `contentEquals` over `equals`?

shubham1g5: no particular reason, they both are same for this use case. `contentEquals` is a more free implementation of `equals` allowing to compare across different `CharSequence` implementations so I just default to it for any string comparison.

src/main/java/org/commcare/resources/model/Resource.java
```diff
@@ -111,7 +112,8 @@
     // Not sure if we want this persisted just yet...
     protected String parent;
 
-    private String descriptor;
+    protected String descriptor;
+    private boolean lazy;
```
ctsims, line 116: I think this means we're also going to need to bump the appdb database version in formplayer

src/main/java/org/commcare/resources/model/Resource.java
```diff
@@ -315,13 +332,15 @@ public Object getMetaData(String fieldName) {
                 return parent == null ? "" : parent;
             case META_INDEX_VERSION:
                 return version;
+            case META_INDEX_LAZY:
+                return lazy;
```
ctsims, line 336: I'm a little surprised to see this works with a boolean, btw. My intuition is that we should hard code an explicit string value here, and convert it to/from boolean. We've started making soft assumptions elsewhere that all metadata values are strings, and I'm worried that implicit conversions.

src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -1175,4 +1198,12 @@ public void setMissingResources(SizeBoundUniqueVector<Resource> missingResources
     public SizeBoundUniqueVector<Resource> getMissingResources() {
         return mMissingResources;
     }
+
+    public Vector<Resource> getLazyResources() {
+        return storage.getRecordsForValues(new String[]{Resource.META_INDEX_LAZY}, new Boolean[]{true});
```
ctsims, line 1203: do we use this now that we're mostly iterating on ID's? This feels a bit like a potential bomb call, it'll grab a LOT of records.

src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java
```diff
@@ -207,6 +207,19 @@
...
+     * Retrieves a Vector of Externalizable objects from the storage for which, for each field in fieldNames,
+     * the record has the correct corresponding value in values
+     *
+     * @param metaFieldNames A list of metadata field names to match
+     * @param values     The values which must match the field names provided
+     *
+     * @return A Vector of Externalizable objects e, such that the fields specified are equal to the corresponding values provided.
+     *
+     */
+    Vector<E> getRecordsForValues(String[] metaFieldNames, Object[] values);
```
ctsims, line 221: Is this the same signature as the custom implementations in Android AND formplayer? Is good to finally have it committed in code.

PR #901, opened 2020-05-26
Add PUT method to retrofit network service

ShivamPokhriyal:


src/main/java/org/commcare/core/network/CommCareNetworkService.java
```diff
@@ -37,4 +38,7 @@
     Call<ResponseBody> makePostRequest(@Url String url, @QueryMap Map<String, String> params,
                                        @HeaderMap Map<String, String> headers,
                                        @Body RequestBody body);
+
+    @PUT
+    Call<ResponseBody> makePutRequest(@Url String url, @Body RequestBody body);
```
ShivamPokhriyal, line 43: Probably not the best way here. I can certainly have this method outside in tests directory but then I'll have to modify the retrofit setup code(`CommCareNetworkServiceGenerator`) to use a different class for tests.

PR #904, opened 2020-05-26
Directional IO Exceptions

shubham1g5:
Changes `writeFromInputToOutputNew` to throw Directional IO Exceptions

### DIFF
--- a/src/main/java/org/javarosa/core/io/StreamsUtil.java
+++ b/src/main/java/org/javarosa/core/io/StreamsUtil.java
@@ -146,17 +146,20 @@ public class StreamsUtil {
                 if (observer != null) {
                     observer.notifyCurrentCount(counter);
                 }
-
-                try {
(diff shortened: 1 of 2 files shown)


src/main/java/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -118,33 +120,54 @@ public static void writeFromInputToOutputUnmanaged(InputStream is, OutputStream
...
             while (count != -1) {
                 counter += count;
                 if (observer != null) {
                     observer.notifyCurrentCount(counter);
                 }
-                os.write(buffer, 0, count);
-                count = is.read(buffer);
+
+                try {
+                    os.write(buffer, 0, count);
```
ctsims, line 151: Nit: this line and the line on 143 above have the same structure, but with different patterns to handle them (one externalized, one internalized) which seems odd to me.

PR #905, opened 2020-05-26
Logical Op Indexed Value Lookup Optimization

ctsims:
New optimization for the xpath engine.

Naively identifies the predicate pattern 

`X or Y` or `X and Y` 

when X and Y are both indexed lookups, and performs the lookup by doing both indexed lookups independently against their index, then performing the appropriate action on the results. 

Only implemented on Formplayer initially, since on FormPlayer the indexed lookups are so fast that there are very, very few situations in which this isn't the optimal query path. On Android, it would be important to check the evaluation context first to ensure that the set being evaluated isn't very small.


snopoke (approved):
A few tests would be great.

PR #910, opened 2020-06-09
Normalize search filter before performing search

ShivamPokhriyal:
https://dimagi-dev.atlassian.net/browse/SUPPORT-1743


shubham1g5:
@ShivamPokhriyal  this seems like something that's visible to all users, so we should apply the appropriate product all users flag.

PR #912, opened 2020-06-11
Reword error message

shubham1g5:
This popped up on Field/dev and was confusing for people to know what exactly is going on. So being more explicit in the error.


ctsims:
@shubham1g5 when the user gets this error, does it come with enough context to pinpoint which question or selection has the empty value? If we are already reviewing the message might be worth it to go one step further and ensure it's highly actionable without searching form for empty selections

PR #914, opened 2020-06-12
Merge Commcare 2.49 to formplayer. 

ShivamPokhriyal:


ShivamPokhriyal:
I realised I didn't merge 2.49 into formplayer. FYI @shubham1g5

shubham1g5:
yeah, we are holding on to it as it's going to require a formplayer DB upgrade and might cause a lot of refreshes simultaneously on Formplayer. So we are just delaying it for now until there is a better time to do it.

PR #916, opened 2020-06-18
Enhancements to Query Prompts definition 

shubham1g5:
[Spec](https://docs.google.com/document/d/1RWrMhP180r2uJFAwrkHE3Fkm8oxuqWhPIqrcKhPGt8g/edit?ts=5ed0523f)

Adds support for - 
1. `appearance` and `input` attributes in `prompt` model
2. `itemset` as a child of `prompt`

### DIFF
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -1,5 +1,6 @@
 package org.commcare.session;
 
+import org.commcare.cases.util.StringUtils;
 import org.commcare.modern.util.Pair;
 import org.commcare.suite.model.QueryPrompt;
 import org.commcare.suite.model.RemoteQueryDatum;
@@ -102,7 +103,10 @@ public class RemoteQuerySessionManager {
         }
         for (Enumeration e = userAnswers.keys(); e.hasMoreElements(); ) {
             String key = (String)e.nextElement();
-            params.put(key, userAnswers.get(key));
+            String value =  userAnswers.get(key);
+            if(!StringUtils.isEmpty(value)) {
+                params.put(key, userAnswers.get(key));
+            }
         }
         return params;
     }
@@ -160,7 +164,7 @@ public class RemoteQuerySessionManager {
     // checks if @param{value} is one of the select choices give in @param{items}
     private boolean checkForValidSelectValue(Vector<SelectChoice> items, String value) {
         // blank is always a valid choice
-        if (value.contentEquals("")) {
+        if (StringUtils.isEmpty(value)) {
             return true;
         }
         for (int i = 0; i < items.size(); i++) {
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -4,6 +4,7 @@ import org.commcare.modern.util.Pair;
 import org.commcare.suite.model.QueryPrompt;
 import org.commcare.suite.model.RemoteQueryDatum;
 import org.commcare.suite.model.SessionDatum;
+import org.javarosa.core.model.SelectChoice;
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.ExternalDataInstance;
 import org.javarosa.core.model.instance.TreeElement;
@@ -23,6 +24,9 @@ import java.io.InputStream;
 import java.net.URL;
 import java.util.Enumeration;
 import java.util.Hashtable;
+import java.util.Vector;
+
+import static org.commcare.suite.model.QueryPrompt.INPUT_TYPE_SELECT1;
 
 /**
  * Manager for remote query datums; get/answer user prompts and build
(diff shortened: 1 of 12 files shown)


src/main/java/org/commcare/xml/QueryPromptParser.java
```diff
@@ -0,0 +1,73 @@
...
+            } else if (NAME_ITEMSET.equals(parser.getName().toLowerCase())) {
+                itemsetBinding = parseItemset();
+            }
+        }
+        return new QueryPrompt(key, appearance, input, display, itemsetBinding);
+    }
+
+    private ItemsetBinding parseItemset() throws IOException, XmlPullParserException, InvalidStructureException {
+        ItemsetBinding itemset = new ItemsetBinding();
+        itemset.contextRef = TreeReference.rootRef();
```
shubham1g5, line 59: Not sure if it's right to assign the `contextRef` to the root here , or possibly it should just be a self reference to the prompt key ?

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -114,4 +134,41 @@ public static String evalXpathExpression(XPathExpression expr,
         return new Pair<>(ExternalDataInstance.buildFromRemote(queryDatum.getDataId(), root, queryDatum.useCaseTemplate()), "");
     }
 
+    public void populateItemSetChoices(QueryPrompt queryPrompt) {
+        evaluationContext.setVariables(userAnswers);
```
ctsims, line 138: If you are going to change scoped values in an EvaluationContext you should always spawn a new context (`evaluationContext.spawnWithCleanLifecycle()`) before doing so, since you don't want to change the "root" context data for future requests.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -114,4 +134,41 @@ public static String evalXpathExpression(XPathExpression expr,
...
+        ItemsetBinding itemset = queryPrompt.getItemsetBinding();
+        if (itemset != null) {
```
ctsims, line 140: Am I right in assuming this is all pulled ~wholesale from the Form Engine?

Am wondering whether it's identical enough to try to put this into a helper method of some kind. Some of the optimizations around populating Select Choices are pretty powerful, but would be very brittle to replicate

src/main/java/org/commcare/xml/QueryPromptParser.java
ctsims: contextRef is probably fine. There really isn't a "default" document in play for the query page, so I think setting it to `null` would be semantically right, but if that causes problems, the root ref is always a safe thing to anchor on.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1473,9 +1474,9 @@ private EvaluationContext getPotentiallyLimitedScopeContext(EvaluationContext qu
         return newReferences;
     }
 
-    private SelectChoice buildSelectChoice(TreeReference choiceRef, ItemsetBinding itemset,
-                                           DataInstance formInstance, EvaluationContext ec,
-                                           ReducingTraceReporter reporter, int index) {
+    public static SelectChoice buildSelectChoice(TreeReference choiceRef, ItemsetBinding itemset,
```
ctsims, line 1477: I think we should definitely move this into a helper class to make it clearer how its context is built and protect potential scope issues (you could imagine an accidental static reference within this file being made to an element in the method), and make it clear that this isn't just an XForm operation

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -129,4 +133,41 @@ public void populateItemSetChoices(QueryPrompt queryPrompt) {
         evalContextWithAnswers.setVariables(userAnswers);
         ItemSetUtils.populateDynamicChoices(queryPrompt.getItemsetBinding(), evalContextWithAnswers);
     }
+
+    // loops over query prompts and validates selection until all selections are valid
+    public void refreshItemSetChoices(Hashtable<String, String> userAnswers) {
+        OrderedHashtable<String, QueryPrompt> userInputDisplays = getNeededUserInputDisplays();
+        boolean dirty = true;
+        while (dirty) {
```
ctsims, line 145: Very minor point: It makes sense why this code should converge, since any invalid choices are blanked out, and the worst case scenario for 5 choice boxes is to loop through 5 times, blanking one answer each time until all are blank.

That being said, while(X) loops make for challenging debugging if anything goes wrong with the plan, since you'll get a hang-and-kill rather than an explicit error.

From a style/future proofing perspective, I think it might make sense to add an explicit fallback in here where you only loop over the questions N times (where N is the size of the question set) and throw a runtime exception if the dirty flag is still set after the Nth pass, rather than continuing to loop. That ensures that the code functioning unexpectedly will give you an explicit exception which we can follow up on, rather than a hang, which we may never be notified of from the crash logs.

ctsims (approved):

PR #921, opened 2020-07-07
Use the new KXML lib across the board to fix Emoji Issues

ctsims:
Formplayer was also experiencing the problem identified in 
https://github.com/dimagi/commcare-core/pull/878

This updates formplayer to the same updated kxml dependency as Android uses


dannyroberts:
Awesome!

PR #922, opened 2020-07-08
255 chars limit on case name, type, owner_id and external_id 

shubham1g5:
Adds 255 chars limit on case_name, case_type, owner_id and external_id to be in sync with [HQ limitations](https://github.com/dimagi/commcare-hq/blob/f7b136fb44787e5fdacde7dce5c922880e9a01c8/corehq/form_processor/backends/sql/update_strategy.py#L54-L60) and fail fast on mobile itself. 

Product Note: Throws an error on saving a form with any of case_name, case_type, owner_id and external_id for a case is greater than 255 chars in length.

### DIFF
--- a/src/main/java/org/commcare/xml/CaseXmlParser.java
+++ b/src/main/java/org/commcare/xml/CaseXmlParser.java
@@ -217,19 +217,19 @@ public class CaseXmlParser extends TransactionParser<Case> {
     private void checkForMaxLength(Case caseForBlock) throws InvalidStructureException {
         if (caseForBlock.getTypeId().length() > 255) {
             throw new InvalidStructureException(
-                    "Invalid <case_type>, value must be 256 characters or less",
+                    "Invalid <case_type>, value must be 255 characters or less",
                     parser);
         } else if (caseForBlock.getUserId().length() > 255) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/xml/CaseXmlParser.java
```diff
@@ -209,6 +211,27 @@ private void updateCase(Case caseForBlock, String caseId) throws InvalidStructur
                     break;
             }
         }
+        checkForMaxLength(caseForBlock);
+    }
+
+    private void checkForMaxLength(Case caseForBlock) throws InvalidStructureException {
+        if (caseForBlock.getTypeId().length() > 255) {
+            throw new InvalidStructureException(
+                    "case_type is longer than 255 chars for case with ID: " + caseForBlock.getCaseId() + " and case_type " + caseForBlock.getTypeId(),
```
ctsims, line 220: I'd rephrase these based on the action they encode

IE: "Invalid <case_type>, value must be 256 characters or less"

shubham1g5:
@ctsims  yeah, any such forms gets quarantined on phone as soon as phone tries to submit them to HQ.

PR #924, opened 2020-07-24
Improve performance for queries against missing case properties

ctsims:
Previously, request patterns which referred to case properties which may or may have existed, like

```
instance('casedb')/casedb/case[3]/my_property = 'value'
```

had an awkward resolution path when the `my_property` case property is not set. Those requests would fail to take advantage of a [core optimization](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xpath/XPathLazyNodeset.java#L104) which allowed much, much faster resolution for 'normal' references which don't use xpath predicates. 

This change lets the normal "Go look up a node" process correctly identify the "shadow" elements which represent a case property that doesn't exist, and returns that virtual node. That prevents the system from creating, throwing, and catching an XPathTypeException potentially tens-of-thousands of times during request evaluation. Since that XPathTypeException requires gaining a lock on a static type, this also will potentially reduce concurrency issues when multiple such requests are running simultaneously.

This PR does change the resolution behavior of CommCare slightly. It will result in `count()` or other queries which depend on an element existing or not (rather than its value) "seeing" the virtual element more, since a real treeelement is created to represent it. This was true before as well, but in many fewer circumstances. That change still needs a bit more scrutiny before this can be included safely.

### DIFF
--- a/src/main/java/org/javarosa/core/model/instance/DataInstance.java
+++ b/src/main/java/org/javarosa/core/model/instance/DataInstance.java
@@ -118,7 +118,7 @@ public abstract class DataInstance<T extends AbstractTreeElement<T>> implements 
             }
             if (mult == TreeReference.INDEX_UNBOUND) {
                 int inferredMultiplicity = node.getChildMultiplicity(name);
-                if (inferredMultiplicity == 1 || inferredMultiplicity ==0) {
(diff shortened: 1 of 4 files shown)


dannyroberts:
Acknowledging I don't have a lot of context, based on your description of the issue above and the code changes, looks good to me

PR #925, opened 2020-07-26
Delete repeat group question when count changes

ShivamPokhriyal:
**Jira** https://dimagi-dev.atlassian.net/browse/SAAS-11078

**Summary**

When the repeat count changes(more specifically when it's updated to a lesser value) the repeat group doesn't update itself. 
i.e. The repeat will +n but not -n. So if I edit count from 1 to 5, I'll get 5 questions which is perfect. But If I reduce it to 1 from 5, I will still have 5 questions. 

Although the form submission will have correct questions. But if user types in a wrong value(and realises it later) it becomes really difficult to correct the form navigation. 

So now, we'll delete the repeat questions from the end if the count is updated to a lower value.

### DIFF
--- a/src/main/java/org/javarosa/form/api/FormEntryModel.java
+++ b/src/main/java/org/javarosa/form/api/FormEntryModel.java
@@ -332,7 +332,7 @@ public class FormEntryModel {
         }
     }
 
-    private static void createModelForGroup(GroupDef g, FormIndex index, FormDef form) {
+    private void createModelForGroup(GroupDef g, FormIndex index, FormDef form) {
         if (g.isRepeat() && g.getCountReference() != null) {
             TreeReference countRef = g.getConextualizedCountReference(index.getReference());
             IAnswerData count = form.getMainInstance().resolveReference(countRef).getValue();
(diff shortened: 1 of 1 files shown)


shubham1g5:
@ctsims Would you be able to have a look at this. I feel like there might be a reason we were not doing this already.

ctsims:
Hi @shubham1g5 

Thanks for flagging. It is definitely intentional behavior to not delete the existing repeats. Since the system doesn't know exactly which repeats the user wants to remove by lowering the count, removing repeats automatically has unpredictable outcomes and can remove data the user has input. This can get much more complex when the repeat counts are set through automated elements in the form. 

This has always been somewhat of a trade for one usability outcome over the other. The upside of the current approach is that it's safer and you can replicate the outcome this PR is completing with a fairly straightforward form design like

https://www.commcarehq.org/a/swat/apps/view/1f97178af7c91272739617ec062d6bc8/form/d71102c5773cab59428db26df4c3133dd286ab0c/source/#form/how_many_repeats

whereas if the delete behavior were the default there wouldn't be a way to replicate the alternative.

shubham1g5:
@ctsims  Thanks for the detailed explaination. The form design you linked does achieve the same behaviour as this PR is shooting for, so are you saying that we want to leave it to app builders to opt or not opt for this design ? Although I can't think of a scenario where the app builder won't like to opt for this form design. 

Or is it the case that we prefer to do this at CommCare app level instead of a doing it at platform level since it gives the flexibility to user to revert the lowering of count in case he noticed that some of the data he actually wanted went missing when the count got lowered.

ctsims:
Hi @shubham1g5. The reason for opting for the design I proposed is that it's nondestructive. Repeats can be added and removed while retaining the underlying data for the repeats which are removed. This wouldn't be possible if the engine itself was removing the elements. This can get quite challenging especially if anything is automated, like model iterations, since those will associate data with individual repeats 

I think we should continue with the non-destructive approach rather than making the change to the core behavior, especially given that if anyone is relying on the current behavior with model iterations or otherwise it could cause issues that users couldn't recover from with form changes.

shubham1g5:
@ctsims  Thanks that makes a ton of sense.

ShivamPokhriyal:
> The reason for opting for the design I proposed is that it's nondestructive. Repeats can be added and removed while retaining the underlying data for the repeats which are removed.

Does it mean that the data in the repeat group will retain itself if user decrease and then increase the count? 

I tried this app https://www.commcarehq.org/a/swat/apps/view/1f97178af7c91272739617ec062d6bc8/form/d71102c5773cab59428db26df4c3133dd286ab0c/source/#form/how_many_repeats and changed the label to a text question in the repeat group.

And then started with a repeat count of 3. Added some text in the 3 repeat questions. Then decreased the count to 0 and increased it again to 3. None of the values in the 3 repeat questions retained. They're all blank.

PR #926, opened 2020-07-29
Logging for null root reference in an instance

shubham1g5:
Trying to get more info on [crash](https://console.firebase.google.com/u/0/project/commcare-a57e4/crashlytics/app/android:org.commcare.lts/issues/5f0542607dbbf9b0149d517821b8c4cb?time=last-thirty-days&sessionId=5F2050F3001D00011E7C6994E3529691_DNE_0_v2)

### DIFF
--- a/src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
+++ b/src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
@@ -318,13 +318,13 @@ public class EvaluationContext {
     private void expandReferenceAccumulator(TreeReference sourceRef, DataInstance sourceInstance,
                                             TreeReference workingRef, Vector<TreeReference> refs,
                                             boolean includeTemplates) {
-        int depth;
-        try {
-            depth = workingRef.size();
-        } catch (NullPointerException e) {
-            throw new RuntimeException("Invalid instance definition encountered while evaluating " + sourceRef.toString() +
-                    " for instance " + sourceInstance.getInstanceId() + " with root: " + sourceInstance.getRoot(), e);
+
+        if (workingRef == null) {
+            throw new RuntimeException("Encountered invalid instance definition while evaluating " + sourceRef.toString() +
+                    " for instance " + sourceInstance.getInstanceId() + " with root: " + sourceInstance.getRoot());
         }
+
+        int depth = workingRef.size();
 
         if (depth == sourceRef.size()) {
             // We've matched fully
--- a/src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
+++ b/src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
@@ -299,11 +299,6 @@ public class EvaluationContext {
         DataInstance baseInstance = retrieveInstance(ref);
         Vector<TreeReference> v = new Vector<>();
 
-        if (baseInstance.getRoot().getRef() == null) {
-            Logger.log(LogTypes.SOFT_ASSERT, "Invalid instance definition encountered while evaluating " + ref.toString() +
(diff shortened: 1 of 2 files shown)


ctsims:
Seems like a reasonable debugging change. 

Two things

1) I saw that evaluation started with a org.commcare.cases.util.StorageBackedTreeRoot, but then jumps into  unpacking a separate reference, which means that it could be expanding a non-StorageBacked tree element.  (the toString() would still help with context either way, though).
2) This is writing the underlying log to Sumo, right?  It seems like it might be a good idea to catch and rethrow what we know will eventually be an error, which will make the error context visible in the crash itself without cross referencing. Either way we are wrapping custom error handling, but one changes the semantic meaning of the error, which seems helpful.

also: did you notice that there's a stream of install issues in the user analytics logs? Any chance those might shine some light on why we are seeing the issue?

shubham1g5:
@ctsims Thanks for the prompts. 

> This is writing the underlying log to Sumo, right?
Yes and crashlytics, the `logs` tab in the crashlytics includes both the analytics events and these logs that happened in the same session as the crash. I can still see that how it's a good idea to throw this logging as the crash itself and have made the respective change. 

> did you notice that there's a stream of install issues in the user analytics logs? Any chance those might shine some light on why we are seeing the issue?
I didn't see those, though It's hard to correlate those with these crashes without further context, there are also instances of this crash without the install issues so it might not be necessarily related.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -323,8 +318,13 @@ public void setQueryContext(QueryContext queryContext) {
     private void expandReferenceAccumulator(TreeReference sourceRef, DataInstance sourceInstance,
                                             TreeReference workingRef, Vector<TreeReference> refs,
                                             boolean includeTemplates) {
-
-        int depth = workingRef.size();
+        int depth;
+        try {
+            depth = workingRef.size();
```
ctsims, line 323: From a style perspective, I'd generally do the most specific check possible (which is a direct check-and-throw), rather than a catch-and-throw, to prevent an issue in the future where more than one dereference occurs in the this block and it obfuscates the error, that also in theory prevents confusion if `workingRef.size()` throws an NPE internally rather than due to the dereference.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -323,8 +318,13 @@ public void setQueryContext(QueryContext queryContext) {
...
+        } catch (NullPointerException e) {
+            throw new RuntimeException("Invalid instance definition encountered while evaluating " + sourceRef.toString() +
```
ctsims, line 325: Another minor style quibble: when we do catch-and-throw I think it's good to get into the habit of including the original throwable as the context (second arg to the RuntimeException). That would, for instance, let someone reviewing the stack trace identify the weird side-case where the NPE was thrown from inside `size()`

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
shubham1g5: Agree and made the change.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
shubham1g5: yeah totally agree. [The commit](https://github.com/dimagi/commcare-core/pull/926/commits/5d6e4d3d1e78043e272ab56f35ac08784d14ec63) did make the exact correction earlier though it's no longer relevant.

PR #928, opened 2020-08-20
Fix Fractional Time Comparisons with DST

ctsims:
When using the fractional date (`double(date())`) evaluation [introduced here](https://github.com/dimagi/commcare-core/commit/e0d29f5e04ce62110de8788ea5054ba1637e5b64), the system relies on the 'default' timezone being the same between the date presented for comparison and the `EPOCH_DATE`.

Currently, this results in an issue when the default TimeZone is DST aware for its locale, and the current DST offset for the current instance is different from the DST offset in the TimeZone on the date of the epoch. For instance, the offset for the "Eastern Time" timezone August 20, 2020 is UTC-04:00, but the offset for the "Eastern Time" timezone on the date of the epoch on January 1, 1970 was UTC-05:00. 

It's not obvious to me whether this has ~always been an issue which simply hasn't come up, or whether changes to the thoroughness of TimeZones on Android has introduced the disparity. 

This change explicitly factors in the offsets being introduced by the associated `Date` objects, regardless of why those offsets were introduced, so it should be a safe no-op in any situation  where the default TimeZone has consistent offsets for both the current instant and for the epoch instant.

### DIFF
--- a/src/main/java/org/javarosa/core/model/utils/DateUtils.java
+++ b/src/main/java/org/javarosa/core/model/utils/DateUtils.java
@@ -32,8 +32,10 @@ public class DateUtils {
     private static TimezoneProvider tzProvider = new TimezoneProvider();
 
     public static final long HOUR_IN_MS = TimeUnit.HOURS.toMillis(1);
-    public static final long DAY_IN_MS = TimeUnit.DAYS.toMillis(1);;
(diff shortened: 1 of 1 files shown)


ShivamPokhriyal (approved):
:

PR #929, opened 2020-08-23
Wires in `InstallRequestSource` in install paths

shubham1g5:
cross-request: https://github.com/dimagi/commcare-android/pull/2322

### DIFF
--- a/src/main/java/org/commcare/resources/model/ResourceTable.java
+++ b/src/main/java/org/commcare/resources/model/ResourceTable.java
@@ -18,6 +18,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Enumeration;
 import java.util.Hashtable;
+import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Stack;
 import java.util.Vector;
@@ -376,7 +377,7 @@ public class ResourceTable {
                                                 boolean upgrade,
                                                 CommCarePlatform platform,
                                                 ResourceTable master,
-                                                boolean recovery)
+                                                boolean recovery, Map<String, String> customHeaders)
             throws UnresolvedResourceException, UnfullfilledRequirementsException, InstallCancelledException {
 
         // TODO: Possibly check if resource status is local and proceeding to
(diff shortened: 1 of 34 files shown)


src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -1169,14 +1163,19 @@ public boolean recoverResources(CommCarePlatform platform, String profileRef, Ve
...
+    public void recoverResource(Resource missingResource, CommCarePlatform platform, String profileRef, InstallRequestSource source)
             throws InstallCancelledException, UnresolvedResourceException, UnfullfilledRequirementsException {
 
         if (missingResource.id.contentEquals(CommCarePlatform.APP_PROFILE_RESOURCE_ID)) {
             addRemoteLocationIfMissing(missingResource, profileRef);
         }
 
-        findResourceLocationAndInstall(missingResource, new Vector<>(), false, platform, null, true);
+        try {
+            platform.registerInstallContext(new ResourceInstallContext(source));
```
ctsims, line 1174: I have mixed feelings about this approach. I think I'd default to passing through at the method level rather than making the Platform object stateful in this way, since it makes the platform object much, much more complex (and it makes it hard for someone to identify that the install context is available / required and hidden inside the platform unless you happen to know) but non-blocking as feedback.

ctsims (approved):

src/main/java/org/commcare/resources/model/ResourceTable.java
shubham1g5: Agree that this design does feel bad. As an alternative do you mean have `ResourceInstallContext` as a method parameter in `ResourceInstaller::install` right ?

ctsims: yeah, I'm not sure if there are limitations there which I'm missing, but I'm imagining it replacing the existing `upgrade` boolean flag, providing a single place where new context fields (like that upgrade boolean) can be expanded without us needing to change all of these signatures again in the future.

ShivamPokhriyal (approved):

src/main/java/org/commcare/resources/model/ResourceTable.java
shubham1g5: @ctsims  merging this after implementing this suggestion, though feel free to have a final review anyway and leave feedback if any.

PR #930, opened 2020-08-31
Adds a method to get all ids for all resources in the table

shubham1g5:
cross-request: https://github.com/dimagi/commcare-android/pull/2335

### DIFF
--- a/src/main/java/org/commcare/modern/database/DatabaseHelper.java
+++ b/src/main/java/org/commcare/modern/database/DatabaseHelper.java
@@ -34,6 +34,11 @@ public class DatabaseHelper {
     public static Pair<String, String[]> createWhere(String[] fieldNames, Object[] values,
                                                      EncryptedModel em, Persistable p)
             throws IllegalArgumentException {
+        // null selection
+        if (fieldNames.length == 0) {
+            return new Pair<>(null, null);
+        }
+
         Set<String> fields = null;
         if (p instanceof IMetaData) {
             IMetaData m = (IMetaData)p;
--- a/src/main/java/org/commcare/resources/model/ResourceTable.java
+++ b/src/main/java/org/commcare/resources/model/ResourceTable.java
@@ -1210,7 +1210,7 @@ public class ResourceTable {
         return storage.getIDsForValue(Resource.META_INDEX_LAZY, LAZY_VAL_TRUE);
     }
 
-    public List<Integer> getResourceIds() {
-        return storage.getAllIDs();
+    public List<Integer> getAllResourceIds() {
+        return storage.getIDsForValues(new String[]{}, new String[]{});
     }
 }
--- a/src/main/java/org/commcare/resources/model/ResourceTable.java
+++ b/src/main/java/org/commcare/resources/model/ResourceTable.java
@@ -18,6 +18,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Enumeration;
 import java.util.Hashtable;
+import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Stack;
 import java.util.Vector;
@@ -1207,5 +1208,9 @@ public class ResourceTable {
 
     public Vector<Integer> getLazyResourceIds() {
         return storage.getIDsForValue(Resource.META_INDEX_LAZY, LAZY_VAL_TRUE);
+    }
+
+    public List<Integer> getResourceIds() {
+        return storage.getIDsForValues(new String[]{}, new String[]{});
     }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/org/commcare/resources/model/ResourceTable.java
```diff
@@ -1208,4 +1209,8 @@ public void setMissingResources(SizeBoundUniqueVector<Resource> missingResources
     public Vector<Integer> getLazyResourceIds() {
         return storage.getIDsForValue(Resource.META_INDEX_LAZY, LAZY_VAL_TRUE);
     }
+
+    public List<Integer> getResourceIds() {
```
ShivamPokhriyal, line 1213: nit: How about `getAllReourceIds`?

ShivamPokhriyal (dismissed):

src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java
```diff
@@ -150,6 +150,12 @@
      */
     Object getAccessLock();
 
+
+    /**
+     * @return a Vector of IDs of all Externalizable objects in storage
+     */
+    Vector<Integer> getAllIDs();
```
ShivamPokhriyal, line 157: Wondering whether it needs to be implemented on formplayer?

ctsims (changes requested):
I don't think that this change should warrant changing the IStorageUtilityIndexed interface / contract, which we should consider a breaking change for implementing platforms.

The getAllIDs() method here could be replicated by either doing an iterate with the 'false' flag set (to ensure that records don't get sent down) and pulling out the integers (which is what the implementation ends up doing) or by calling getIDsForValues() with empty String inputs as filters (which is how this would work in a sql-ish pattern). My guess is that the latter is the safest / best implementation for our needs, and that if it the current code doesn't run that way it's safe to change it to function that way, based on the method contract.

shubham1g5:
@ctsims Updated based on the suggestion.

ctsims (approved):

src/main/java/org/commcare/modern/database/DatabaseHelper.java
```diff
@@ -34,6 +34,11 @@
     public static Pair<String, String[]> createWhere(String[] fieldNames, Object[] values,
```
ctsims, line 34: Can we add a test to confirm that the underlying endpoints that use this return all results and not an error, so the behavior can be regressed?

shubham1g5: Good point, does [this](https://github.com/dimagi/commcare-core/pull/930/commits/7cb5743d401602d9382cc2e8b8b242878b8b4985) look good ?

src/test/java/org/javarosa/core/storage/IndexedStorageUtilityTests.java
```diff
@@ -131,22 +128,32 @@ public void testBulkMetaMatching() {
...
     }
 
+    @Test
+    public void testReadingAllEntries() {
+        writeBulkSets();
+        List<Integer> matches = storage.getIDsForValues(new String[]{}, new String[]{});
+        Set<Integer> expectedMatches = getIdsFromModels(tenSizesOfMensNikes);
+        expectedMatches.addAll(getIdsFromModels(eightSizesOfWomensNikes));
+        expectedMatches.addAll(getIdsFromModels(fiveSizesOfMensVans));
+        Assert.assertEquals("Failed index match [brand,style][nike,mens]", expectedMatches, new HashSet<>(matches));
```
ctsims, line 138: I think this assert has the wrong text.

src/test/java/org/javarosa/core/storage/IndexedStorageUtilityTests.java
```diff
@@ -131,22 +128,32 @@ public void testBulkMetaMatching() {
         Assert.assertEquals("Failed index match [brand,style][nike,mens]", getIdsFromModels(tenSizesOfMensNikes), getIdsFromModels(matchedRecords.toArray(new Shoe[]{})));
     }
 
+    @Test
+    public void testReadingAllEntries() {
+        writeBulkSets();
+        List<Integer> matches = storage.getIDsForValues(new String[]{}, new String[]{});
```
ctsims, line 134: Style nit: You can also instantiate this with `new String[0]` which might make the intent slightly clearer

ctsims (dismissed):
Left a couple tiny point of feedback, but the new test looks good (And thanks for the linting)

ctsims (approved):

PR #931, opened 2020-09-23
Kev/add timezone

knguyen142:
Related to https://dimagi-dev.atlassian.net/browse/SAAS-11203 and adds the option to provide timezones into the date string format, but I realized that it does not fix our problem. I first thought that providing the timezone would make our code accept different timezones other than UTC, but then saw that https://github.com/dimagi/commcare-hq/blob/579b52f1037d30c8e9184214f7278bb1d6a8b9d2/corehq/ex-submodules/dimagi/utils/parsing.py#L47 pretty much forces us into `Z`.

Still, this allows the option for users to specify timezone. 

## Testing 

Added test cases, does not break current tests, and tested locally.

Screen Shot 2020-09-23 at 5 18 21 PM

Verified UTC shift in commcare-hq
Screen Shot 2020-09-24 at 11 46 59 AM


knguyen142:
@ctsims what are the steps to get this onto staging so we can test and see how adding timezones interacts with other pieces of formplayer/cchq?

ctsims:
@knguyen142 

Two things

1) In terms of staging, I believe the mechanism for deploying new formplayer code is to pull the commcare-core head in the `autostaging` branch of the formplayer repo up to your commit, wait for Jenkins to complete the formplayer-staging branch and then do a staging formplayer deploy.

2) If we are going to start supporting this I think we need to be really careful about what the timezone offset actually ends up meaning. There are three edge cases specifically that are important to verify.

a) Making sure that the offset changes track across DST in DST sensitive locales (IE: Depending on the Day of the year, a DateTime on a phone in Boston should have either a -4 or -5 ISO offest.

b) Making sure that non-integer timezone representations are formatted properly: India's timezone is UTC+05:30, we should make sure that the %Z formatting follows the proper convention for that

c) Making sure that locale offsets properly follow historical changes: Nepal's UTC Offset was +05:30 until 1986, when it became +05:45. 

All of those might have tests elsewhere in the code, but because this method is introducing the datetime parsing and timezone management in a new context, I think it would be good to make sure as many of those elements are tested as possible to make sure that the offset information is introduced in a Date sensitive context.

knguyen142:
Thanks @ctsims. I've added some tests https://github.com/dimagi/commcare-core/pull/931/commits/41aac35fe257ad1e781232807039b20fe94bf536 for completeness, but I might not be understanding how some of your use cases would apply to this context. 

The `format-date` function works by using the timezone offset (in ms) that's passed up from the client to use in the [timezone provider](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/utils/TimezoneProvider.java#L8). So they are already giving us the numerical offset, and that is the source of truth as opposed to string interpretations of the timezone. At least that is my understanding, there might be some other use case I'm not familiar with?

ctsims:
@knguyen142 Ah, interesting. That's definitely what I was worried about, since that means the engine on formplayer isn't actually able to provide the user with an accurate timestamp. If yesterday was DST and today is non-DST, if the user somehow wants to display a format a datetime from yesterday "at 10am", they're actually going to be formatting a datetime from yesterday at 11am, since the system is applying the offset blind based on today's locale info, not the offset at the date being requested. Phones can likely do a bit better, since they have the full locale information available when we do these calculations, not just an offset.

You said that this is no longer coupled with SaaS's actual bug, and the fix for that is getting handled on the server side now, right? I don't want to block this if there's a meaningful fix that it's preventing, but I'm generally hesitant for us to push forward changes to the engine which expose incorrect data. If our system is weak on handling timezones, I'd rather that weakness be exposed through limited information rather than wrong information IE: Not attaching a timezone offset because we can't be confident it is correct rather than attaching an offset which is correct most of the time but wrong in others.

knguyen142:
Thanks for bringing up that case @ctsims, we wanted to see which use cases would be affected by this change. I'm not familiar with apps well enough to know how the yesterday on a DST change would go about. What would be the XPATH expression for that? I looked at the [functions](https://confluence.dimagi.com/display/commcarepublic/CommCare+Functions#CommCareFunctions-today) and didn't know how.

I'm looking at the ticket again, and I'm not sure if this change is decoupled from the clients request.

ctsims:
@knguyen142 The DST situation would just consist of a user trying to use the `format-date` function with a datetime that isn't the same as the current instance. The way that your method is referring to the method uses an explicit offset, but there are multiple paths to the same date math, and in some of them the offset is pulled from the JVM locale, which means that it will change depending on the date of the instant.

For example: If it is current October 31st at 9pm and I am in Boston, the `format-date` function for the current instant will give me

`2020-10-31T21:00:00-04`

The time 10 hours after the current instant in "Eastern Time" would be
`2020-11-01T10:00:00-05`

But it would also be acceptable for that time to be 
`2020-11-01T11:00:00-04`

What is important to confirm here is that the dates and offsets that are used won't produce

`2020-11-01T10:00:00-04`

For an instant in the future. This can happen if the datetime isn't being normalized and one representation is tracking "Wall Clock" time with the offset being locked in independent of the date. 

Unfortunately I don't think there's an obvious way in the engine to produce this case, but it should be possible to calculate by manually manipulating a Date that you pass to the supporting function implementation.

Sorry this is such a pain, the engine was originally optimized to "smooth over" dates and timezones to ensure that the system produced a consistent wall-calendar date. As we build "over" that functionality, I want to make sure we aren't producing code that looks like it's more capable if we actually aren't.

knguyen142:
@ctsims I don't have too much experience with datetime strings in products, but I'm ramping up and want to see if providing a standard way of providing the locale/timezone would fix our issue:

1. Allowing users to use the project settings timezone:
Screen Shot 2020-10-08 at 9 46 49 AM

The interface could be accomplished by creating a new function e.g. `format-date-with-project-timezone(now(), "%Y-%m-%d %H:%M:%S%Z")` or passing in a parameter `format-date(now(), "%Y-%m-%d %H:%M:%S%Z", "PROJECT_TIMEZONE")`. Note, I am not sure if the request has access to the project timezone during that function. 

2. Allowing users to provide a parameter specifying the timezone/locale (from a fixed, standard format) in the the function, e.g., `format-date(now(), "%Y-%m-%d %H:%M:%S%Z", "US/Eastern")`. 

In both cases, the parameter would override the timezone offset passed by the user, and I believe should work on mobile and web. It standardizes things so users from another timezone would not have inconsistencies. It is an optional feature so should be backwards compatible. And timezone format specifier %Z will fail if the timezone parameter is not provided.  

Question:
Should we still output the timezone offset (e.g. `-04`) or the long form (e.g. `EDT` or `US/Eastern`). or provide 2 format specifiers? This is where I'm not sure if it will have other ramifications.

ctsims:
Hi Kevin,

The device certainly won't have access to the timezone offset from the server in most cases. I think the server timezone should only influence transforms, and/or how the server interprets "Ambiguous" dates with no timezone information.

I think we might be getting ahead of things a bit, so let me try to reframe where we are:
 * I think including the timezone offset for iso datetimes is a good goal
 * Timezone offsets are a bit tricky because they have dependencies on things like the date on which the offset is being requested in the current 'locale' of the evaluation
 * Mobile devices WILL have access to the 'locale' being evaluated. When we do date operations [like this one](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/utils/DateUtils.java#L197), we use the default Calendar instance on device (which will be in the timezone of the device locale), so I believe that if you simply request the offset field from the Calendar it will work correctly
* On *formplayer* this causes issues because formplayer provides a [raw timezone offset](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/utils/DateUtils.java#L202), that is provided by [the browser](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/BrowserValuesProvider.java#L14)

I think all we need to do is provide a second way to provide timezone information to the DateUtils function with the [TimezoneProvider](https://github.com/dimagi/commcare-core/blob/c4cb0d9c3d140cb3b811779f3ff4e25fdade4178/src/main/java/org/javarosa/core/model/utils/TimezoneProvider.java) which provides the browser's timezone by *locale* rather than by offset, so that we can use the [existing named timezone setter](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/utils/DateUtils.java#L136) in date functions

PR #932, opened 2020-09-30
Fixes issue where time values gets stripped from DateTime data in Xpath calculations

shubham1g5:
String casts dateTime data while unpacking a Xpath reference to avoid an issue where time values gets stripped from dateTime values because of `Date` casted values [getting converted to DateData in recalculate expressions by default](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/condition/Recalculate.java#L117) in case no explicit dataType is specificed for the Xpath node. Including a test case for sake of clarity.

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathPathExpr.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathPathExpr.java
@@ -258,8 +258,6 @@ public class XPathPathExpr extends XPathExpression {
             return (new XFormAnswerDataSerializer()).serializeAnswerData(val);
         } else if (val instanceof DateData) {
             return val.getValue();
-        } else if (val instanceof DateTimeData) {
-            return val.getValue();
         } else if (val instanceof BooleanData) {
             return val.getValue();
         } else if (val instanceof GeoPointData) {
--- a/src/test/java/org/javarosa/core/model/test/FormDefTest.java
+++ b/src/test/java/org/javarosa/core/model/test/FormDefTest.java
@@ -7,6 +7,7 @@ import org.javarosa.core.model.QuestionDef;
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.condition.IFunctionHandler;
 import org.javarosa.core.model.data.DateData;
+import org.javarosa.core.model.data.DateTimeData;
 import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.data.IntegerData;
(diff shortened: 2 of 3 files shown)


shubham1g5:
@ctsims Revisiting this and I unfortunately don't recollect what we decided as the outcome here. This is a data error and if we think fixing this might break more things, I would like to note this issue somewhere explicitly in the product docs as a limitation.

ctsims:
@shubham1g5 Yeah, I think we decided that this introduced ambiguous behavior we don't have the ability to resolve right now. 

I'm inclined to say that this github wiki is my preferred place to describe technical limitations like this, maybe we should make a page about Date and Time handling / limitations in general? I think there's a lot to write down on that topic, and putting it all in one place would be helpful

shubham1g5:
Thanks. Do you recall explicitly what was the ambiguous behaviour associated with this change ? Is is that people might be relying on the incorrect behaviour today ?

Agree that it would be good to add a page on date time limitations. I can try putting it together though would you be able to give me some pointers our outline of other limitations you are talking about.

ctsims:
Hi Shubham,

Sorry for the delay. I thought I remembered this a bit more clearly than I seem to in reviewing the code. 

I think the underlying issue you are describing is that DateTime's don't propagated in full, their time components end up stripped if you try to refer to them through untyped data. I think that is currently an intentionally enforced limitation, where the engine 'fuzzes' datetimes in most contexts to ensure that a calendar date stays a calendar date regardless of timezones, but I'm not 100% sure I can trace the logic for why this change breaks that without actually stepping through in more detail.

shubham1g5:
@ctsims I see, I am happy to close this out if this issue is not a big concern. Just want to flag though this issue exists in behaviour today when a CommCare App tries to save the value of a dateTime question as a case property.

ctsims:
One idea: maybe it would be good to at least encode the current behavior very specifically in the regression / unit tests as the place in-code that we document what happens? That would make it clear what the current behavior actually is, and if anyone makes an update that changes that behavior, we can add notes or a link to the explanation inline

shubham1g5:
> maybe it would be good to at least encode the current behavior very specifically in the regression / unit tests as the place in-code that we document what happens? 

You mean in the mobile code ? This PR [has a test](https://github.com/dimagi/commcare-core/blob/efe4b79c3051f901f4a5df0dbcf449366d63a844/src/test/java/org/javarosa/core/model/test/FormDefTest.java#L495) that checks for this behaviour. So would a similar test on **current** datetime behaviour is what you are referring to ?

PR #937, opened 2020-10-19
add option for minimal entity screen

calellowitz:
The steps this cuts represent about 2/3 of the time to initialize an EntityScreen, and are not needed for case lists in formplayer, although they do appear to be needed for case detail requests. There will be a linked formplayer PR to take advantage of this, if the approach seems reasonable.

The part I was most unsure about was autoselect, so I only made the optimization take effect if autoselect could not be invoked anyway (more than 1 case), since in that case the cost of the skipped operations would be minimal anyway.

I am reasonably certain this is safe, and tests pass (both automated and some manual), but I am not totally sure of all possible uses of the `EntityListSubscreen`, although it appears to only be relevant for case details, which is handled in a separate request in formplayer.

Easiest to review with `?w=1`

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityScreen.java
@@ -50,6 +50,7 @@ public class EntityScreen extends CompoundScreenHost {
     private Hashtable<String, TreeReference> referenceMap;
 
     private boolean handleCaseIndex;
+    private boolean full = true;
 
     private Vector<TreeReference> references;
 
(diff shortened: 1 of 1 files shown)


ctsims (changes requested):
This all makes sense, my only real feedback is that it's gross for us to decouple these so strongly in practice, but not in contract. I think making a brighter separation there code-wise doesn't feel practical (IE: making formplayer not rely on Screens at all but replicating a different, similar class)

Two things I would push on, though:

First: this method is part of the interface contract so it's especially abnormal to change the functioning of the class by overloading it's signature. Since `init` is an interface method, you don't have control everywhere that it's called.

I'd recommend instead putting the `full` flag into the *constructor,* which is a valid place to change the function of the class as a whole while retaining the validity of the `init` events. That might save you a headache as well if your EntryScreens are actually secretly init'd somewhere polymorphically.

The second is that I would like the inline documentation (probably method level javadoc) to more clearly specify what happens when you change this flag, not every detail, but at least at the high level. Ideally it would be great if the app also did something like throw a named error if you try to use the class members that don't get initialized anymore when 'full' is off, but taht's probably overkill from a practical perspective.

calellowitz:
@ctsims 
Updated based on your feedback

ctsims (approved):

PR #939, opened 2020-10-23
Ability to require search for a case list

orangejenny:
See https://github.com/dimagi/commcare-hq/pull/28677

This adds `require-search` as a boolean option on `<detail>` elements in the suite.
I considered adding it to the `<datum>` instead, as `detail-list-require-search`. Open to feedback.

### DIFF
--- a/src/main/java/org/commcare/suite/model/Detail.java
+++ b/src/main/java/org/commcare/suite/model/Detail.java
@@ -72,11 +72,13 @@ public class Detail implements Externalizable {
     private OrderedHashtable<String, String> variables;
     private OrderedHashtable<String, XPathExpression> variablesCompiled;
 
-
     private Vector<Action> actions;
 
     // Force the activity that is showing this detail to show itself in landscape view only
     private boolean forceLandscapeView;
+
+    // Only fetch entities if user has provided a search term
+    private boolean requireSearch;
 
     private XPathExpression focusFunction;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/suite/model/Detail.java
```diff
@@ -509,6 +515,10 @@ private boolean templatePathValid(String templatePathProvided) {
         return false;
     }
 
+    public boolean doesRequireSearch() {
```
orangejenny, line 518: I don't love this name, but `isRequireSearch` sounds awkward and `requiresSearch` is confusing.

ctsims:
@orangejenny Can you add some basic tests for this?

I think very simply model tests like these:
https://github.com/dimagi/commcare-core/blob/c4cb0d9c3d140cb3b811779f3ff4e25fdade4178/src/test/java/org/commcare/backend/suite/model/test/AppStructureTests.java#L31

are file, testing that a module without the flag doesn't have a new default and that the flag sets it

src/main/java/org/commcare/suite/model/Detail.java
snopoke: what about `isSearchRequired()`? (swapping the verb and the adjective to match other similar functions)

orangejenny: sounds good, 8964eb526

orangejenny:
@ctsims Thanks for pointing me to those tests. Added new ones in 5bcb9e4d6ac0fb97eb64af8adad2fe09bc07d7d4

orangejenny:
Closing in favor of https://github.com/dimagi/commcare-core/pull/949

PR #940, opened 2020-10-27
Corrects Cycle Detection Algo

shubham1g5:
Jira: https://dimagi-dev.atlassian.net/browse/SAAS-11463

While doing `depthFirstSearch` for detecting a cycle of XPathReferences, the current logic was using an [invalid construct](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/util/ShortestCycleAlgorithm.java#L96) `superList.contains(subList)` which always returns false for our use case since this method checks for the whole `subList` to be contained as a single element in `superList`. This was causing some of the cycles to go undetected. This PR correct this logic by implementing the right construct for this use case as `CollectionUtils::containsAny(superList, subList)`. I have not deep dived into when this issue will surface and when not but have added the real form as regression test to ensure future compatibility. 

Product Note: Fixes error handling for instances when cyclic references are present in a XForm

### DIFF
--- a/src/main/java/org/commcare/util/CollectionUtils.java
+++ b/src/main/java/org/commcare/util/CollectionUtils.java
@@ -1,5 +1,6 @@
 package org.commcare.util;
 
+import java.util.ArrayList;
 import java.util.Vector;
 
 /**
@@ -27,5 +28,20 @@ public class CollectionUtils {
             result[i] = second.elementAt(i - first.size());
         }
         return result;
+    }
+
+    /**
+     * Checks if any element in {@code subList} is contained in {@code superList}
+     * @param superList
+     * @param subList
+     * @return true if {@code superList} contains any element of {@code subList}, false otherwise
+     */
+    public static boolean containsAny(ArrayList<String> superList, ArrayList<String> subList) {
+        for (String item : subList) {
+            if(superList.contains(item)){
+                return true;
+            }
+        }
+        return false;
     }
 }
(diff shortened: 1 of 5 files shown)


ctsims (approved):
Impressive catch! Great work, Shubham.

src/main/java/org/commcare/util/CollectionUtils.java
```diff
@@ -28,4 +29,19 @@
...
+
+    /**
+     * Checks if any element in {@code subList} is contained in {@code superList}
+     * @param superList
+     * @param subList
+     * @return true if {@code superList} contains any element of {@code subList}, false otherwise
+     */
+    public static boolean containsAny(ArrayList<String> superList, ArrayList<String> subList) {
+        for (String item : subList) {
+            if(superList.contains(item)){
```
millerdev, line 41: What are the expected upper bounds for the sizes of these lists? This is O(M*N) time complexity, right? Could that cause performance issues since `containsAny()` is called in a loop in `ShortestCycleAlgorithm.java`?

shubham1g5: @millerdev That's a good point. I think the size of these lists should not go beyond O(100) for most practical purposes. Also this code only gets called to single out the nodes constituting the cycle when we have already detected that there is a cycle that exists in the form. So this will end up getting called only in scenarios where forms have errors(cycles) making me think we should not worry about performance here too much.

PR #943, opened 2020-10-28
add option to removeArchiveFile

knguyen142:
Related to https://dimagi-dev.atlassian.net/browse/SAAS-11333

We run into an issue where if the ccz that downloaded is corrupted, we hit an exception and formplayer does not try to download a new ccz. This allows formplayer to remove the zip file from memory so subsequent requests can try to download it.


knguyen142:
Never mind, I decided to close this because we cannot reproduce the underlying issue, so these changes might end up doing more harm than good

PR #944, opened 2020-10-30
Add encrypt-string as an XPath function

solleks:
Here is a version of encrypt-string as an XPath function. It replaces my first attempt to put encryption in the formplayer using Beans.

Please see [this document](https://docs.google.com/document/d/1OljNDYf-Z9nARKn5ww-ZZf2tWUIknHoEVJ__9Oq7YIo/edit) for some background on this change and choices made in the implementation.

Essentially, this encryption function accepts a message string and an encoded secret key, performs AES encryption using the GCM mode and returns the encoded message along with an initialization vector needed to decode it. It is intended to allow other algorithms to be used in the future, but only AES is implemented for now and certain encryption options are hardcoded that the message receiver will need to accomodate.

The code is passing its unit tests, but I have not yet tested it with the form that @ctsims designed. That won't happen until tomorrow. @calellowitz @orangejenny, I'll be asking you for review, but I understand if you prefer to wait until I've tested this with a form.

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
@@ -10,7 +10,6 @@ import org.javarosa.xpath.parser.XPathSyntaxException;
 
 import javax.crypto.Cipher;
 import javax.crypto.SecretKey;
-import javax.crypto.spec.GCMParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
@@ -47,7 +46,6 @@ public class XPathEncryptStringFunc extends XPathFuncExpr {
         String algorithm = FunctionUtils.toString(o3);
 
         final String ENCRYPT_ALGO = "AES/GCM/NoPadding";
-        final int TAG_LENGTH_BIT = 128;
         final int IV_LENGTH_BYTE = 12;
         final int KEY_LENGTH_BIT = 256;
 
@@ -70,13 +68,11 @@ public class XPathEncryptStringFunc extends XPathFuncExpr {
         }
         SecretKey secret = new SecretKeySpec(keyBytes, "AES");
 
-        byte[] iv = new byte[IV_LENGTH_BYTE];
-        new SecureRandom().nextBytes(iv);
-
         try {
             Cipher cipher = Cipher.getInstance(ENCRYPT_ALGO);
-            cipher.init(Cipher.ENCRYPT_MODE, secret, new GCMParameterSpec(TAG_LENGTH_BIT, iv));
+            cipher.init(Cipher.ENCRYPT_MODE, secret);
             byte[] encryptedMessage = cipher.doFinal(message.getBytes(Charset.forName("UTF-8")));
+            byte[] iv = cipher.getIV();
             byte[] ivPlusMessage = ByteBuffer.allocate(iv.length + encryptedMessage.length)
                 .put(iv)
                 .put(encryptedMessage)
--- a/src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
@@ -16,7 +16,7 @@ import java.nio.charset.Charset;
 import java.security.SecureRandom;
 
 public class XPathEncryptStringFunc extends XPathFuncExpr {
-    public static final String NAME = "encrypt-string";
+    public static final String NAME = "encrypt";
     private static final int EXPECTED_ARG_COUNT = 3;
 
     public XPathEncryptStringFunc() {
--- a/src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
@@ -1,0 +1,84 @@
+package org.javarosa.xpath.expr;
+
+
+import org.javarosa.core.model.condition.EvaluationContext;
+import org.javarosa.core.model.instance.DataInstance;
+import org.javarosa.xpath.XPathException;
+import org.javarosa.xpath.parser.XPathSyntaxException;
+
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.security.SecureRandom;
+import java.util.Base64;
+
+public class XPathEncryptStringFunc extends XPathFuncExpr {
+    public static final String NAME = "encrypt-string";
+    private static final int EXPECTED_ARG_COUNT = 3;
+
+    public XPathEncryptStringFunc() {
+        name = NAME;
+        expectedArgCount = EXPECTED_ARG_COUNT;
+    }
+
+    public XPathEncryptStringFunc(XPathExpression[] args) throws XPathSyntaxException {
+        super(NAME, args, EXPECTED_ARG_COUNT, true);
+    }
+
+    @Override
+    public Object evalBody(DataInstance model, EvaluationContext evalContext, Object[] evaluatedArgs) {
+        return encryptString(evaluatedArgs[0], evaluatedArgs[1], evaluatedArgs[2]);
+    }
+
+    /**
+     * Encrypt a message with the given algorithm and key.
+     *
+     * @param message    a message to be encrypted
+     * @param key        the key used for encryption
+     * @param algorithm  the encryption algorithm to use
+     */
+    private static String encryptString(Object o1, Object o2, Object o3) {
+        String message = FunctionUtils.toString(o1);
+        String key = FunctionUtils.toString(o2);
+        String algorithm = FunctionUtils.toString(o3);
+
+        final String ENCRYPT_ALGO = "AES/GCM/NoPadding";
+        final int TAG_LENGTH_BIT = 128;
+        final int IV_LENGTH_BYTE = 12;
+        final int KEY_LENGTH_BIT = 256;
+
+        if (!algorithm.equals("AES")) {
+            throw new XPathException("Unknown algorithm \"" + algorithm +
+                                     "\" for " + NAME);
+        }
+
+        Base64.Decoder keyDecoder = Base64.getUrlDecoder();
+        byte[] keyBytes = keyDecoder.decode(key);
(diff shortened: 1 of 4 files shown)


damagatchi:
Can one of the admins verify this patch?

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
```diff
@@ -0,0 +1,84 @@
...
+                                     " bits long, not " + 8 * keyBytes.length);
+        }
+        SecretKey secret = new SecretKeySpec(keyBytes, "AES");
+
+        byte[] iv = new byte[IV_LENGTH_BYTE];
+        new SecureRandom().nextBytes(iv);
+
+        try {
+            Cipher cipher = Cipher.getInstance(ENCRYPT_ALGO);
+            cipher.init(Cipher.ENCRYPT_MODE, secret, new GCMParameterSpec(TAG_LENGTH_BIT, iv));
```
ctsims, line 72: Hm, this might cause issues on Android, because android didn't support GCM schemes until API 19 
https://developer.android.com/reference/javax/crypto/spec/GCMParameterSpec.

We still have devices running older versions, which means that this code probably won't actually even compile for us on Android. 

Is there a way to get the GCM from a polymorphic factory indirection callout like the Cipher instance factory? It's fine for the code to not *work* on Android 4.1, but I think the class signature needs to compile.

I'm trying to find an easy way to make that testable, but it looks like our cross-request CI builder only works with branches that are in our repo directly.

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
```diff
@@ -0,0 +1,84 @@
...
+     * @param algorithm  the encryption algorithm to use
+     */
+    private static String encryptString(Object o1, Object o2, Object o3) {
+        String message = FunctionUtils.toString(o1);
+        String key = FunctionUtils.toString(o2);
+        String algorithm = FunctionUtils.toString(o3);
+
+        final String ENCRYPT_ALGO = "AES/GCM/NoPadding";
+        final int TAG_LENGTH_BIT = 128;
+        final int IV_LENGTH_BYTE = 12;
```
ctsims, line 51: Question: Isn't it normal to use 16 bytes for the IV on GCM or do I have that wrong?

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
```diff
@@ -0,0 +1,84 @@
...
+
+        byte[] iv = new byte[IV_LENGTH_BYTE];
+        new SecureRandom().nextBytes(iv);
+
+        try {
+            Cipher cipher = Cipher.getInstance(ENCRYPT_ALGO);
+            cipher.init(Cipher.ENCRYPT_MODE, secret, new GCMParameterSpec(TAG_LENGTH_BIT, iv));
+            byte[] encryptedMessage = cipher.doFinal(message.getBytes(StandardCharsets.UTF_8));
+            byte[] ivPlusMessage = ByteBuffer.allocate(iv.length + encryptedMessage.length)
+                .put(iv)
```
ctsims, line 75: If we are stuffing a byte array would it be worth writing the IV length as the first arg so that the array doesn't rely on a hardcoded iv length?

ctsims (changes requested):
Hi Charlie,

This looks great overall, but I think we might need to be a bit creative to get the GCM param configured due to android (Trying to figure out how to test that). I also left a few other comments, but they are just thoughts, not blockers.

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
solleks: I based this on some example code and didn't know what the rationale for 12 bytes is. However, it seems that 12 bytes is the default IV length for GCM ([stackoverflow](https://stackoverflow.com/questions/31132162/what-size-of-initialization-vector-needed-for-aes-256-encryption-in-java#:~:text=The%20size%20of%20the%20IV%20depends%20on%20the%20mode%2C%20but,the%20default%20is%20highly%20recommended.), [stackexchange](https://crypto.stackexchange.com/questions/41601/aes-gcm-recommended-iv-size-why-12-bytes)).

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
solleks: I can do that if you like, but unless we think the IV length might change, I'd argue it's no cleaner to start with a 1 byte length than a 12 byte IV.

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
solleks: Some good news here. At encryption time, if we just specify the algorithm as "AES/GCM/NoPadding", it will use the same parameters we were specifying manually - 12 byte initialization vector, generated internally, 128-bit tag. So we can make that code compatible back to [API level 10](https://developer.android.com/guide/topics/security/cryptography#SupportedCipher), I believe.

In the test code, however, I haven't been able to get rid of the GCMParameterSpec. In the past, it seems you could use an IvParameterSpec, but that is causing an exception for me using the crypto provider in my JDK. Does the test code need to compile for Android 4.1 too?

I have not tracked down any factory to make the GCM parameters yet.

solleks:
What do you think about the current state of this PR: no GCMParameterSpec in the encryption function, but one in test code?

I have not been able to get my dev environment running with formplayer and the modified commcare-core yet, so I haven't been able to exercise your form. I'll keep trying.

I renamed the function to 'encrypt' to match what's in the form.

ctsims:
@solleks Abstractly this looks pretty solid to me in the current state. I can test again to make sure everything is good on different android versions.

The only thing I'm a bit hesitant about is the presumption that the internally generated IV is reasonably secure. We use a platform static pattern for similar operations in other contexts, I'll take a swing at that structure in a separate PR when I check this build

ctsims:
@solleks Actually, reading through the Cipher spec, it looks like the guarantee about initialization is quite strong, so I'm aligned with your approach. Leaving feedback on a few more minor notes

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
```diff
@@ -0,0 +1,85 @@
...
+
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+import java.security.SecureRandom;
+
+public class XPathEncryptStringFunc extends XPathFuncExpr {
+    public static final String NAME = "encrypt";
```
ctsims, line 19: I think `encrypt-string` was a stronger name for this method and reduces ambiguity, so I'd propose changing this back (easy enough to change the test form function name)

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
```diff
@@ -0,0 +1,85 @@
...
+            throw new XPathException("Key should be " + KEY_LENGTH_BIT +
+                                     " bits long, not " + 8 * keyBytes.length);
+        }
+        SecretKey secret = new SecretKeySpec(keyBytes, "AES");
+
+        try {
+            Cipher cipher = Cipher.getInstance(ENCRYPT_ALGO);
+            cipher.init(Cipher.ENCRYPT_MODE, secret);
+            byte[] encryptedMessage = cipher.doFinal(message.getBytes(Charset.forName("UTF-8")));
+            byte[] iv = cipher.getIV();
```
ctsims, line 75: Since we are getting this implicitly from the Cipher now, I think we should explicitly confirm that this came out as 12 bytes and failfast if not so that we won't produce uninterpretable messages.

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
ctsims: Fair enough as long as we are consistent with it (left a quick note on that above)

ctsims (changes requested):
This looks good to me with the change you made, and no worries with the test initialization, it's only an issue in the runtime code and Android tests are running fine for me now. 

I noted two minor things, but after that I think this looks good to go.

solleks:
My apologies, but I missed your last comments for a couple of days. I've addressed them now.

I was able to test the encrypt-string function in CommCare HQ using app preview. Here is a form that should prompt for a message and a key, then show the encrypted result (IV + cipher text): [AES_Form.txt](https://github.com/dimagi/commcare-core/files/5499804/AES_Form.txt).

"qyP22cu4DrZ3x/kBll3ahlhMsRUVATP3L1eklc56pro=" is a valid 256-bit key in Base64.

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
solleks: Good point. I've added a length check.

src/main/java/org/javarosa/xpath/expr/XPathEncryptStringFunc.java
solleks: OK, I've changed it back to encrypt-string.

ctsims (approved):

ctsims:
Looks great! Merging up. Thanks for the contribution!

ctsims:
@solleks 

It turns out we're getting a test failure on commcare-android because the default android lib is producing a 16 bit IV.

```
org.javarosa.xpath.XPathException: Exception during encryption: org.javarosa.xpath.XPathException: Initialization vector should be 12 bytes long, not 16
	at org.javarosa.xpath.expr.XPathEncryptStringFunc.encryptString(XPathEncryptStringFunc.java:87)
	at org.javarosa.xpath.expr.XPathEncryptStringFunc.evalBody(XPathEncryptStringFunc.java:33)
	at org.javarosa.xpath.expr.XPathFuncExpr.evalRaw(XPathFuncExpr.java:55)
	at org.javarosa.xpath.expr.XPathExpression.eval(XPathExpression.java:45)
	at org.javarosa.xpath.test.XPathEvalTest.evalExpr(XPathEvalTest.java:102)
	at org.javarosa.xpath.test.XPathEvalTest.encryptAndCompare(XPathEvalTest.java:796)
	at org.javarosa.xpath.test.XPathEvalTest.testEncryptString(XPathEvalTest.java:818)
```

I'm going to revert for now. Not sure if this is an issue with something about Android using the wrong IV, but I'd guess the easiest fix is using the first Byte to write the length of the IV like we were talking about a bit earlier.

solleks:
Ok, I'll make that change.

PR #945, opened 2020-11-02
mark initialization

calellowitz:
I believe https://github.com/dimagi/formplayer/pull/780/files introduced the possibility that an entity screen could be initialized multiple times. This makes sure that if it is called again, it will be a no-op.

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityScreen.java
@@ -54,6 +54,8 @@ public class EntityScreen extends CompoundScreenHost {
 
     private Vector<TreeReference> references;
 
+    private boolean initialized = false;
+
     public EntityScreen(boolean handleCaseIndex) {
         this.handleCaseIndex = handleCaseIndex;
     }
@@ -72,6 +74,9 @@ public class EntityScreen extends CompoundScreenHost {
     }
 
     public void init(SessionWrapper session) throws CommCareSessionException {
+        if (initialized) {
+            return;
+        }
         SessionDatum datum = session.getNeededDatum();
         if (!(datum instanceof EntityDatum)) {
             throw new CommCareSessionException("Didn't find an entity select action where one is expected.");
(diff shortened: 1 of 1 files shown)


ctsims (approved):
Looks good to me, although it might be worth noting that this will result in the SessionWrapper retaining the original evaluationContext from the first pass. I think this will make the EntityScreen safe to init twice, but if it's being asked to init twice with different sessionwrappers it might make the outcome wonky

calellowitz:
> Looks good to me, although it might be worth noting that this will result in the SessionWrapper retaining the original evaluationContext from the first pass. I think this will make the EntityScreen safe to init twice, but if it's being asked to init twice with different sessionwrappers it might make the outcome wonky

I don't think that is a concern because the session wrapper is shared by `MenuSession`, so changes happening in either place should propagate to both places and any reinitialization should use the same shared session wrapper. At least on the formplayer side, not sure if the CLI is slightly different in that regard.

ctsims:
> > Looks good to me, although it might be worth noting that this will result in the SessionWrapper retaining the original evaluationContext from the first pass. I think this will make the EntityScreen safe to init twice, but if it's being asked to init twice with different sessionwrappers it might make the outcome wonky
> 
> I don't think that is a concern because the session wrapper is shared by `MenuSession`, so changes happening in either place should propagate to both places and any reinitialization should use the same shared session wrapper. At least on the formplayer side, not sure if the CLI is slightly different in that regard.

Just to close this out, when we have assumptions like this ("X is fine because Y and Z always produce the same input") I think it's good for us to force the system to fail if the assumption fails. Alternatively it's good to fall back to more conservative behavior, although I think that's an anti-pattern more often than not because performance 'features' can fail silently.

calellowitz:
That is a good point. Do you just mean raising an exception if that happens? We should be able to compare the two session wrappers at that point in the code (the one attached to the `EntityScreen` already, and the one passed to `init`)

ctsims:
@calellowitz Yeah, just doing a comparison and throwing an exception so that we are failing fast, rather than failing in potentially deeply-confusing ways later on when we're primed with one data source then querying another.

PR #947, opened 2020-11-07
Allow variable length initialization vector

solleks:
@ctsims, here is the change to allow variable length initialization vectors. Sorry about the breakage. I believe this will fix it. Currently, it accepts vectors from 1 to 255 bytes, but we could limit that further.


damagatchi:
Can one of the admins verify this patch?

ctsims (approved):

ctsims:
Thanks for the quick turnaround!

PR #949, opened 2020-11-16
Added auto_launch option for actions

orangejenny:
Support for https://github.com/dimagi/formplayer/pull/783

This uses the new "minimal" entity screen code to avoid running all of the usual entity list logic. Since minimal entities don't allow for selecting a case later on, this uses `getEntityFromID` to select the case (see commcare-core PR).

Merging into `formplayer` instead of `master` since mobile will ignore this attribute.


ctsims (approved):

orangejenny:
Updated as described in https://github.com/dimagi/formplayer/pull/783#issuecomment-729832330

PR #954, opened 2020-12-02
Adds Session endpoints model

shubham1g5:
[Spec](https://docs.google.com/document/d/111mch19MnKMcXf91h5dmjvomYnket5x696Io3yLEjcE/edit?ts=5f724532#heading=h.5h6283rsmrrl)

### DIFF
--- a/src/main/java/org/commcare/suite/model/StackFrameStep.java
+++ b/src/main/java/org/commcare/suite/model/StackFrameStep.java
@@ -236,13 +236,4 @@ public class StackFrameStep implements Externalizable {
     public String getElementType() {
         return elementType;
     }
-
-    // find and replace any arguments $argument with their value in {@code arguments}
-    public void substituteArgument(HashMap<String, String> arguments) {
-        if (value.indexOf("$") != -1) {
-            for (Map.Entry<String, String> argsSet : arguments.entrySet()) {
-                value = value.replace("$" + argsSet.getKey(), "'" + argsSet.getValue() + "'");
-            }
-        }
-    }
 }
--- a/src/main/java/org/commcare/suite/model/StackFrameStep.java
+++ b/src/main/java/org/commcare/suite/model/StackFrameStep.java
@@ -18,7 +18,9 @@ import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.Hashtable;
+import java.util.Map;
 
 /**
  * @author ctsims
@@ -233,5 +235,14 @@ public class StackFrameStep implements Externalizable {
     @SuppressWarnings("unused")
     public String getElementType() {
         return elementType;
+    }
+
+    // find and replace any arguments $argument with their value in {@code arguments}
+    public void substituteArgument(HashMap<String, String> arguments) {
+        if (value.indexOf("$") != -1) {
+            for (Map.Entry<String, String> argsSet : arguments.entrySet()) {
+                value = value.replace("$" + argsSet.getKey(), "'" + argsSet.getValue() + "'");
+            }
+        }
     }
 }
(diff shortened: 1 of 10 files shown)


src/main/java/org/commcare/suite/model/Suite.java
```diff
@@ -46,18 +46,20 @@
...
     }
 
     public Suite(int version, Hashtable<String, Detail> details,
-                 Hashtable<String, Entry> entries, Vector<Menu> menus) {
+                 Hashtable<String, Entry> entries, Vector<Menu> menus, Hashtable<String, Endpoint> endpoints) {
         this.version = version;
         this.details = details;
         this.entries = entries;
         this.menus = menus;
+        this.endpoints = endpoints;
```
ShivamPokhriyal, line 62: Should this be `externalized`?

src/main/java/org/commcare/xml/EndpointParser.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final String NAME_ARGUMENT = "argument";
+    private static final String NAME_STACK = "stack";
+
+    public EndpointParser(KXmlParser parser) {
+        super(parser);
+    }
+
+    @Override
+    public Endpoint parse() throws InvalidStructureException, IOException, XmlPullParserException, UnfullfilledRequirementsException {
+        String id = parser.getAttributeValue(null, "id");
```
ShivamPokhriyal, line 26: Can this `id` be null or empty?

ShivamPokhriyal (approved):
Looks good to me.

src/main/java/org/commcare/xml/EndpointParser.java
shubham1g5: it should not be. I should add a validation here to that point. Thanks.

src/main/java/org/commcare/suite/model/Suite.java
shubham1g5: yes, totally missed out on it. Thanks for the catch

ShivamPokhriyal (dismissed):

src/main/java/org/commcare/suite/model/StackFrameStep.java
```diff
@@ -234,4 +236,13 @@ public void setType(String elementType) {
     public String getElementType() {
         return elementType;
     }
+
+    // find and replace any arguments $argument with their value in {@code arguments}
+    public void substituteArgument(HashMap<String, String> arguments) {
+        if (value.indexOf("$") != -1) {
```
ctsims, line 242: I was imagining that we would provide arguments through the normal XPath engine. You can set variables in the Evaluation Context, and they will be evaluated as normal.

Munging the string in place can come with pretty complex issues, especially if the user inputs are complex

ctsims (changes requested):
Made a note on the way we introduce and process arguemnts.

src/main/java/org/commcare/suite/model/StackFrameStep.java
```diff
@@ -18,7 +18,9 @@
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.Enumeration;
+import java.util.HashMap;
```
ctsims, line 21: Nit: These are no-ops now

src/main/java/org/commcare/xml/EndpointParser.java
```diff
@@ -0,0 +1,52 @@
...
+import org.xmlpull.v1.XmlPullParserException;
+
+import java.io.IOException;
+import java.util.Vector;
+
+public class EndpointParser extends ElementParser<Endpoint> {
+
+    static final String NAME_ENDPOINT = "endpoint";
+    private static final String NAME_ARGUMENT = "argument";
+    private static final String NAME_STACK = "stack";
```
ctsims, line 18: Suggestion: Moving this ID to the StackOpsParser and re-using the static String between this and the other stack ops parsing contexts.

ctsims (approved):
Looks good! Optional suggestions in feedback.

ctsims (approved):
Looks great! Thanks again for all of the throughness on getting this implemented!

PR #957, opened 2020-12-07
index state category

calellowitz:
This adds two new indexed case properties (`state` and `category`) with two aliases (`current_status` and `patient_type`, respectively). It also adds functionality to allow indexed lookups for `!=` constructions, and implements support for that in the CLI data store.

### DIFF
--- a/src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
+++ b/src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
@@ -27,6 +27,8 @@ public class CaseInstanceTreeElement extends StorageInstanceTreeElement<Case, Ca
     private final static XPathPathExpr CASE_INDEX_EXPR = XPathReference.getPathExpr("index/*");
     private final static XPathPathExpr OWNER_ID_EXPR = XPathReference.getPathExpr("@owner_id");
     private final static XPathPathExpr EXTERNAL_ID_EXPR = XPathReference.getPathExpr("@external_id");
+    private final static XPathPathExpr CATEGORY_EXPR = XPathReference.getPathExpr("category");
+    private final static XPathPathExpr STATE_EXPR = XPathReference.getPathExpr("state");
     private final static XPathPathExpr PATIENT_TYPE_EXPR = XPathReference.getPathExpr("patient_type");
     private final static XPathPathExpr CURRENT_STATUS_EXPR = XPathReference.getPathExpr("current_status");
 
@@ -72,8 +74,10 @@ public class CaseInstanceTreeElement extends StorageInstanceTreeElement<Case, Ca
         indices.put(CASE_INDEX_EXPR, Case.INDEX_CASE_INDEX_PRE);
         indices.put(OWNER_ID_EXPR, Case.INDEX_OWNER_ID);
         indices.put(EXTERNAL_ID_EXPR, Case.INDEX_EXTERNAL_ID);
-        indices.put(PATIENT_TYPE_EXPR, Case.INDEX_PATIENT_TYPE);
-        indices.put(CURRENT_STATUS_EXPR, Case.INDEX_CURRENT_STATUS);
+        indices.put(CATEGORY_EXPR, Case.INDEX_CATEGORY);
+        indices.put(STATE_EXPR, Case.INDEX_STATE);
+        indices.put(PATIENT_TYPE_EXPR, Case.INDEX_CATEGORY);
+        indices.put(CURRENT_STATUS_EXPR, Case.INDEX_STATE);
 
         return indices;
     }
--- a/src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
+++ b/src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
@@ -27,6 +27,8 @@ public class CaseInstanceTreeElement extends StorageInstanceTreeElement<Case, Ca
     private final static XPathPathExpr CASE_INDEX_EXPR = XPathReference.getPathExpr("index/*");
     private final static XPathPathExpr OWNER_ID_EXPR = XPathReference.getPathExpr("@owner_id");
     private final static XPathPathExpr EXTERNAL_ID_EXPR = XPathReference.getPathExpr("@external_id");
+    private final static XPathPathExpr PATIENT_TYPE_EXPR = XPathReference.getPathExpr("patient_type");
+    private final static XPathPathExpr CURRENT_STATUS_EXPR = XPathReference.getPathExpr("current_status");
 
     public CaseInstanceTreeElement(AbstractTreeElement instanceRoot,
                                    IStorageUtilityIndexed<Case> storage) {
@@ -70,6 +72,8 @@ public class CaseInstanceTreeElement extends StorageInstanceTreeElement<Case, Ca
         indices.put(CASE_INDEX_EXPR, Case.INDEX_CASE_INDEX_PRE);
         indices.put(OWNER_ID_EXPR, Case.INDEX_OWNER_ID);
         indices.put(EXTERNAL_ID_EXPR, Case.INDEX_EXTERNAL_ID);
+        indices.put(PATIENT_TYPE_EXPR, Case.INDEX_PATIENT_TYPE);
+        indices.put(CURRENT_STATUS_EXPR, Case.INDEX_CURRENT_STATUS);
 
         return indices;
     }
--- a/src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
+++ b/src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
@@ -27,10 +27,8 @@ public class CaseInstanceTreeElement extends StorageInstanceTreeElement<Case, Ca
     private final static XPathPathExpr CASE_INDEX_EXPR = XPathReference.getPathExpr("index/*");
     private final static XPathPathExpr OWNER_ID_EXPR = XPathReference.getPathExpr("@owner_id");
     private final static XPathPathExpr EXTERNAL_ID_EXPR = XPathReference.getPathExpr("@external_id");
-    private final static XPathPathExpr CATEGORY_EXPR = XPathReference.getPathExpr("category");
-    private final static XPathPathExpr STATE_EXPR = XPathReference.getPathExpr("state");
-    private final static XPathPathExpr PATIENT_TYPE_EXPR = XPathReference.getPathExpr("patient_type");
-    private final static XPathPathExpr CURRENT_STATUS_EXPR = XPathReference.getPathExpr("current_status");
+    private final static XPathPathExpr CATEGORY_EXPR = XPathReference.getPathExpr("@category");
+    private final static XPathPathExpr STATE_EXPR = XPathReference.getPathExpr("@state");
 
     public CaseInstanceTreeElement(AbstractTreeElement instanceRoot,
                                    IStorageUtilityIndexed<Case> storage) {
@@ -76,8 +74,6 @@ public class CaseInstanceTreeElement extends StorageInstanceTreeElement<Case, Ca
         indices.put(EXTERNAL_ID_EXPR, Case.INDEX_EXTERNAL_ID);
         indices.put(CATEGORY_EXPR, Case.INDEX_CATEGORY);
         indices.put(STATE_EXPR, Case.INDEX_STATE);
-        indices.put(PATIENT_TYPE_EXPR, Case.INDEX_CATEGORY);
-        indices.put(CURRENT_STATUS_EXPR, Case.INDEX_STATE);
 
         return indices;
     }
--- a/src/main/java/org/commcare/cases/model/Case.java
+++ b/src/main/java/org/commcare/cases/model/Case.java
@@ -30,6 +30,8 @@ public class Case implements Persistable, IMetaData {
     public static final String EXTERNAL_ID_KEY = "external_id";
     public static final String CATEGORY_KEY = "category";
     public static final String STATE_KEY = "state";
+    public static final String PATIENT_TYPE_KEY = "patient_type";
+    public static final String CURRENT_STATUS_KEY = "current_status";
     public static final String STORAGE_KEY = "CASE";
 
     public static final String INDEX_CASE_ID = "case-id";
@@ -123,7 +125,11 @@ public class Case implements Persistable, IMetaData {
     }
 
     public String getCategory() {
-        return (String)data.get(CATEGORY_KEY);
+        String category = (String)data.get(CATEGORY_KEY);
+        if (category == null) {
+            category = (String)data.get(PATIENT_TYPE_KEY);
+        }
+        return category;
     }
 
     public void setCategory(String id) {
@@ -131,7 +137,11 @@ public class Case implements Persistable, IMetaData {
     }
 
     public String getState() {
-        return (String)data.get(STATE_KEY);
+        String state = (String)data.get(STATE_KEY);
+        if (state == null) {
+            state = (String)data.get(CURRENT_STATUS_KEY);
+        }
+        return state;
     }
 
     public void setState(String id) {
--- a/src/main/java/org/commcare/cases/model/Case.java
+++ b/src/main/java/org/commcare/cases/model/Case.java
@@ -30,8 +30,12 @@ public class Case implements Persistable, IMetaData {
     public static final String EXTERNAL_ID_KEY = "external_id";
     public static final String CATEGORY_KEY = "category";
     public static final String STATE_KEY = "state";
+
+    // Allowed as alias for preferred property category
     public static final String PATIENT_TYPE_KEY = "patient_type";
+    // Allowed as alias for preferred property state
     public static final String CURRENT_STATUS_KEY = "current_status";
+
     public static final String STORAGE_KEY = "CASE";
 
     public static final String INDEX_CASE_ID = "case-id";
--- a/src/main/java/org/commcare/cases/model/Case.java
+++ b/src/main/java/org/commcare/cases/model/Case.java
@@ -28,8 +28,8 @@ import java.util.Vector;
 public class Case implements Persistable, IMetaData {
     public static final String USER_ID_KEY = "userid";
     public static final String EXTERNAL_ID_KEY = "external_id";
-    public static final String PATIENT_TYPE_KEY = "patient_type";
-    public static final String CURRENT_STATUS_KEY = "currrent_status";
+    public static final String CATEGORY_KEY = "category";
+    public static final String STATE_KEY = "state";
     public static final String STORAGE_KEY = "CASE";
 
     public static final String INDEX_CASE_ID = "case-id";
@@ -38,8 +38,8 @@ public class Case implements Persistable, IMetaData {
     public static final String INDEX_CASE_INDEX_PRE = "case-in-";
     public static final String INDEX_OWNER_ID = "owner-id";
     public static final String INDEX_EXTERNAL_ID = "external-id";
-    public static final String INDEX_PATIENT_TYPE = "patient-type";
-    public static final String INDEX_CURRENT_STATUS = "current-status";
+    public static final String INDEX_CATEGORY = "category";
+    public static final String INDEX_STATE = "state";
 
     private static final String ATTACHMENT_PREFIX = "attachmentdata";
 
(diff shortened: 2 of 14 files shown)


src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
```diff
@@ -146,8 +147,13 @@ private void collectNativePredicateProfiles(Vector<XPathExpression> predicates,
                             //sure the best way to do that....? Maybe tell the evaluation context to skip out here if it detects a request
                             //to resolve in a certain area?
                             Object o = FunctionUtils.unpack(((XPathEqExpr)xpe).b.eval(evalContext));
-                            optimizations.addElement(new IndexedValueLookup(filterIndex, o));
-
+                            PredicateProfile lookup;
+                            if (((XPathEqExpr)xpe).op == XPathEqExpr.EQ) {
+                                lookup = new IndexedValueLookup(filterIndex, o);
+                            } else {
```
ctsims, line 153: This should be a positive match for `== NEQ`, and if that fails it should either throw a not-implemented exception or (more likely) should just fail to trigger an optimization, otherwise any new `eq` expression type will be optimized as a negative indexed value lookup.

src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
```diff
@@ -27,6 +27,10 @@
     private final static XPathPathExpr CASE_INDEX_EXPR = XPathReference.getPathExpr("index/*");
     private final static XPathPathExpr OWNER_ID_EXPR = XPathReference.getPathExpr("@owner_id");
     private final static XPathPathExpr EXTERNAL_ID_EXPR = XPathReference.getPathExpr("@external_id");
+    private final static XPathPathExpr CATEGORY_EXPR = XPathReference.getPathExpr("category");
```
ctsims, line 30: We discussed these as being `@` expressions (IE: In the attribute set for the `<case>` node rather than a child element), right? 

Is there a reason you went with this structure?

src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
```diff
@@ -27,6 +27,10 @@
...
+    private final static XPathPathExpr STATE_EXPR = XPathReference.getPathExpr("state");
+    private final static XPathPathExpr PATIENT_TYPE_EXPR = XPathReference.getPathExpr("patient_type");
```
ctsims, line 32: I think you don't actually want to add these here for aliasing, see my comment on the Case model below

src/main/java/org/commcare/cases/model/Case.java
```diff
@@ -118,6 +122,22 @@ public void setExternalId(String id) {
         data.put(EXTERNAL_ID_KEY, id);
     }
 
+    public String getCategory() {
+        return (String)data.get(CATEGORY_KEY);
```
ctsims, line 126: I think this is where you want to be aliasing the values. 

IE: The "Category" index should be a lookup of the key `category` if that key exists in the set, or a lookup of the key `patient_type` if it doesn't.

Alternatively you could do that aliasing in the `getMetaFields` block. I'm not sure, honestly, which one would result in less confusing behavior for someone who's not expecting it.

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
```diff
@@ -291,23 +297,41 @@ protected void initBasicQueryHandlers(QueryPlanner queryPlanner) {
...
+            keyDescription += name + "|";
         }
+        String[] namesArray = namesToMatch.toArray(new String[namesToMatch.size()]);
+        String[] valuesArray = valuesToMatch.toArray(new String[valuesToMatch.size()]);
+        String[] inverseNames = namesToInverseMatch.toArray(new String[namesToInverseMatch.size()]);
+        String[] inverseValues = valuesToInverseMatch.toArray(new String[valuesToInverseMatch.size()]);
         mMostRecentBatchFetch = new String[2][];
-        mMostRecentBatchFetch[0] = namesToMatch;
-        mMostRecentBatchFetch[1] = valuesToMatch;
+        mMostRecentBatchFetch[0] = namesArray;
```
ctsims, line 333: You'll need to make sure the NegativeIndexedValue lookups make it into this key set as well. This is what is used to index the cache of results, so if the negative indexes aren't in there, if you first query `[a = b][c != d]` then query `[a=b]` it'll give you the cached result from the first query.

src/main/java/org/commcare/xml/CaseXmlParser.java
```diff
@@ -204,6 +204,18 @@ private void updateCase(Case caseForBlock, String caseId) throws InvalidStructur
                 case "external_id":
                     caseForBlock.setExternalId(value);
                     break;
+                case "category":
+                    caseForBlock.setCategory(value);
+                    break;
+                case "state":
+                    caseForBlock.setState(value);
+                    break;
+                case "patient_type":
```
ctsims, line 213: You should remove the `patient_type` and `current_status` aliasing, here, otherwise the system won't set the original values.

We still want `patient_type` to update the `patient_type` key in the case, we just also want it to populate a totally separate index.

src/main/java/org/commcare/xml/CaseXmlParser.java
```diff
@@ -204,6 +204,18 @@ private void updateCase(Case caseForBlock, String caseId) throws InvalidStructur
                 case "external_id":
                     caseForBlock.setExternalId(value);
                     break;
+                case "category":
```
ctsims, line 207: (non-actionable) I have mixed feelings about keeping these here (since they aren't doing anything other than the core code path), but it makes sense to stay consistent with the `external_id` behavior.

src/main/java/org/commcare/xml/bulk/BulkProcessingCaseXmlParser.java
```diff
@@ -213,6 +213,18 @@ private void updateCase(TreeElement updateElement,
                 case "external_id":
                     caseForBlock.setExternalId(value);
                     break;
+                case "category":
+                    caseForBlock.setCategory(value);
+                    break;
+                case "state":
+                    caseForBlock.setState(value);
+                    break;
+                case "patient_type":
```
ctsims, line 222: Same as comment above

src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java
```diff
@@ -193,6 +193,23 @@
...
+    *
+    * @param metaFieldNames A list of metadata field names to match
+    * @param values     The values which must match the field names provided
+    * @param inverseFieldNames A list of field names to inverse match (!=)
+    * @param inverseValues     The values which the field must not match
+    * @param returnSet A LinkedHashSet of integers which match the return value
+    * @return A Vector of Integers such that retrieving the Externalizable object with any
+    * of those integer IDs will result in an object for which the fields specified are equal
+    * to the value provided.
+    * @throws RuntimeException (Fix this exception type) if the field is unrecognized by the
```
ctsims, line 208: I'd clarify here that the exception will be thrown if any field is unrecognized

src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java
```diff
@@ -193,6 +193,23 @@
...
+    /**
+    * Retrieves a Vector of IDs of Externalizable objects in storage for which the field
+    * specified contains the values specified.
+    *
+    * @param metaFieldNames A list of metadata field names to match
+    * @param values     The values which must match the field names provided
+    * @param inverseFieldNames A list of field names to inverse match (!=)
+    * @param inverseValues     The values which the field must not match
+    * @param returnSet A LinkedHashSet of integers which match the return value
+    * @return A Vector of Integers such that retrieving the Externalizable object with any
```
ctsims, line 205: The behavior of this method is ambiguous if no keys are provided at all (IE: all incoming arrays are empty). Presumably, for instance, you can provide empty positive match fields, and a single negative match `a !='b'`, and get all the records where `a` doesn't equal `b`. From that logic, if you then remove the negative match you might expect to just get all of the records in the storage utility.

I'd clarify whether that's an error condition, or what to expect if not.

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
```diff
@@ -134,7 +135,7 @@ private void collectNativePredicateProfiles(Vector<XPathExpression> predicates,
         for (XPathExpression xpe : predicates) {
             //what we want here is a static evaluation of the expression to see if it consists of evaluating
             //something we index with something static.
-            if (xpe instanceof XPathEqExpr && ((XPathEqExpr)xpe).op == XPathEqExpr.EQ) {
+            if (xpe instanceof XPathEqExpr) {
```
ctsims, line 138: One thing that might also be worth evaluating is the pattern

`not(a=b)`

I don't remember if the `not` function is parsed as a straight-up xpath function or whether it's parsed as syntactic sugar, but if it's the former it should be fairly straightforward to just check `xpe instanceof XPathNotFunctexpr` or whatnot, and then check the single argument as being a `XPathEqExpr` with an EQ op like we do here.

ctsims (changes requested):
Initial feedback inline. This should also come with tests that trigger the core conditions. You can see some examples here:

https://github.com/dimagi/commcare-core/blob/master/src/test/java/org/commcare/cases/test/CaseXPathQueryTest.java

src/main/java/org/commcare/modern/database/DatabaseHelper.java
```diff
@@ -34,6 +34,20 @@
...
                                                      EncryptedModel em, Persistable p)
             throws IllegalArgumentException {
+        return createWhere(fieldNames, values, null, null, em, p);
+    }
+
+     public static Pair<String, String[]> createWhere(String[] fieldNames, Object[] values,
+                                                     String[] inverseFieldNames, Object[] inverseValues,
+                                                     Persistable p)
+             throws IllegalArgumentException {
+         return createWhere(fieldNames, values, null, null, null, p);
```
ctsims, line 44: The inverse values and inverse field names should still be getting passed on here.

src/main/java/org/commcare/modern/database/DatabaseHelper.java
```diff
@@ -76,6 +90,26 @@
...
+                    if (!fields.contains(columnName)) {
+                        continue;
+                    }
+                }
+                if (set) {
+                    stringBuilder.append(" AND ");
+                }
+
+                stringBuilder.append(columnName);
+                stringBuilder.append("!=?");
```
ctsims, line 106: SQL has two inequality operators `<>` and `!=`. I'm not sure if there's a meaningful difference between different engines, is that something you've evaluated and we have a clear sense of?

src/main/java/org/commcare/modern/database/DatabaseHelper.java
```diff
@@ -76,6 +90,26 @@
 
             set = true;
         }
+        if (inverseFieldNames != null) {
```
ctsims, line 93: This should never be null, I think, it should be an empty array.

This code in general should be refactored so that `fieldNames` and `inverseFieldNames` are treated equivalently, so if there's a null or empty check on the latter, there should be one on the former (people should be able to run a query with only inverse matches)

src/main/java/org/commcare/modern/database/DatabaseHelper.java
```diff
@@ -34,6 +34,20 @@
     public static Pair<String, String[]> createWhere(String[] fieldNames, Object[] values,
                                                      EncryptedModel em, Persistable p)
             throws IllegalArgumentException {
+        return createWhere(fieldNames, values, null, null, em, p);
```
ctsims, line 37: As per the below comment in `createWhere`, my intuition is that these should be instantiated as empty arrays, not null values.

src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
calellowitz: I wanted the same expression to work before and after the change, so projects would be able to make the ordering changes now without having to worry about when this gets deployed, but i do like the idea of `@` for consistency with other indexed properties, so see both sides of this.

src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
calellowitz: So the idea would be that you have to access the index with `@category` but it will find `patient_type` on the backend, rather than either expression being picked up? I did it up here because that is how `CASE_ID_EXPR_TWO` does it and I was basing the aliasing off that. It seems like this is necessary for either expression to be accepted in the app, but if that is not what you had intended, I can remove this.

src/main/java/org/commcare/xml/CaseXmlParser.java
calellowitz: I felt the same way. I am confused why those `case`s exist since the data is saved the same as the `default`

src/main/java/org/commcare/modern/database/DatabaseHelper.java
calellowitz: They are equivalent for postgres and sqlite, happy to use either one if we have a convention.

src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java
calellowitz: you just mean change `the` to `any`?

src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java
calellowitz: Do you know what existing behavior is if no values are passed? I imagine that isn't something the code ever does (it doesn't make sense), but expect this to behave similarly.

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
calellowitz: Where is this used? I could not actually find the usages of this field, just a few places where it is set.

calellowitz:
@ctsims I think I got to all your feedback except adding the test. I left a couple questions, but assuming this approach looks good, I will add those.

src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
ctsims: My intuition is that I don't think it's safe for us to index individual properties, and I also don't' want to set an expectation with project teams that we'll configure request contracts in a way that fits their custom workflows and improve performance in place. 

The property aliasing is already more custom than I'm comfortable with, but we will at least be able to gracefully move away from that. Adding the custom query indexes feels a "forever" change that I'm not ok making part of CommCare's permanent request contract.

src/main/java/org/commcare/cases/instance/CaseInstanceTreeElement.java
ctsims: `CASE_ID_EXPR` and `CASE_ID_EXPR_TWO` are semantically identical, so they aren't aliases in the same way that these would be. The new "alias" approach would be like what you described. The `@category` attribute would be a virtual property that is present if either `/category` or `/patient_type` are defined, and contains the value of `/category` if it exists, and `/patient_type` if it doesn't.

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
ctsims: This may only be used in platform specific code. android appears to have a request path for it that goes through the following:
https://github.com/dimagi/commcare-android/blob/f1f383e478850de92418e8f1057422b3ecd84595/app/src/org/commcare/engine/cases/AndroidCaseInstanceTreeElement.java#L237

https://github.com/dimagi/commcare-android/blob/f1f383e478850de92418e8f1057422b3ecd84595/app/src/org/commcare/models/AsyncNodeEntityFactory.java#L103

I believe it's also used in Child Index resolution, but would need to look deeper through the call stack

src/main/java/org/javarosa/core/services/storage/IStorageUtilityIndexed.java
ctsims: I believe the current code will return a full set iterator because it treats the query as a filter (Semantically: "Return all results which meet all filters" with an empty set of filters).

ctsims (changes requested):

ctsims (changes requested):
Asked for a few smaller changes (one more of a consideration than a request)

src/test/java/org/commcare/cases/test/CaseXPathQueryTest.java
```diff
@@ -118,4 +118,34 @@ public void caseQueryNotEqualsTest() throws Exception {
...
+    public void caseIndexAliasTest() throws Exception {
+        config.parseIntoSandbox(
+                this.getClass().getResourceAsStream("/case_query_testing.xml"), sandbox);
+        EvaluationContext ec =
+                MockDataUtils.buildContextWithInstance(sandbox, "casedb",
+                        CaseTestUtils.CASE_INSTANCE);
+
+        Assert.assertTrue(CaseTestUtils.xpathEvalAndCompare(ec,
+                "count(instance('casedb')/casedb/case[@category = 'real'])", 1.0));
+        Assert.assertTrue(CaseTestUtils.xpathEvalAndCompare(ec,
```
ctsims, line 132: I think you still need a test for the third (joint indexing) case above

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
```diff
@@ -291,23 +299,43 @@ protected void initBasicQueryHandlers(QueryPlanner queryPlanner) {
...
-        mMostRecentBatchFetch[1] = valuesToMatch;
+        String[] namesArray = namesToMatch.toArray(new String[namesToMatch.size()]);
+        String[] valuesArray = valuesToMatch.toArray(new String[valuesToMatch.size()]);
+        String[] inverseNames = namesToInverseMatch.toArray(new String[namesToInverseMatch.size()]);
+        String[] inverseValues = valuesToInverseMatch.toArray(new String[valuesToInverseMatch.size()]);
+        mMostRecentBatchFetch = new String[4][];
+        mMostRecentBatchFetch[0] = namesArray;
+        mMostRecentBatchFetch[1] = valuesArray;
+        mMostRecentBatchFetch[2] = inverseNames;
+        mMostRecentBatchFetch[3] = inverseValues;
```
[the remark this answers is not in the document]
ctsims: Bump on this one again too. It's fine if you don't have time to jump into it, but wanted to flag just in case it was possible.

ctsims (changes requested):
Hey Cal,

Just one quick test to add (and one optional point to follow up on) and this looks good to me.

src/main/java/org/commcare/cases/util/StorageBackedTreeRoot.java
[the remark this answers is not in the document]
calellowitz: I thought I replied here, but must have actually sent it over slack. I am hesitant to jump into this because this object is only used in the mobile code (the only use in formplayer code is to clear it, it is never read), and I don't feel like I have enough context to make the correct changes, since I do not really know how it is used there.

ctsims (approved):
All of my feedback has been addressed. I'd still like a deprecation plan outlined somewhere actionable to remove the aliases. 

@calellowitz Confirming you are aware-of and will drive next steps on sequencing a breaking db rollout on formplayer?

Finally - Can you update any named sets of indexed fields here: https://github.com/dimagi/commcare-core/wiki/Technical-Details-on-Performance ? It's fine to list that they are only supported on Formplayer for now, and it can be a task for the Android team to update the docs when a mobile release goes out which merges and migrates the new indexes).

calellowitz:
Finally going forward with this today. Will update the wiki as soon it is merged. As far as rollout steps, I am already coordinating to have all the USH user restores primed, which should be a significant enough percentage of global usage to minimize the impact. In terms of deprecating the aliases, that's a much harder question because of existing data in these projects. We can push new projects to use the proper forms, including focusing our documentation there, but the aliases will need to stick around for the existing data.

src/main/java/org/commcare/cases/model/Case.java
```diff
@@ -28,6 +28,10 @@
 public class Case implements Persistable, IMetaData {
     public static final String USER_ID_KEY = "userid";
     public static final String EXTERNAL_ID_KEY = "external_id";
+    public static final String CATEGORY_KEY = "category";
+    public static final String STATE_KEY = "state";
+    public static final String PATIENT_TYPE_KEY = "patient_type";
```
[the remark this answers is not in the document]
shubham1g5: were these aliases added because we didn't want to make an app change to name the properties category and state instead ?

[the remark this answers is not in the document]
ctsims: Yes, which I think is not a great outcome but we can evenetually deprecate.

Externally speaking we can only communicate about the updates to `category` and `state` and as soon as no remaining internal projects are using the aliases we can drop them.

PR #958, opened 2020-12-09
Added translation for Case Claim title on the 'search all cases' page

stephherbers:
Goes along with other PRs including this one https://github.com/dimagi/commcare-hq/compare/smh/case-claim-translation?expand=1

Tested locally and I was able to changes "Case Claim" and change it back successfully.

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -7,6 +7,8 @@ import org.commcare.session.RemoteQuerySessionManager;
 import org.commcare.suite.model.DisplayUnit;
 import org.javarosa.core.model.instance.ExternalDataInstance;
 import org.javarosa.core.util.OrderedHashtable;
+import org.javarosa.core.services.locale.Localization;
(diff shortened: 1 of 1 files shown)


orangejenny (approved):
Looking at other calls of `Localization.get`, we don't typically check for a `NoLocalizedTextException`, since if the translation key is added to the necessary places, there isn't really a risk of it being missing.

Also, until there's a PR into master, mobile isn't going to use the new translation.

I think this PR makes sense to release this week, and there should be a followup that removes the try/catch and is merged into both the `master` and `formplayer` branches.

stephherbers:
> I think this PR makes sense to release this week, and there should be a followup that removes the try/catch and is merged into both the `master` and `formplayer` branches.

Sure-- sounds like a plan

calellowitz:
I am reverting this. I just got the following exception on staging `The Localizer could not find a definition for ID: case.search.title in the 'en' locale.`, and need to roll out some other features that will require a commcare-core update. I can't confirm these are related but it seems likely, and since it isn't updated in the other repo anyway, it seems harmless to revert.

PR #959, opened 2020-12-09
Allow Default Expressions for Query Screen Prompts

ctsims:
Allows users to define a default expression for [Remote Request Query prompts](https://github.com/dimagi/commcare-core/wiki/Suite20#remote-request) which will be preloaded in before the user begins providing values.

### DIFF
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -53,7 +53,11 @@ public class RemoteQuerySessionManager {
         for (Enumeration en = queryPrompts.keys(); en.hasMoreElements(); ) {
             String promptId = (String)en.nextElement();
             QueryPrompt prompt = queryPrompts.get(promptId);
-            userAnswers.put(prompt.getKey(), "");
+            String defaultValue = "";
+            if(prompt.getDefaultValueExpr() != null) {
+                defaultValue = FunctionUtils.toString(prompt.getDefaultValueExpr().eval(evaluationContext));
+            }
+            userAnswers.put(prompt.getKey(), defaultValue);
         }
     }
 
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -15,6 +15,7 @@ import org.javarosa.xml.ElementParser;
 import org.javarosa.xml.TreeElementParser;
 import org.javarosa.xml.util.InvalidStructureException;
 import org.javarosa.xml.util.UnfullfilledRequirementsException;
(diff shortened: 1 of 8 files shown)


src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -53,7 +53,11 @@ private void initUserAnswers() {
         for (Enumeration en = queryPrompts.keys(); en.hasMoreElements(); ) {
             String promptId = (String)en.nextElement();
             QueryPrompt prompt = queryPrompts.get(promptId);
-            userAnswers.put(prompt.getKey(), "");
+            String defaultValue = "";
+            if(prompt.getDefaultValueExpr() != null) {
+                defaultValue = FunctionUtils.toString(prompt.getDefaultValueExpr().eval(evaluationContext));
```
shubham1g5, line 58: Since this might result in XPathExceptions, we should make sure top level calls to this path are catching the exception and showing the error to the user.

src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Tests for basic app models for case claim
+ *
+ * @author ctsims
+ */
+public class CaseClaimModelTests {
```
shubham1g5, line 20: If the idea here is just to test the parsing, it seems to be happening already as part of [AppStructureTests](https://github.com/dimagi/commcare-core/blob/master/src/test/java/org/commcare/backend/suite/model/test/AppStructureTests.java) today.

shubham1g5 (changes requested):

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
ctsims: Good catch! I added the throws to the method signature for clarity, and added a catch for this in the Android PR.

src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java
ctsims: It is to test parsing for now.

I created a new config because those tests were catching the issues passively. So if, say, someone removed the remote query definition (Say, because they weren't sure what in that suite was part of the actual tests), it wouldn't actually trigger a failure. 

It can be a bit tough with the really dense app files to know when and how changes can be made, so I figured if we were going to extend this functionality at all it probably would be easier to have a smaller, more focused test case.

shubham1g5 (approved):
Thanks, looks great.

PR #962, opened 2020-12-17
Added translation for Case Claim

stephherbers:
As pointed out in Jenny's comment [here](https://github.com/dimagi/commcare-core/pull/958), this is PR into master is needed for the next mobile.

PR #963, opened 2020-12-17
Removed try/except from Case Claim translation

stephherbers:
As pointed out in Jenny's comment [here](https://github.com/dimagi/commcare-core/pull/958), the try/catch is not necessary/misleading. 

This is the [other PR](https://github.com/dimagi/commcare-core/pull/962) connected to that comment

Tested locally and it works the same as before on an app with the modified title and another app that does not modify the "Case Claim" title.


orangejenny (approved):

calellowitz:
I am reverting this. I just got the following exception on staging `The Localizer could not find a definition for ID: case.search.title in the 'en' locale.`, and need to roll out some other features that will require a commcare-core update. I can't confirm these are related but it seems likely, and since it isn't updated in the other repo anyway, it seems harmless to revert.

PR #964, opened 2020-12-21
No badges

calellowitz:
This PR makes the value of badges on a menu screen lazily evaluated. Currently when formplayer walks through your menu session it evaluates them when constructing the main screen, which means for apps with badges every request is quite slow since all navigations goes through the home screen.

### DIFF
--- a/src/cli/java/org/commcare/util/screen/MenuScreen.java
+++ b/src/cli/java/org/commcare/util/screen/MenuScreen.java
@@ -34,11 +34,22 @@ public class MenuScreen extends Screen {
     private String mTitle;
 
     public String[] getBadges() {
+        if (badges == null) {
+            calculateBadges();
+        }
         return badges;
     }
 
     public void setBadges(String[] badges) {
         this.badges = badges;
+    }
+
+    private void calculateBadges() {
+        badges = new String[mChoices.length];
+        for (int i = 0; i < mChoices.length; i++) {
+            MenuDisplayable menu = mChoices[i];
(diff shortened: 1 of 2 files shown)


src/cli/java/org/commcare/util/screen/MenuScreen.java
```diff
@@ -34,13 +34,24 @@
...
+        }
         return badges;
     }
 
     public void setBadges(String[] badges) {
         this.badges = badges;
     }
 
+    private void calculateBadges() {
+        badges = new String[mChoices.length];
```
orangejenny, line 48: I don't have a clue if we have a convention about this, but I think using `this.memberVar` instead of `memberVar` is more readable.

src/main/java/org/commcare/suite/model/MenuLoader.java
```diff
@@ -42,6 +42,16 @@ public MenuLoader(CommCarePlatform platform, SessionWrapperInterface sessionWrap
         this.getMenuDisplayables(platform, sessionWrapper, menuId, hideTrainingRoot);
     }
 
+    public MenuLoader(CommCarePlatform platform, SessionWrapperInterface sessionWrapper,
```
orangejenny, line 45: Can the pre-existing constructor use the convention of calling this one and passing a value for `includeBadges`? It's not a lot of duplicate code, but it's going to be easy for a future dev to update one of the contructors instead of both.

orangejenny (approved):

PR #966, opened 2020-12-28
Merge master into formplayer

orangejenny:
There are several case claim improvements not yet available to formplayer, notably https://github.com/dimagi/commcare-core/pull/959 which is needed soon for the new USH dedup workflow.

I'm hoping to deploy this the week of the 4th and also deploy https://github.com/dimagi/formplayer/pull/787 as per Shubham's [comment](https://github.com/dimagi/formplayer/pull/787#issuecomment-743177409).

@shubham1g5 @ShivamPokhriyal @calellowitz @ctsims could you each look over your commits and approve this?

I am not currently planning to QA this but am open to disagreement.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ dependencies {
     compile 'joda-time:joda-time:2.9.4'
     compile 'com.carrotsearch:hppc:0.7.2'
     compile 'com.squareup.retrofit2:retrofit:2.3.0'
-    compile 'com.squareup.okhttp3:okhttp:4.4.1'
+    compile 'com.squareup.okhttp3:okhttp:3.12.12'
     compile 'com.google.code.findbugs:jsr305:3.0.2'
     compile 'io.reactivex.rxjava2:rxjava:2.1.1'
 
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -15,6 +15,7 @@ import org.javarosa.xml.ElementParser;
 import org.javarosa.xml.TreeElementParser;
 import org.javarosa.xml.util.InvalidStructureException;
 import org.javarosa.xml.util.UnfullfilledRequirementsException;
+import org.javarosa.xpath.XPathException;
 import org.javarosa.xpath.expr.FunctionUtils;
 import org.javarosa.xpath.expr.XPathExpression;
 import org.kxml2.io.KXmlParser;
@@ -42,23 +43,27 @@ public class RemoteQuerySessionManager {
             new Hashtable<>();
 
     private RemoteQuerySessionManager(RemoteQueryDatum queryDatum,
-                                      EvaluationContext evaluationContext) {
+                                      EvaluationContext evaluationContext) throws XPathException {
         this.queryDatum = queryDatum;
         this.evaluationContext = evaluationContext;
         initUserAnswers();
     }
 
-    private void initUserAnswers() {
+    private void initUserAnswers() throws XPathException {
         OrderedHashtable<String, QueryPrompt> queryPrompts = queryDatum.getUserQueryPrompts();
         for (Enumeration en = queryPrompts.keys(); en.hasMoreElements(); ) {
             String promptId = (String)en.nextElement();
             QueryPrompt prompt = queryPrompts.get(promptId);
-            userAnswers.put(prompt.getKey(), "");
+            String defaultValue = "";
+            if(prompt.getDefaultValueExpr() != null) {
+                defaultValue = FunctionUtils.toString(prompt.getDefaultValueExpr().eval(evaluationContext));
+            }
+            userAnswers.put(prompt.getKey(), defaultValue);
         }
     }
 
     public static RemoteQuerySessionManager buildQuerySessionManager(CommCareSession session,
-                                                                     EvaluationContext sessionContext) {
+                                                                     EvaluationContext sessionContext)  throws XPathException {
         SessionDatum datum;
         try {
             datum = session.getNeededDatum();
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -1,20 +1,22 @@
 package org.commcare.session;
 
+import org.commcare.cases.util.StringUtils;
 import org.commcare.modern.util.Pair;
-import org.commcare.suite.model.DisplayUnit;
+import org.commcare.suite.model.QueryPrompt;
 import org.commcare.suite.model.RemoteQueryDatum;
 import org.commcare.suite.model.SessionDatum;
+import org.javarosa.core.model.SelectChoice;
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.instance.ExternalDataInstance;
 import org.javarosa.core.model.instance.TreeElement;
+import org.javarosa.core.model.utils.ItemSetUtils;
 import org.javarosa.core.util.OrderedHashtable;
-import org.javarosa.xpath.expr.FunctionUtils;
 import org.javarosa.xml.ElementParser;
 import org.javarosa.xml.TreeElementParser;
 import org.javarosa.xml.util.InvalidStructureException;
 import org.javarosa.xml.util.UnfullfilledRequirementsException;
+import org.javarosa.xpath.expr.FunctionUtils;
 import org.javarosa.xpath.expr.XPathExpression;
-import org.javarosa.xpath.expr.XPathFuncExpr;
 import org.kxml2.io.KXmlParser;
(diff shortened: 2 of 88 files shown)


shubham1g5 (commented):
Looks good to me. There are some minor changes this PR will require on Formplayer side as seen in this [commit](https://github.com/dimagi/formplayer/pull/799/commits/7011ed5ac9075fb07bd832dc3feac1ccd9067475). I am happy to PR them separately with just the changes needed to support the default value in case claim if needed. 

I think it would be nice for this to go through a cursory QA given the large amount of changes. 

The bigger challenge in getting this out is on infra side though as bumping the DB version would force fresh restores for all users. @ctsims  and @dannyroberts  were looking at this issue earlier and we should wait for a go ahead from one of them before we deploy this.

ctsims:
@orangejenny I 100% think we should QA this.

@shubham1g5 I will read through to review for model changes, but is there a particular one that you think will require a userdb rebuild? There are 100% changes that will require app rebuilds, but I did a cursory review the other day and didn't spot any userdb changes specifically.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -40,6 +45,16 @@ private RemoteQuerySessionManager(RemoteQueryDatum queryDatum,
                                       EvaluationContext evaluationContext) {
         this.queryDatum = queryDatum;
         this.evaluationContext = evaluationContext;
+        initUserAnswers();
+    }
+
+    private void initUserAnswers() {
+        OrderedHashtable<String, QueryPrompt> queryPrompts = queryDatum.getUserQueryPrompts();
+        for (Enumeration en = queryPrompts.keys(); en.hasMoreElements(); ) {
+            String promptId = (String)en.nextElement();
```
ctsims, line 54: I think this needs to get updated to make sure this pr is included:
https://github.com/dimagi/commcare-core/pull/959

which has the init behavior for the query screen that USH needs for performing de-dupe.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1343,177 +1347,16 @@ public String fillTemplateString(String template, TreeReference contextRef, Hash
...
-     * Identify the itemset in the backend model, and create a set of SelectChoice
-     * objects at the current question reference based on the data in the model.
-     *
-     * Will modify the itemset binding to contain the relevant choices
-     *
-     * @param itemset The binding for an itemset, where the choices will be populated
-     * @param curQRef A reference to the current question's element, which will be
-     *                used to determine the values to be chosen from.
-     */
-    public void populateDynamicChoices(ItemsetBinding itemset, TreeReference curQRef) {
```
ctsims, line 1356: @shubham1g5 Shout out again for refactoring this code out. Anything that drops the size of the FormDef object by 200 lines is a huge improvement to the readability of the code.

ctsims (approved):
Only one concrete thing I think needs to be updated. 

I didn't identify anything in this PR that would require a UserDB rebuild, by the way, so I think we can avoid a cold start. Would be very helpful for someone to confirm, though.

build.gradle
```diff
@@ -70,7 +70,7 @@ dependencies {
     compile 'joda-time:joda-time:2.9.4'
     compile 'com.carrotsearch:hppc:0.7.2'
     compile 'com.squareup.retrofit2:retrofit:2.3.0'
-    compile 'com.squareup.okhttp3:okhttp:4.4.1'
+    compile 'com.squareup.okhttp3:okhttp:3.12.12'
```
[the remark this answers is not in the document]
orangejenny: @ShivamPokhriyal

orangejenny:
Cool, [QA ticket](https://dimagi-dev.atlassian.net/browse/QA-2218) created. I just asked for a formplayer regression and verified that includes case claim. If there's anything additional QA should focus on, please comment on that ticket.

@shubham1g5 So https://github.com/dimagi/formplayer/pull/799, or at least that commit, should be deployed at the same time as this PR? That sounds fine.

I'm planning to chat with @dannyroberts about the rollout, yes. Even if it's only app rebuilds, we might want to consider deploying over two nights instead of all at once.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
orangejenny: Thanks for flagging, my master was a couple of commits behind HEAD. That PR is now included, along with a few more commits from @shubham1g5 and @stephherbers

ctsims: great, thanks. Reviewed the new commits and didn't see any UserDB changes in those either.

build.gradle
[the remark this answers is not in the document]
shubham1g5: This was intentionally done for mobile as newer okhttp versions only provide support for Android 5+ and `3.12.12` is a okhttp backport for apps supporting Android 4 or lesser.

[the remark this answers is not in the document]
ctsims: Ah, thanks for confirming. I imagine this should be fine for formplayer and if there are any issues they'd show up in a general QA pass.

shubham1g5:
> I didn't identify anything in this PR that would require a UserDB rebuild, by the way, so I think we can avoid a cold start. Would be very helpful for someone to confirm, though.

Went through commits and confirming this is the case indeed. This merge seem harmless in that scenario though we won't be able to push https://github.com/dimagi/formplayer/pull/787  out with this. 

> So dimagi/formplayer#799, or at least that commit, should be deployed at the same time as this PR? That sounds fine.

Yes, dimagi/formplayer#799 will need the corresponding HQ changes to be deployed first though while that commit can be deployed without the HQ changes.

shubham1g5:
@orangejenny We should also update the [version here](https://github.com/dimagi/commcare-core/blob/master/src/cli/java/org/commcare/util/engine/CommCareConfigEngine.java#L71) to `50` as part of this PR as this merge will upgrade Formplayer to all changes added till CommCare 2.50 mobile version.

orangejenny:
@shubham1g5 Thank you, just added fdc85703c to update the version.

ctsims:
> > I didn't identify anything in this PR that would require a UserDB rebuild, by the way, so I think we can avoid a cold start. Would be very helpful for someone to confirm, though.
> 
> Went through commits and confirming this is the case indeed. This merge seem harmless in that scenario though we won't be able to push [dimagi/formplayer#787](https://github.com/dimagi/formplayer/pull/787) out with this.
> 
> > So [dimagi/formplayer#799](https://github.com/dimagi/formplayer/pull/799), or at least that commit, should be deployed at the same time as this PR? That sounds fine.
> 
> Yes, [dimagi/formplayer#799](https://github.com/dimagi/formplayer/pull/799) will need the corresponding HQ changes to be deployed first though while that commit can be deployed without the HQ changes.

Thanks for the spot check, glad to hear I didn't miss anything, and definitely happy for 787 to go out with its own lifecycle.

I think it would still be helpful for someone to physically confirm that sandboxes don't break by creating some piece of state of each type (a synced user with incomplete forms) and ensuring that all of those pieces work after updating to the new head. 

Since we don't do migrations it might actually be a good idea to have a way to easily produce a pre-saved DB file based on a CCZ and based on a restore XML file, and keep a pre-made sandbox of the "Current" clean state in the test assets. That would give us a lot more confidence into whether or not an update breaks one build or the other. I can see if I can put that together quickly to use for this.

ctsims:
@orangejenny I made a pr in formplayer to test what I was describing https://github.com/dimagi/formplayer/pull/802 . 

I can run it against this as soon as https://github.com/dimagi/formplayer/pull/800 is updated to target this head in the cc-core library reference.

shubham1g5 (approved):

orangejenny:
> I think it would still be helpful for someone to physically confirm that sandboxes don't break by creating some piece of state of each type (a synced user with incomplete forms) and ensuring that all of those pieces work after updating to the new head.

@ctsims Is this no longer needed because of https://github.com/dimagi/formplayer/pull/802? Or should I still do this?

ctsims:
@orangejenny I think https://github.com/dimagi/formplayer/pull/802 takes care of this. I did confirm that the test caught sandboxes breaking in a few different ways.

The only thing I'm not sure about is incomplete forms, which I'd forgotten about until re-reading this. That's certainly lower risk (I'm not sure USH even uses them), but might be worth testing directly.

orangejenny:
@ctsims Cool, I can test incomplete forms.

orangejenny:
Confirming that I tested incomplete forms locally:
* Checked out formplayer `master` and commcare-core `formplayer`, built jar
* Opened a form and filled out some fields
* Checked out formplayer `formplayer_2.50` and commcare-core `jls/formplayer-merge-master`, built jar
* Went into incomplete forms, found form, opened it and saw previous answers, finished and submitted it.

orangejenny:
QA is complete.

PR #971, opened 2021-01-20
Skip to Default Search and Search Again

shubham1g5:
Adds 'default_search' to Query Datum and 'redo_last`parameter to an Action

### DIFF
--- a/src/main/java/org/commcare/suite/model/Action.java
+++ b/src/main/java/org/commcare/suite/model/Action.java
@@ -83,6 +83,10 @@ public class Action implements Externalizable {
         return isAutoLaunching;
     }
 
+    public boolean isRedoLast() {
+        return redoLast;
+    }
+
     public boolean hasActionBarIcon() {
         return !"".equals(iconReferenceForActionBar);
     }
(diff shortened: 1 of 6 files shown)


src/main/java/org/commcare/suite/model/Action.java
```diff
@@ -30,6 +30,7 @@
     private XPathExpression relevantExpr;
     private String iconReferenceForActionBar;
     private boolean isAutoLaunching;
+    private boolean redoLast;
```
orangejenny, line 33: Does this need a getter and setter? I'm getting this locally:
```
> Task :compileJava
warning: unknown enum constant When.MAYBE
  reason: class file for javax.annotation.meta.When not found
/Users/jschweers/Documents/formplayer/src/main/java/org/commcare/formplayer/beans/menus/EntityListResponse.java:299: error: invalid method reference
        return formatActionsWithFilter(session, Action::isRedoLast);
```

shubham1g5: Ahh weird, can you make sure you are on Java 8 ? It's compiling successfully for me. The error seems to happen because of the method reference syntax `Action::isRedoLast` but also weird that you are not getting the same for `Action::isAutoLaunching`.

orangejenny: I worked around it with a local modification for a couple of days and then the error just disappeared, so... 🤷  ? At least it's good now.

orangejenny (approved):

PR #972, opened 2021-01-21
refreshes itemset choices after answering each prompt

shubham1g5:
https://dimagi-dev.atlassian.net/browse/QA-2340

The issue was happening since we were not refreshing the itemset choices after setting a select prompt value but only after we have set all prompts values ([removed here](https://github.com/dimagi/formplayer/pull/803/commits/f5efd3383123738ac9e04348540aa75b3d08c59e)).

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -15,6 +15,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
 import java.net.URL;
+import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.Map;
 import java.util.Vector;
@@ -126,7 +127,8 @@ public class QueryScreen extends Screen {
     }
 
     public void answerPrompts(Hashtable<String, String> answers) {
-        for (String key : userInputDisplays.keySet()) {
+        for (Enumeration en = userInputDisplays.keys(); en.hasMoreElements(); ) {
+            String key = (String)en.nextElement();
             QueryPrompt queryPrompt = userInputDisplays.get(key);
             String answer = answers.get(key);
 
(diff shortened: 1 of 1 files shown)


src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -140,9 +142,15 @@ public void answerPrompts(Hashtable<String, String> answers) {
...
-                answer = selectChoices.get(choiceIndex).getValue();
+                if (choiceIndex < selectChoices.size()) {
+                    answer = selectChoices.get(choiceIndex).getValue();
+                } else {
+                    // answer is no longer a valid choice, so clear it out
+                    answer = "";
+                }
             }
             remoteQuerySessionManager.answerUserPrompt(key, answer);
+            refreshItemSetChoices();
```
orangejenny, line 153: Checking my understanding: `answer` is initially the index in the old choice list, then `getValue` converts it to a value, then `refreshItemSetChoices` updates the dropdown options and also takes care of making sure that if the value is now at a different index, the index gets updated?

orangejenny (approved):

src/cli/java/org/commcare/util/screen/QueryScreen.java
shubham1g5: @orangejenny That's correct. `refreshItemSetChoices` just updates the dropdown options based on the new `values` set and if the current value is not present in the updated dropdown options, it reset the current value for the prompt to empty. This is [the part](https://github.com/dimagi/formplayer/blob/caseClaim+jls/parent-child-case-search/src/main/java/org/commcare/formplayer/beans/menus/QueryResponseBean.java#L46-L49) that updates the index later when sending the answers to the web apps.

orangejenny: ah, I hadn't found that last part. Thank you for explaining.

orangejenny: I think either I'm misunderstanding or this isn't working quite right. This is with master checked out, and the "mood" dropdown varies depending on the "rating" dropdown. But changing rating is changing the value of mood - that shouldn't happen, right? Shouldn't it either keep the same value or get cleared?
UGOYa2fw3B

shubham1g5: yeah the mood value should get cleared. Will see if I can fix it.

PR #975, opened 2021-02-03
Auto Advance with auto launch in a single request

shubham1g5:
Supporting changes for https://github.com/dimagi/formplayer/pull/843

### DIFF
--- a/src/cli/java/org/commcare/util/screen/CompoundScreenHost.java
+++ b/src/cli/java/org/commcare/util/screen/CompoundScreenHost.java
@@ -31,8 +31,8 @@ public abstract class CompoundScreenHost extends Screen {
     }
 
     @Override
-    public final boolean handleInputAndUpdateSession(CommCareSession session, String input) throws CommCareSessionException {
-        if (getCurrentScreen().handleInputAndUpdateHost(input, this)) {
+    public final boolean handleInputAndUpdateSession(CommCareSession session, String input, boolean allowAutoLaunch) throws CommCareSessionException {
+        if (getCurrentScreen().handleInputAndUpdateHost(input, this, allowAutoLaunch)) {
             this.updateSession(session);
             return false;
         }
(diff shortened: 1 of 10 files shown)


src/cli/java/org/commcare/util/screen/CompoundScreenHost.java
```diff
@@ -31,8 +31,8 @@ public void prompt(PrintStream out) throws CommCareSessionException {
     }
 
     @Override
-    public final boolean handleInputAndUpdateSession(CommCareSession session, String input) throws CommCareSessionException {
-        if (getCurrentScreen().handleInputAndUpdateHost(input, this)) {
+    public final boolean handleInputAndUpdateSession(CommCareSession session, String input, boolean allowAutoLaunch) throws CommCareSessionException {
```
ctsims, line 34: Flags like this always cause me a bit of pause because it's not obvious when and how the autolaunch behavior should be triggered and propagated.

What are the circumstances in which one wouldn't propagate the autolaunch? Is the flag there primarily to hedge against the possibility that it might cause problems in some circumstance?

ctsims (approved):

src/cli/java/org/commcare/util/screen/CompoundScreenHost.java
orangejenny: > Is the flag there primarily to hedge against the possibility that it might cause problems in some circumstance?

Originally, yes. That flag exists because at the time of adding it, I wasn't confident enough in formplayer test coverage to be sure it wasn't going to break another workflow. I don't remember if there was actually a known workflow that broke without the flag.

shubham1g5: I took a paas at refining the `allowAutoLaunch` flag a bit and it seems like the only place it's tunred off now is while [rebuilding the session](https://github.com/dimagi/formplayer/blob/autoLaunchRevamp/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L79) which seems intended as we probably want to bypass the mark and rewind worklfow caused by case search while rebuilding session.

orangejenny:
Force pushed because I accidentally merged in master. No real changes.

PR #976, opened 2021-02-08
reuse xformparser method for loading form XML into form def

snopoke:
These two methods are essentially identical:

https://github.com/dimagi/commcare-core/blob/5184a5d9c4c76192b9d76f599823830c5f40ded0/src/main/java/org/javarosa/xform/parse/XFormParser.java#L2826
https://github.com/dimagi/commcare-core/blob/2f3530ad5df21f233c357ae07630514accb8fece/src/translate/java/org/javarosa/xform/schema/FormInstanceLoader.java#L46

Refactored the one used in less places (the utility in `translate` to call the one in `XFormParser`

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -2814,8 +2814,8 @@ public class XFormParser {
         }
     }
 
-    public void loadXmlInstance(FormDef f, Reader xmlReader) throws IOException {
-        loadXmlInstance(f, getXMLDocument(xmlReader));
+    public static FormDef loadXmlInstance(FormDef formDef, Reader xmlReader) throws IOException {
+        return loadXmlInstance(formDef, getXMLDocument(xmlReader));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2814,21 +2814,24 @@ private void checkDependencyCycles() {
         }
     }
 
-    public void loadXmlInstance(FormDef f, Reader xmlReader) throws IOException {
-        loadXmlInstance(f, getXMLDocument(xmlReader));
+    public static FormDef loadXmlInstance(FormDef formDef, Reader xmlReader) throws IOException {
```
snopoke, line 2817: This was calling a static method so no reason it shouldn't itself be static

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2839,14 +2842,9 @@ public static void loadXmlInstance(FormDef f, Document xmlInst) {
...
 
         // populated model to current form
-        f.getMainInstance().setRoot(templateRoot);
-
-        // if the new instance is inserted into the formdef before f.initialize() is called, this
-        // locale refresh is unnecessary
-        //   Localizer loc = f.getLocalizer();
-        //   if (loc != null) {
-        //       f.localeChanged(loc.getLocale(), loc);
-        //     }
```
snopoke, line 2849: This has been commented out since 2011 so I think it's safe to delete.

src/translate/java/org/javarosa/xform/schema/FormInstanceLoader.java
```diff
@@ -45,29 +45,12 @@ public static FormDef loadInstance(InputStream formInput,
...
-                formDef.getInstance().getRoot().deepCopy(true);
-
-        // weak check for matching forms
-        if (!savedRoot.getName().equals(templateRoot.getName()) ||
-                savedRoot.getMult() != 0) {
-            System.out.println("Instance and template form definition don't match");
+        FormInstance formInstance = XFormParser.restoreDataModel(instanceInput, null);
+        try {
+            return XFormParser.loadXmlInstance(formDef, formInstance);
+        } catch (RuntimeException e) {
```
shubham1g5, line 51: I think it would be ok to change behaviour here and not catch this exception here but let this bubble up to top level Formplayer classes just like the `XFormParser.loadXmlInstance`

shubham1g5 (dismissed):
Looks good, left one optional suggestion.

src/translate/java/org/javarosa/xform/schema/FormInstanceLoader.java
snopoke: Won't that make the cli crash?

shubham1g5: yeah though I think it's better to fail hard and crash rather than causing unseen issues. Like currently if there is an error in this part of the code, we will hardly get to know about them unless we are proactively looking out for the system log.

snopoke: 22fc0f38b

shubham1g5 (approved):

PR #977, opened 2021-02-18
Add receive attribute to QueryPrompt

sravfeyn:
Required to support Geocoder receive functionality. https://docs.google.com/document/d/1vzgtemtA0Zijj4g5zZQ413IVvFn4O6_m_w6pao4UM70/edit?ts=602d3494

Formplayer PR https://github.com/dimagi/formplayer/pull/877

@orangejenny

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -8,6 +8,7 @@ import org.commcare.suite.model.DisplayUnit;
 import org.commcare.suite.model.QueryPrompt;
 import org.javarosa.core.model.instance.ExternalDataInstance;
 import org.javarosa.core.util.OrderedHashtable;
+import org.javarosa.core.services.locale.Localization;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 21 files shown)


orangejenny (approved):

sravfeyn:
This is my first commcare-core PR, is there any other steps I should take before I can merge this? @ctsims

orangejenny:
One point is that this is a PR into master, but it'll need to go into the formplayer branch for web apps to have access.

ctsims:
@sravfeyn A couple things:

1) As Jenny mentioned, we deploy the `formplayer` branch to formplayer, so we generally make changes directly against that branch. This is also helpful for the other apps because it defers breaking changes into more discrete merges which can be planned for on the android side (IE: If an android change was necessary we can figure that out in the formplayer master merge rather than the build breaking outright). 
2) Generally it's good to post a PR like this along with the Formplayer PR that it goes with (there's a label for 'cross requesting'), since changes the Formplayer PR could bubble back down to change the core interface again. As a reviewer I generally like to review both PR's in batches to understand where this one's going 
3) The formplayer PR should be tagged with "breaks appdb" since this changes the serialization code for the business logic model.

sravfeyn:
Point noted, thanks Clayton.

sravfeyn:
Closing this. New PR https://github.com/dimagi/commcare-core/compare/formplayer...rec?expand=1

PR #980, opened 2021-03-08
make LocaleFileInstaller.cache protected

snopoke:
Note this is one commit on top of the current formplayer submodule reference


ctsims (changes requested):
I think this should be targeting the `formplayer` head, I think

snopoke:
> I think this should be targeting the `formplayer` head, I think

yes, I posted on slack about this - I'll rebase and push again

snopoke:
not overriding this class anylonger

PR #985, opened 2021-03-31
Break out of purge in case of cycles

shubham1g5:
https://github.com/dimagi/formplayer/pull/948

### DIFF
--- a/src/cli/java/org/commcare/util/cli/ApplicationHost.java
+++ b/src/cli/java/org/commcare/util/cli/ApplicationHost.java
@@ -1,6 +1,7 @@
 package org.commcare.util.cli;
 
 import org.commcare.cases.util.CasePurgeFilter;
+import org.commcare.cases.util.InvalidCaseGraphException;
 import org.commcare.core.interfaces.UserSandbox;
 import org.commcare.core.parse.CommCareTransactionParserFactory;
 import org.commcare.core.parse.ParseUtils;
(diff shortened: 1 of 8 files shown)


src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -258,6 +259,9 @@ private boolean loopSession() throws IOException {
 
                     //Restart
                     return true;
+                } catch (InvalidCaseGraphException e) {
```
ctsims, line 262: This doesn't quite seem like the right place to catch this to me. The other two exceptions are broad-context exceptions that could get thrown in any number of circumstances, but this one is highly specific to one or two states of execution

ctsims (changes requested):
One small point of feedback on how this is caught in the CLI.

It definitely would be useful to be able to specifically output at least one case ID for a node in the cycle if possible to help identify the problem and prevent users from being caught in cycles where they are regularly 412ing.

src/cli/java/org/commcare/util/cli/ApplicationHost.java
ctsims: I think it's fine to start with just "Catch and Rethrow as a Runtime Exception" for now, or to simply print the error message and move on (as is done right now inside the `performCasePurge` method for the Error/Warning cases)

ctsims (approved):

PR #986, opened 2021-04-01
Detect DAG cycles in case graphs explicitly

ctsims:
Fixes the detection portion of https://github.com/dimagi/commcare-core/pull/985. I was incorrect about the initial hunch on guessing the maximum size of a graph walk, since the propagation algorithm can end up walking nodes redundantly (apologies, @shubham1g5 , bad advice on my part).

This replaces the implicit detection with a one-time explicit DFS dive into the graph to find cycles before beginning the purge process. Since we're already doing as (or less) efficient walks as part of the purge, this shouldn't meaningfully affect performance.

https://github.com/dimagi/formplayer/pull/948 should be updated against this once ready.


shubham1g5 (approved):
This is great. thanks for adding this.

PR #987, opened 2021-04-19
add method to check for presence of case id

calellowitz:
Follow up for PR feedback on https://github.com/dimagi/formplayer/pull/898 to move the bulk of the logic to the `EntityScreen`

This mostly just copies the logic that already existed, but does check if the `referenceMap` exists, which allows constant time lookup of the answer, but is not generally instantiated during this check (since it involves resolving every item in the entity list).

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityScreen.java
@@ -109,8 +109,9 @@ public class EntityScreen extends CompoundScreenHost {
 
         if (full || references.size() == 1) {
             referenceMap = new Hashtable<>();
+            EntityDatum needed = (EntityDatum) session.getNeededDatum();
             for(TreeReference reference: references) {
-                referenceMap.put(getReturnValueFromSelection(reference, (EntityDatum) session.getNeededDatum(), evalContext), reference);
+                referenceMap.put(getReturnValueFromSelection(reference, needed, evalContext), reference);
             }
 
             // for now override 'here()' with the coords of Sao Paulo, eventually allow dynamic setting
@@ -324,8 +325,9 @@ public class EntityScreen extends CompoundScreenHost {
         if (referenceMap != null) {
             return referenceMap.containsKey(stepValue);
         }
+        EntityDatum needed = (EntityDatum) session.getNeededDatum();
         for (TreeReference ref: references) {
-            String id = getReturnValueFromSelection(ref,(EntityDatum) session.getNeededDatum(), evalContext);
+            String id = getReturnValueFromSelection(ref, needed, evalContext);
             if (id.equals(stepValue)) {
                 return true;
             }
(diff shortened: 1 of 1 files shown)


src/cli/java/org/commcare/util/screen/EntityScreen.java
```diff
@@ -318,4 +318,18 @@ public String toString() {
...
         return referenceMap;
     }
+
+    @SupressWarnings("unsused")
+    public boolean referencesContainStep(String stepValue) {
+        if (referenceMap != null) {
+            return referenceMap.containsKey(stepValue);
+        }
+        for (TreeReference ref: references) {
+            String id = getReturnValueFromSelection(ref,(EntityDatum) session.getNeededDatum(), evalContext);
```
snopoke, line 328: should `session.getNeededDatum()` be pulled out of the loop?

snopoke (approved):
Small comment but not blocking

src/cli/java/org/commcare/util/screen/EntityScreen.java
ctsims: Definitely. There's a lot of potential processing in that step.

src/cli/java/org/commcare/util/screen/EntityScreen.java
```diff
@@ -318,4 +318,18 @@ public String toString() {
     public Hashtable<String, TreeReference> getReferenceMap() {
         return referenceMap;
     }
+
+    @SupressWarnings("unsused")
+    public boolean referencesContainStep(String stepValue) {
```
ctsims, line 323: This is going to be a fairly expensive call, by the way, since it re-processes all of the datum outputs. Is this being done for performance reasons?

src/cli/java/org/commcare/util/screen/EntityScreen.java
calellowitz: Interesting, I did it this way because that is how it is done when creating the `referenceMap` above. Sounds like we should change that one too?

calellowitz: https://github.com/dimagi/commcare-core/blob/formplayer/src/cli/java/org/commcare/util/screen/EntityScreen.java#L113

calellowitz: Followup question, is there a reason not to use the `mNeededDatum` attribute in both places, rather than fetch it from the session again.

ctsims: Reusing mNeededDatum would require a bit of a deeper look than I'm familiar with currently.

My strong sense is that this wasn't pulled out in the other loop simply because originally this was toy code for the CLI that wasn't written with performance in mind. I agree that it would be good to pull it out of the loop above as well

snopoke (approved):

snopoke:
@calellowitz @ctsims is this GTG?

ctsims (approved):

PR #988, opened 2021-04-27
Fix backward navigation from child case registration

ShivamPokhriyal:
Jira https://dimagi-dev.atlassian.net/browse/SAAS-12043

In a parent-child case management app, when you open a child registration form and press back(without completing the form) then there's a bug in commcare where it opens parent case list rather than child case list. This PR will fix the backward navigation so that pressing back from child case registration will return the user to the child case list.

### DIFF
--- a/src/main/java/org/commcare/session/CommCareSession.java
+++ b/src/main/java/org/commcare/session/CommCareSession.java
@@ -374,6 +374,15 @@ public class CommCareSession {
         String neededData = getNeededData(evalContext);
         String poppedType = popped == null ? "" : popped.getType();
 
+        if (popped != null) {
+            Boolean wasLastStepCompound = (Boolean) popped.getExtra(SessionFrame.COMPOUND_STEP);
+            Boolean isCurrentStepCompound = (Boolean) frame.getTopStepExtra(SessionFrame.COMPOUND_STEP);
+            Boolean yes = true;
+            if (yes.equals(wasLastStepCompound) && yes.equals(isCurrentStepCompound)) {
+                return true;
+            }
+        }
+
         if (neededData == null ||
                 SessionFrame.STATE_DATUM_COMPUTED.equals(neededData) ||
                 SessionFrame.STATE_DATUM_COMPUTED.equals(poppedType) ||
@@ -669,7 +678,7 @@ public class CommCareSession {
     private void createFrame(SessionFrame createdFrame,
                              StackOperation op, EvaluationContext ec) {
         if (op.isOperationTriggered(ec)) {
-            performPushInner(op, createdFrame, ec);
+            performPushInner(op, createdFrame, ec, false);
             pushNewFrame(createdFrame);
         }
     }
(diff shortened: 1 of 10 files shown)


src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -374,6 +374,15 @@ private boolean shouldPopNext(EvaluationContext evalContext) {
         String neededData = getNeededData(evalContext);
         String poppedType = popped == null ? "" : popped.getType();
 
+        if (popped != null) {
+            Boolean wasLastStepCompound = (Boolean) popped.getExtra(SessionFrame.COMPOUND_STEP);
+            Boolean isCurrentStepCompound = (Boolean) frame.getTopStepExtra(SessionFrame.COMPOUND_STEP);
+            Boolean yes = true;
+            if (yes.equals(wasLastStepCompound) && yes.equals(isCurrentStepCompound)) {
```
shubham1g5, line 381:
```suggestion
            if (wasLastStepCompound && isCurrentStepCompound) {
```

src/main/java/org/commcare/session/SessionFrame.java
```diff
@@ -75,6 +75,8 @@
      */
     public static final String STATE_FORM_XMLNS = "FORM_XMLNS";
 
+    public static final String COMPOUND_STEP = "COMPOUND_STEP";
```
shubham1g5, line 78: can we add a comment here in terms of what does a compound step denotes.

shubham1g5 (changes requested):
looks like a nice approach to me though it's hard to know if it will cause any regressions, Have you been able to run the integration test suite against these changes ?

@ctsims Would be good to have your eyes on this as well.

src/main/java/org/commcare/session/CommCareSession.java
ShivamPokhriyal: Just realised why I had to use `yes.equals`. Will remove the last commit. 
`wasLastStepCompound ` and `isCurrentStepCompound` can be null so if I do `if (null  && null)` it'll cause NPE.

shubham1g5: there is not much value in defaulting a boolean to `null` , we should default it to `false` if possible.

ShivamPokhriyal: It come from here: 
https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/suite/model/StackFrameStep.java#L123

and to get a default `false` we'll have to modify that to something like
```
if (SessionFrame.COMPOUND_STEP.equals(key)) {
  Boolean obj = (Boolean) extras.get(key);
  return obj == null ? false : obj
} else {
  return extras.get(key);
}
```

or we can have add another method there: 
```
public boolean isCompoundStep() {
  Boolean obj = (Boolean) extras.get(SessionFrame.COMPOUND_STEP);
  return obj == null ? false : obj
}
```
I like the second option better

ShivamPokhriyal:
@ctsims bumping for review

ctsims:
Hi @ShivamPokhriyal,

So the code itself here looks appropriate to me, but I'm not quite sure if I'm signed off on the actual scope of the change. 

As I mentioned in the ticket, there's a big limitation to this approach

> I don’t think we always want to treat multiple stack steps as compound. When we do end of form nav, for instance, we do want the user’s “Back” presses to move them back one step in the process. I think we should think of this bad config as being the result of either the user’s computed back state being invalid, or the result of a lack of expressiveness in the suite file to give app builders a way to say “These four steps should really be considered one step”

As far as I can tell, after this change **all** `<stack>` operations will basically lose their "back" information, which is a regression for most apps. If the app end-of-form navigates me to the page I was on before entering a form, pressing back and being taken to the home screen is a worse user experience than I have today.

I think it's fine to introduce the concept of a compound step which the app builder can inject into the `<stack>` ops they generate in order to work around problems like this, but if we take that approach we should also be able to communicate over the circumstances under which they should add a compound tag.

shubham1g5:
Closing it since this doesn't seem to be right solution for the issue.

PR #990, opened 2021-04-28
Get case name using 'case-name-descriptor' itext entry

ShivamPokhriyal:
Jira https://dimagi-dev.atlassian.net/browse/SAAS-11500

Uses a special itext `case-name-descriptor` to grab case name from the form. 
cross-request: https://github.com/dimagi/commcare-android/pull/2462

### DIFF
--- a/src/main/java/org/commcare/util/FormMetaIndicatorUtil.java
+++ b/src/main/java/org/commcare/util/FormMetaIndicatorUtil.java
@@ -10,7 +10,7 @@ import io.reactivex.annotations.Nullable;
  */
 public class FormMetaIndicatorUtil {
 
-    public static final String CASE_NAME_DESCRIPTOR = "Pragma-Case-Name";
+    public static final String FORM_DESCRIPTOR = "Pragma-Form-Descriptor";
 
     @Nullable
     public static String getPragma(String key, FormDef formDef, TreeReference contextRef) {
--- a/src/main/java/org/commcare/util/FormMetaIndicatorUtil.java
+++ b/src/main/java/org/commcare/util/FormMetaIndicatorUtil.java
@@ -1,0 +1,24 @@
+package org.commcare.util;
+
+import org.javarosa.core.model.FormDef;
+import org.javarosa.core.model.instance.TreeReference;
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/javarosa/form/api/FormController.java
```diff
@@ -121,6 +122,17 @@ public void postProcessInstance() {
         mFormEntryController.getModel().getForm().postProcessInstance();
     }
 
+    @Nullable
+    public String getFormDescriptor() {
+        FormDef formDef = mFormEntryController.getModel().getForm();
+        String caseNameDescriptor = formDef.getLocalizer().getText("case-name-descriptor");
```
ShivamPokhriyal, line 128: We should probably move this string out to a constant. But I'm not sure where to put it. Is there a place where we keep these special itexts?

ctsims (dismissed):
This looks good, but I agree it would be worth thinking through some of the details about what to name the key and where to put it.

Can you also find a place to document this behavior? On the wiki is probably good.

shubham1g5 (dismissed):

src/main/java/org/commcare/util/FormMetaIndicatorUtil.java
```diff
@@ -0,0 +1,24 @@
...
+import org.javarosa.core.model.instance.TreeReference;
+
+import io.reactivex.annotations.Nullable;
+
+/**
+ * @author $|-|!˅@M
+ */
+public class FormMetaIndicatorUtil {
+
+    public static final String CASE_NAME_DESCRIPTOR = "Pragma-Case-Name";
```
ctsims, line 13: I think this should be titled around the output of the descriptor, not the most common input. The summary descriptor here can reference any data from within the form or without, so it might cause confusion to imply that its function is specific to cases.

ctsims (changes requested):

ctsims (approved):

shubham1g5:

shubham1g5:

shubham1g5:

shubham1g5:

shubham1g5 (approved):

PR #992, opened 2021-04-28
Revert "Revert "add method to check for presence of case id""

snopoke:
Reverts dimagi/commcare-core#991
Redo of https://github.com/dimagi/commcare-core/pull/987

Follow up for PR feedback on dimagi/formplayer#898 to move the bulk of the logic to the EntityScreen

This mostly just copies the logic that already existed, but does check if the referenceMap exists, which allows constant time lookup of the answer, but is not generally instantiated during this check (since it involves resolving every item in the entity list).


snopoke:
@calellowitz @ctsims I've added 2 commits that fix the build

calellowitz (approved):

PR #995, opened 2021-05-20
Adds Json utility class

shubham1g5:
cross-request: https://github.com/dimagi/formplayer/pull/964

### DIFF
--- a/src/main/java/org/commcare/util/JsonUtils.java
+++ b/src/main/java/org/commcare/util/JsonUtils.java
@@ -1,0 +1,24 @@
+package org.commcare.util;
+
+import org.json.JSONArray;
+
+import javax.annotation.Nullable;
+
+public class JsonUtils {
+
+    /**
+     * Converts a JSON Array to a String Array
+     * @param jsonArray A JSON Array containing string objects that we wish to convert to a String Array
+     * @return A String Array representation for the given jsonArray
+     */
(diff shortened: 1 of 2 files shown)


snopoke (approved):

src/main/java/org/commcare/util/JsonUtils.java
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Converts a JSON Array to a String Array
+     * @param jsonArray A JSON Array containing string objects that we wish to convert to a String Array
+     * @return A String Array representation for the given jsonArray
+     */
+    public static String[] toArray(@Nullable JSONArray jsonArray) {
+        if (jsonArray != null) {
+            String[] stringArray = new String[jsonArray.length()];
+            for (int i = 0; i < jsonArray.length(); i++) {
+                stringArray[i] = jsonArray.optString(i);
```
snopoke, line 18: if the item in the list isn't a String this will add a `null` to the output list.

shubham1g5:
````
 public String optString(int index, String defaultValue) {
        Object object = this.opt(index);
        return JSONObject.NULL.equals(object) ? defaultValue : object
                .toString();
    }
````

where `defaultValue` is empty string. So It should be empty if no such key is there otherwise the object would be converted to string.

PR #996, opened 2021-05-24
Add hidden attr to case search screen

sravfeyn:
@shubham1g5 

cross-request: https://github.com/dimagi/formplayer/pull/969


shubham1g5 (approved):
Would be good to add `hidden` to the prompt [here](https://github.com/dimagi/commcare-core/blob/master/src/test/resources/case_claim_example/suite.xml#L67) so that the code can run through the parsing logic for `hidden` as part of the test.

sravfeyn:
QA is finished.

PR #998, opened 2021-05-28
Support for multi-select fields in case search

shubham1g5:
cross-request: https://github.com/dimagi/formplayer/pull/973

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -143,14 +156,13 @@ public void answerPrompts(Hashtable<String, String> answers) {
...
-                if (choiceIndex < selectChoices.size()) {
-                    answer = selectChoices.get(choiceIndex).getValue();
-                } else {
-                    // answer is no longer a valid choice, so clear it out
-                    answer = "";
+            if (queryPrompt.isSelect() && !answer.isEmpty()) {
+                String[] selectedChoices = RemoteQuerySessionManager.extractSelectChoices(answer);
+                for (String selectedChoice : selectedChoices) {
+                    int choiceIndex = Integer.parseInt(selectedChoice);
+                    Vector<SelectChoice> selectChoices = queryPrompt.getItemsetBinding().getChoices();
```
snopoke, line 163: could move this out the loop

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -171,14 +169,16 @@ public void refreshItemSetChoices(Hashtable<String, String> userAnswers) {
...
-                    Vector<SelectChoice> items = queryPrompt.getItemsetBinding().getChoices();
-                    if (!checkForValidSelectValue(items, answer)) {
-                        // if it's not a valid select value, blank it out
-                        userAnswers.put(promptId, "");
-                        dirty = true;
+                    String[] selectedChoices = extractSelectChoices(answer);
+                    for (String selectedChoice : selectedChoices) {
+                        if (!checkForValidSelectValue(queryPrompt.getItemsetBinding(), selectedChoice)) {
+                            // if it's not a valid select value, blank it out
+                            userAnswers.put(promptId, answer.replace(selectedChoice, ""));
```
snopoke, line 179: will this handle the delimiters in `answer` correctly? Won't it just replace the value: `a#,#b#,#c` -> `a#,##,#,c`?

Could be worth a test.

src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -143,15 +156,17 @@ public void answerPrompts(Hashtable<String, String> answers) {
...
-                    answer = selectChoices.get(choiceIndex).getValue();
-                } else {
-                    // answer is no longer a valid choice, so clear it out
-                    answer = "";
+                StringBuilder answerWithSelectValues = new StringBuilder();
+                for (int i = 0; i < selectedChoices.length; i++) {
+                    int choiceIndex = Integer.parseInt(selectedChoices[i]);
+                    boolean validChoice = choiceIndex < selectChoices.size() && choiceIndex > -1;
+                    answerWithSelectValues.append(validChoice ? selectChoices.get(choiceIndex).getValue() : "");
+                    answerWithSelectValues.append(i == selectedChoices.length - 1 ? "" : RemoteQuerySessionManager.MULTI_SELECT_DELIMITER);
```
snopoke, line 167: similar comment to on https://github.com/dimagi/formplayer/pull/973/: might be cleaner to put the indices in an array or list and then use `String.join` with the delimiter to get the final result.

shubham1g5: addressed.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
shubham1g5: great catch! Corrected the code here and added a test on Formplayer PR

src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -143,15 +157,17 @@ public void answerPrompts(Hashtable<String, String> answers) {
...
-                int choiceIndex = Integer.parseInt(answer);
+            if (queryPrompt.isSelect() && !answer.isEmpty()) {
                 Vector<SelectChoice> selectChoices = queryPrompt.getItemsetBinding().getChoices();
-                if (choiceIndex < selectChoices.size()) {
-                    answer = selectChoices.get(choiceIndex).getValue();
-                } else {
-                    // answer is no longer a valid choice, so clear it out
-                    answer = "";
+                String[] indicesOfSelectedChoices = RemoteQuerySessionManager.extractSelectChoices(answer);
+                ArrayList<String> selectedChoices = new ArrayList<>(indicesOfSelectedChoices.length);
```
snopoke, line 163: is it necessary to specify the length here? Just wondering what happens if a choice is invalid, will that result in the final answer having `nulls` in it?

shubham1g5: it's not the length but initial capacity for the ArrayList. It defaults to 10 if we don't supply any capacity argument here but generally a good practice to supply it explicitly if we have a good estimate to avoid any resizing as the list grows.

PR #999, opened 2021-06-09
Rotate large labels on x-axis in graph

ShivamPokhriyal:
Rotates the graph label on x-axis when the label size is too large. 
cross-request: https://github.com/dimagi/commcare-android/pull/2494

### DIFF
--- a/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
+++ b/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
@@ -5,6 +5,7 @@ import org.commcare.core.graph.model.SeriesData;
 import org.commcare.core.graph.util.GraphException;
 import org.commcare.core.graph.util.GraphUtil;
 
+import org.commcare.core.graph.util.StringWidthUtil;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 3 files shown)


src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
```diff
@@ -118,10 +121,21 @@ private void addTickConfig(JSONObject axis, String key, String varName) throws G
                         } else {
                             values.put(parseDouble(location, key));
                         }
+                        try {
+                            String current = labels.getString(location);
+                            if (current.length() > largestLabel.length()) {
+                                largestLabel = current;
+                            }
+                        } catch (JSONException e) { }
```
shubham1g5, line 129: should we throw an error here ?

shubham1g5, line 129: should we throw an error here ?

shubham1g5 (approved):

src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
ShivamPokhriyal: Nope, the graphs work perfectly fine as of now. Just that the labels can overlap. Here, I'm trying to get the largest label length, and if `getString` throws an exception, then it doesn't matter, we simply won't rotate the label.

shubham1g5: makes sense, thanks

PR #1002, opened 2021-06-14
Supports Case Search for Multi-Valued Keys

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1103
cross-request: https://github.com/dimagi/formplayer/pull/976


snopoke (approved):

src/main/java/org/javarosa/core/util/externalizable/ExtWrapArrayListMultiMap.java
```diff
@@ -0,0 +1,93 @@
...
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.List;
+
+public class ExtWrapArrayListMultiMap extends ExternalizableWrapper {
```
ctsims, line 11: For simplicity I'd probably just call this class `ExtWrapMultiMap`  and see the below

src/main/java/org/javarosa/core/util/externalizable/ExtWrapArrayListMultiMap.java
```diff
@@ -0,0 +1,93 @@
...
+            for (long l = 0; l < numberOfValues; l++) {
+                multimap.put(key, ExtUtil.read(in, new ExtWrapTagged(), pf));
+            }
+        }
+        val = multimap;
+    }
+
+    @Override
+    public void writeExternal(DataOutputStream out) throws IOException {
+        ArrayListMultimap multimap = (ArrayListMultimap)val;
```
ctsims, line 70: It seems a bit unnecessary to me that this whole wrapper be dependent on the implementation of the multimap, and potentially very dangerous for the future if code signatures which use just the multimap signature change their underlying implementation.

I'd suggest doing a type cast here to an arraylist multimap by checking whether that's the implementation first and casting if so, but converting the object type if not, rather than crashing.

src/main/java/org/javarosa/core/util/externalizable/ExtWrapArrayListMultiMap.java
```diff
@@ -0,0 +1,93 @@
...
+        if (keyType == null) {
+            throw new NullPointerException();
+        }
+
+        this.keyType = keyType;
+    }
+
+    @Override
+    public ExternalizableWrapper clone(Object val) {
+        return new ExtWrapArrayListMultiMap((Multimap)val, keyType);
```
ctsims, line 50: I'm probably misremembering, but should this method call deep-copy the `val` object? (Easiest to just check other implementations)

shubham1g5: Thanks for the prompt. I checked other implementations ([here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/util/externalizable/ExtWrapMap.java#L76) and [here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/util/externalizable/ExtWrapList.java#L65)) and it seems to be a shallow copy.

shubham1g5:
@ctsims Made some changes to generalize the Ext wrapper for Multimap. 

I checked the size impact of Guava on Android apk and that seems minimal (few Kbs) , most probably proguagrd is stripping off most of the unused code.

src/main/java/org/javarosa/core/util/externalizable/ExtWrapMultiMap.java
```diff
@@ -0,0 +1,120 @@
...
+    public ExternalizableWrapper clone(Object val) {
+        return new ExtWrapMultiMap((Multimap)val, keyType);
+    }
+
+    @Override
+    public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
+        long size = ExtUtil.readNumeric(in);
+        try {
+            Multimap<Object, Object> multimap;
+            if (mapImplementation.equals(ArrayListMultimap.class)) {
```
ctsims, line 68: Sorry, to eliminate confusion: I don't think you need to retain the multimap implementation, here. The output of the ExtWrapMultiMap class can always be an ArrayListMultimap, and I think the added complexity for reflecting it here isn't worth it (and adds dangerous assumptions like the existence of a create method).

My only concern in the previous review was that in the *write* method, this wrapper was doing a blind cast to ArrayListMultimap, which seemed dangerous. It looks like you were able to change that by just relying on the generic type methods, though.

ctsims (changes requested):
Hey, one small change and I think it looks set. Thanks for digging into the feedback.

Glad to hear that proguard is eliminating the extra inputs. I had a hunch that one of the roles Guava plays sometimes has to do with serialization and reflection, so I wasn't sure if that would prevent stuff from being obfuscated out. It'll be great to have the new utility libraries to rely on moving forward!

src/main/java/org/javarosa/core/util/externalizable/ExtWrapMultiMap.java
shubham1g5: Thanks, Did the revert. I mostly was trying to copy the existing pattern from [ExtWrapList](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/util/externalizable/ExtWrapList.java#L74), but your concerns does sound valid.

ctsims (approved):

PR #1003, opened 2021-06-17
Adds Encryption utils for AES

shubham1g5:
Refactors AES encryption methods inside a utility class. 

cross-request: https://github.com/dimagi/commcare-android/pull/2502

### DIFF
--- a/src/main/java/org/commcare/util/EncryptionUtils.java
+++ b/src/main/java/org/commcare/util/EncryptionUtils.java
@@ -18,11 +18,11 @@ import javax.crypto.spec.SecretKeySpec;
 public class EncryptionUtils {
 
     /**
-     * Encrypts a message using the AES encryption
+     * Encrypts a message using the AES encryption and produces a base64 encoded payload containing the ciphertext, and a random IV which was used to encrypt the input.
      *
-     * @param message a message to be encrypted
-     * @param key     the key used for encryption
-     * @return AES encrypted message using the given key
+     * @param message a UTF-8 encoded message to be encrypted 
+     * @param key     A base64 encoded 256 bit symmetric key
+     * @return A base64 encoded payload containing the IV and AES encrypted ciphertext, which can be decoded by this utility's decrypt method and the same symmetric key
      */
     public static String encrypt(String message, String key) throws EncryptionException {
         final String ENCRYPT_ALGO = "AES/GCM/NoPadding";
@@ -69,7 +69,7 @@ public class EncryptionUtils {
     }
 
     /**
-     * Decrypts a message encoded with the AES encryption
+     * Decrypts a base64 payload containing an IV and AES encrypted ciphertext using the provided key
      *
      * @param message a message to be decrypted
      * @param key     key that should be used for decryption
--- a/src/main/java/org/commcare/util/EncryptionUtils.java
+++ b/src/main/java/org/commcare/util/EncryptionUtils.java
@@ -1,0 +1,114 @@
+package org.commcare.util;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/util/EncryptionUtils.java
```diff
@@ -0,0 +1,114 @@
...
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+public class EncryptionUtils {
+
+    /**
+     * Encrypts a message using the AES encryption
```
ctsims, line 21:
```suggestion
     * Encrypts a message using the AES encryption and produces a base64 encoded payload containing the ciphertext, and a random IV which was used to encrypt the input.
```

src/main/java/org/commcare/util/EncryptionUtils.java
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @param message a message to be encrypted
```
ctsims, line 23:
```suggestion
     * @param message a UTF-8 encoded message to be encrypted 
```

src/main/java/org/commcare/util/EncryptionUtils.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param key     the key used for encryption
```
ctsims, line 24:
```suggestion
     * @param key     A base64 encoded 256 bit symmetric key
```

src/main/java/org/commcare/util/EncryptionUtils.java
```diff
@@ -0,0 +1,114 @@
...
+     * @return AES encrypted message using the given key
```
ctsims, line 25: A base64 encoded payload containing the IV and AES encrypted ciphertext, which can be decoded by this utility's decrypt method and the same symmetric key

src/main/java/org/commcare/util/EncryptionUtils.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+        if (8 * keyBytes.length != KEY_LENGTH_BIT) {
+            throw new EncryptionException("Key should be " + KEY_LENGTH_BIT +
+                    " bits long, not " + 8 * keyBytes.length);
+        }
+        return new SecretKeySpec(keyBytes, "AES");
+    }
+
+    /**
+     * Decrypts a message encoded with the AES encryption
```
ctsims, line 72:
```suggestion
     * Decrypts a base64 payload containing an IV and AES encrypted ciphertext using the provided key
```

ctsims (changes requested):
These kinds of methods are really important to document clearly, since the details of how something is encrypted matter quite a bit. Since this is now something that is going to be a more "library" kind of method I think it's a really good opportunity to document a bit more clearly.

I made some initial suggestions on what should be more clear, but it might make more sense to document the class javadoc instead outlining the details, and then leaving the method specific details less verbose. 

One big remaining point of documentation, though: When working with these kinds of encryption methods which are intended to consume inputs that come from another system or provide outputs which could be read by another system (a base64 encoded payload with the IV and ciphertext), it's really, really helpful to be more specific than you might think you need to be about those elements.

The user story I'm thinking of is someone who needs to generate the symmetric key that will be used here. Someone will need to create that key, and it will probably be the case that they won't be using this Java library to do so, so they'll need to know exactly what to do in a totally separate system. There are [multiple base64](https://en.wikipedia.org/wiki/Base64#Implementations_and_history) encoding schemes, for instance, and which one is the "Default" is very different between platforms.

I'd recommend both being specific about any of the relevant details, but also potentially including a code snippet (in the commcare-core wiki docs, probably) in python which will generate a compatible key.

ctsims:
@shubham1g5 Let me know if this is time sensitive, btw. I do think we should make the changes I mentioned, but I don't want to block work on them if we need to get this in for the release

src/main/java/org/commcare/util/EncryptionUtils.java
shubham1g5, line 25:
```suggestion
     * @return A base64 encoded payload containing the IV and AES encrypted ciphertext, which can be decoded by this utility's decrypt method and the same symmetric key
```

ctsims (approved):
Approving for now based on our conversation, we can plan out key docs after initital testing

PR #1004, opened 2021-06-23
Formplayer to Mobile Merge

shubham1g5:
- Formplayer Merge 
- Adds a change to disrupt case index on new index creation so that the case purge process can be called on new indexes getting created and discover any invalid case relationship while saving a form

cross-request: https://github.com/dimagi/commcare-android/pull/2504

### DIFF
--- a/src/cli/java/org/commcare/util/cli/ApplicationHost.java
+++ b/src/cli/java/org/commcare/util/cli/ApplicationHost.java
@@ -488,6 +488,7 @@ public class ApplicationHost {
                     SandboxUtils.extractEntityOwners(sandbox));
         } catch (InvalidCaseGraphException e) {
             printStream.println(e.getMessage());
+            return;
         }
 
         int removedCases = sandbox.getCaseStorage().removeAll(purger).size();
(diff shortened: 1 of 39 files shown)


src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -481,8 +482,13 @@ private void syncAndReport() {
...
     public void performCasePurge(UserSandbox sandbox) {
         printStream.println("Performing Case Purge");
-        CasePurgeFilter purger = new CasePurgeFilter(sandbox.getCaseStorage(),
-                SandboxUtils.extractEntityOwners(sandbox));
+        CasePurgeFilter purger = null;
+        try {
+            purger = new CasePurgeFilter(sandbox.getCaseStorage(),
+                    SandboxUtils.extractEntityOwners(sandbox));
+        } catch (InvalidCaseGraphException e) {
+            printStream.println(e.getMessage());
```
ctsims, line 490: Small thing, but I think it would be good to bail out entirely here (just `return`), since the purger will come back from this null and might cause problems downtream.

ctsims (dismissed):
Didn't see any indications of a bad merge. Caught one late very small thing but it's probably just a mistake from the original code, so no need to hold off on merging if needed.

src/cli/java/org/commcare/util/cli/ApplicationHost.java
shubham1g5: Thanks for pointing. Created a new [PR](https://github.com/dimagi/commcare-core/pull/1005) to `formplayer` since that seem mostly relevant to web apps.

ctsims (approved):

PR #1007, opened 2021-06-28
Use configuration from HQ for rotating x-label

ShivamPokhriyal:
Responding to https://forum.dimagi.com/t/commcare-2-52-0-pre-release-available/8279/2 user feedback from forum. 
This PR adds a config `x-label-rotation` to set a rotation angel from HQ. If none is set, we'll default to using 75 whenever label is large. 

Need to update the graph wiki post merge/release.

### DIFF
--- a/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
+++ b/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
@@ -132,9 +132,16 @@ public class AxisConfiguration extends Configuration {
                     mVariables.put(varName, labels.toString());
                     usingCustomText = true;
                     // These custom labels can be large, rotating them ensures that they're shown correctly on X axis.
-                    height = StringWidthUtil.getStringWidth(largestLabel);
-                    if (isX && largestLabel.length() > 5 && height != -1) {
-                        tick.put("rotate", 75);
+                    if (isX) {
+                        height = StringWidthUtil.getStringWidth(largestLabel);
+                        if (largestLabel.length() > 5 && height != -1) {
+                            tick.put("rotate", 75);
+                        }
+                        // Use configuration from HQ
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
```diff
@@ -132,9 +132,16 @@ private void addTickConfig(JSONObject axis, String key, String varName) throws G
...
-                        tick.put("rotate", 75);
+                    if (isX) {
+                        height = StringWidthUtil.getStringWidth(largestLabel);
+                        if (largestLabel.length() > 5 && height != -1) {
+                            tick.put("rotate", 75);
+                        }
+                        // Use configuration from HQ
+                        String rotation = mData.getConfiguration("x-label-rotation");
+                        if (rotation != null) {
+                            tick.put("rotate", Integer.parseInt(rotation));
```
shubham1g5, line 143: 1. I don't think it'a a good idea to let user input the rotation here. There would only be a limited range of rotation values here that would result a good output for the user. 

2. I also don't think it's a good idea for us to add settings for small UI behaviours like this without a good use case. Can you provide more info on why does the standard way of rotating the labels don't work out for the use case you are trying to solve for ?

ShivamPokhriyal: Agree that there will be a limited range of good values here, but if user is going to use this value, it'll be upto them to dry-run the app and test whether their input produces a good output or not. 
I don't really know a good use-case, I just thought that the feedback was really simple and achievable so why not do it.

shubham1g5: I see, It would be great to have more details on the use case here. Best solution here would be if we can solve the issue on product level without providing anything custom.

PR #1009, opened 2021-06-28
Renamed function

orangejenny:
Tiny. Cross-requested with https://github.com/dimagi/commcare-android/pull/2507


shubham1g5 (approved):
Ahh Thanks for correcting. It took me 2 long mins to spot the difference 🙈

PR #1011, opened 2021-07-05
Rotate 75 degrees for all values

ShivamPokhriyal:
Fix for https://dimagi-dev.atlassian.net/browse/QA-3223
Earlier we were only rotating labels when someone was using json specified `x-labels`. Now, we'll be rotating x labels for all values.

### DIFF
--- a/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
+++ b/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
@@ -86,7 +86,7 @@ public class AxisConfiguration extends Configuration {
         JSONObject tick = new JSONObject();
         boolean usingCustomText = false;
         boolean isX = key.startsWith("x");
-        int height = -1;
+        int largestHeight = -1;
 
         mVariables.put(varName, "{}");
         if (labelString != null) {
@@ -131,7 +131,7 @@ public class AxisConfiguration extends Configuration {
                     mVariables.put(varName, labels.toString());
                     usingCustomText = true;
                     // These custom labels can be large, rotating them ensures that they're shown correctly on X axis.
-                    height = StringWidthUtil.getStringWidth(largestLabel);
+                    largestHeight = StringWidthUtil.getStringWidth(largestLabel);
                 } catch (JSONException e) {
                     // Assume labelString is just a scalar, which
                     // represents the number of labels the user wants.
@@ -161,13 +161,13 @@ public class AxisConfiguration extends Configuration {
         if ((isX && !isRotatedBarGraph) || (isRotatedBarGraph && key.startsWith("y"))) {
             String largestLabel = mData.getLargestXLabel();
             if (largestLabel != null && !largestLabel.isEmpty()) {
-                int largestHeight = StringWidthUtil.getStringWidth(largestLabel);
-                if (height < largestHeight) {
-                    height = largestHeight;
+                int height = StringWidthUtil.getStringWidth(largestLabel);
+                if (height > largestHeight) {
+                    largestHeight = height;
                 }
             }
             tick.put("rotate", 75);
-            axis.put("height", height);
+            axis.put("height", largestHeight);
         }
         if (tick.length() > 0) {
             axis.put("tick", tick);
--- a/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
+++ b/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
@@ -191,7 +191,7 @@ public class AxisConfiguration extends Configuration {
 
         // Show title regardless of whether or not it exists, to give all graphs consistent padding
         JSONObject label = new JSONObject();
-        label.put("text", "title, rad");
+        label.put("text", title);
         label.put("position", position);
         axis.put("label", label);
     }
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/core/graph/c3/DataConfiguration.java
```diff
@@ -373,15 +375,27 @@ private void setColumns(String xID, String yID, SeriesData s) throws GraphExcept
...
             } else {
                 if (mData.getType().equals(GraphUtil.TYPE_TIME)) {
-                    xValues.put(parseTime(p.getX(), description));
+                    currentXLabelStr = parseTime(p.getX(), description);
+                    xValues.put(currentXLabelStr);
                 } else {
-                    xValues.put(parseDouble(p.getX(), description));
+                    double val = parseDouble(p.getX(), description);
+                    xValues.put(val);
+                    currentXLabelStr = "" + val;
```
shubham1g5, line 390: nit: should use String.valueOf(), It's a bad practice to use String Concatenation for string conversion

src/main/java/org/commcare/core/graph/c3/DataConfiguration.java
```diff
@@ -373,15 +375,27 @@ private void setColumns(String xID, String yID, SeriesData s) throws GraphExcept
...
                 }
             }
+            if (isRotatedBarGraph) {
+                currentXLabelStr = "" + parseDouble(p.getY(), description);
```
shubham1g5, line 394: String.valueof()

src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
```diff
@@ -187,7 +191,7 @@ private void addTitle(JSONObject axis, String key, String position) throws JSONE
 
         // Show title regardless of whether or not it exists, to give all graphs consistent padding
         JSONObject label = new JSONObject();
-        label.put("text", title);
+        label.put("text", "title, rad");
```
shubham1g5, line 194: what does this do ?

src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
```diff
@@ -162,11 +158,19 @@ private void addTickConfig(JSONObject axis, String key, String varName) throws G
             }
         }
 
+        if ((isX && !isRotatedBarGraph) || (isRotatedBarGraph && key.startsWith("y"))) {
+            String largestLabel = mData.getLargestXLabel();
+            if (largestLabel != null && !largestLabel.isEmpty()) {
+                int largestHeight = StringWidthUtil.getStringWidth(largestLabel);
+                if (height < largestHeight) {
```
shubham1g5, line 165: puzzled why do we need to redo the comparison here when we are already using the `largestLabel` to get the height.

shubham1g5 (changes requested):

src/main/java/org/commcare/core/graph/c3/DataConfiguration.java
```diff
@@ -373,15 +375,27 @@ private void setColumns(String xID, String yID, SeriesData s) throws GraphExcept
                 xValues.put(barIndex + 1);
                 mBarCount = Math.max(mBarCount, barIndex + 1);
                 if (addBarLabels) {
-                    mBarLabels.put(p.getX());
+                    if (!isRotatedBarGraph) {
+                        currentXLabelStr = p.getX();
+                    }
+                    mBarLabels.put(currentXLabelStr);
```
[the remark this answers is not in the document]
ShivamPokhriyal: So I'm not making any changes to labels here. The additional code is just for finding the largest labels. 
```
if (!isRotatedBarGraph) {
                        currentXLabelStr = p.getX();
                    }
```
For rotated bar graph, we have rotated axis so not using x axis values to find the largest in this case.

src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
ShivamPokhriyal: Ahh, I actually used it for testing, accidentally pushed it. Thanks for catching

src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
ShivamPokhriyal: Here I'm using `mData.getLargestXLabel()` which is getting the value from `data` configuration. It is different than the values used above in the json config.

shubham1g5: I see, it would be nice to rename the variables here accordingly. ` if (height < largestHeight) {` the statement is confusing because semantically it's always true.

src/main/java/org/commcare/core/graph/c3/DataConfiguration.java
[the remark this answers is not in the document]
shubham1g5: sorry don't think I was clear last time. `currentXLabelStr` would be empty when `isRotatedBarGraph` is `true` and would cause `mBarLabels` to add an empty string when earlier it was adding a label. Is that right ?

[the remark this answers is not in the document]
ShivamPokhriyal: Ohhhh, I see what you're saying now. Thanks!

shubham1g5 (approved):

PR #1013, opened 2021-07-07
Mobile  (2.52) to Formplayer Merge

shubham1g5:
Changes seems quite minimal in this direction, mostly error handling and some refactors. Going to request a sanity QA for this. 

QA - https://dimagi-dev.atlassian.net/browse/QA-3246

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,7 @@ dependencies {
 
     cliCompile 'org.json:json:20140107'
     cliCompile group: 'regexp', name: 'regexp', version: '1.3'
-    cliCompile 'kxml2:kxml2:2.3.0'
+    cliCompile 'com.github.stefanhaustein:kxml2:2.4.1'
     cliCompile 'commons-cli:commons-cli:1.3.1'
     cliCompile 'xpp3:xpp3:1.1.4c'
     cliCompile 'joda-time:joda-time:2.9.4'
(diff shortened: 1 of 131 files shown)


ctsims (approved):
Look  good to me. 

I'm guessing it would be worth QA going through the Form Volatility testing workflows post-merge. They come up quite frequently in false alarms for customers, so since this PR makes changes in the code area if we don't validate that they are still working afterwards it might trigger someone to over-react if a ticket implies an issue.

shubham1g5:
@ctsims I might be missing something, though this doesn't seem to touch any parts relevant to Form Volatility. Are you referring to the `FormMetaIndicatorUtil` class, if so that's not in use on Formplayer yet.

ctsims:
Yeah, I was thinking of the Pragma extraction, I assumed it refactored over to the FormDef, but you're right that it doesn't appear to. Good catch!

shubham1g5:
QA passed on this, going for the merge.

ShivamPokhriyal:
I'm guessing there are a few more commits in 2.52 not included here.

shubham1g5:
yeah we can do another post 2.52 merge for those. This is needed to be in formplayer for https://github.com/dimagi/formplayer/pull/984 , so did an early merge.

PR #1014, opened 2021-07-08
Fix issues with graph label overlapping

ShivamPokhriyal:
https://dimagi-dev.atlassian.net/browse/QA-3223

This PR addresses 2 issues: 
- When we use horizontal bar graphs(rotated=true), y and x axis interchange. https://c3js.org/reference.html#axis-x-height C3 supports setting height for x-axis but not for y-axis, so the label overlap with ticks. 
To overcome this, for rotated bar graphs, we'll not set the height and simply show the labels inside the graph [inner-right](https://c3js.org/samples/axes_label_position.html)
- Secondly, for time graphs, c3 accepts the ticks in format [YYYY-MM-DD HH:MM:SS](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/core/graph/c3/Configuration.java#L49), but in the graph we truncate the ticks and only show [YYYY-MM-DD](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java#L148) part. This was causing issues because the height calculation was taking the full string. This PR will change that and use the `largest label` to store  the actual string that'll be shown on graph. 

Rotated Bar Graph Before: 
before_rotated

Rotated Bar Graph Updated:
after_rotated

Time Graph Before: 
before_time

Time Graph Updated: 
after_time

### DIFF
--- a/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
+++ b/src/main/java/org/commcare/core/graph/c3/AxisConfiguration.java
@@ -158,7 +158,7 @@ public class AxisConfiguration extends Configuration {
             }
         }
 
-        if ((isX && !isRotatedBarGraph) || (isRotatedBarGraph && key.startsWith("y"))) {
+        if (isX && !isRotatedBarGraph) {
             String largestLabel = mData.getLargestXLabel();
             if (largestLabel != null && !largestLabel.isEmpty()) {
                 int height = StringWidthUtil.getStringWidth(largestLabel);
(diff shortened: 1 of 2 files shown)


shubham1g5 (approved):
Looks good, though would be nice for PR to have screenshots for before and after of issues you menioned in description.

ShivamPokhriyal:
Updated description with screenshots

shubham1g5:
Thanks looks great!

PR #1017, opened 2021-07-14
Allow searching for Blank values

shubham1g5:
cross-request: https://github.com/dimagi/formplayer/pull/987

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -177,9 +177,13 @@ public class QueryScreen extends Screen {
                 String[] indicesOfSelectedChoices = RemoteQuerySessionManager.extractSelectChoices(answer);
                 ArrayList<String> selectedChoices = new ArrayList<>(indicesOfSelectedChoices.length);
                 for (int i = 0; i < indicesOfSelectedChoices.length; i++) {
-                    int choiceIndex = Integer.parseInt(indicesOfSelectedChoices[i]);
-                    if (choiceIndex < selectChoices.size() && choiceIndex > -1) {
-                        selectedChoices.add(selectChoices.get(choiceIndex).getValue());
+                    if (indicesOfSelectedChoices[i].isEmpty()) {
+                        selectedChoices.add("");
+                    } else {
+                        int choiceIndex = Integer.parseInt(indicesOfSelectedChoices[i]);
+                        if (choiceIndex < selectChoices.size() && choiceIndex > -1) {
+                            selectedChoices.add(selectChoices.get(choiceIndex).getValue());
+                        }
                     }
                 }
                 answer = String.join(RemoteQuerySessionManager.MULTI_SELECT_DELIMITER, selectedChoices);
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -2,6 +2,7 @@ package org.commcare.util.screen;
 
 import com.google.common.collect.Multimap;
 
+import org.commcare.cases.util.StringUtils;
 import org.commcare.modern.session.SessionWrapper;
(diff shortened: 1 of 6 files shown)


snopoke (approved):

snopoke (commented):
A little surprised there are no unit tests for these changes in core

src/main/java/org/javarosa/xpath/expr/XPathEqExpr.java
```diff
@@ -70,7 +70,11 @@ public void writeExternal(DataOutputStream out) throws IOException {
     public static boolean testEquality(Object aval, Object bval) {
         boolean eq = false;
 
-        if (aval instanceof Boolean || bval instanceof Boolean) {
+        if (aval == null && bval == null) {
```
shubham1g5, line 73: without these changes we get a NPE below at line 98/99 when comparing with a `null` value

ctsims, line 73: Hi Shubham

Unfortunately this may require a deeper change.  The engine can't actually handle null values for XPath expression outputs, and it should be treated as an error for any values to ever be null.

I can't quite find a concrete code reference that describes this, although documentation in various places does reflect it:
https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xpath/expr/FunctionUtils.java#L170

https://github.com/dimagi/commcare-core/blob/d9066ea561f4ce85717b25263f9ddad695c98b29/src/test/java/org/javarosa/core/model/data/test/StringDataTests.java#L50

ctsims (changes requested):

src/main/java/org/javarosa/xpath/expr/XPathEqExpr.java
shubham1g5: I see, I am running into this limitation while evaluating the dependent itemset choices for ex - `<itemset nodeset="instance('district')/district_list/district[state_id = $state]">` . Since this PR changes logic to initialize the fields with null values instead of empty string, the evalualtion for `[state_id = null]` fails due to a NPE in this function. It seems likely that we can still set a `null` variable as an empty string just for the purpose of evaluating itemsets without introducing any regressions. Does that sound like a good workaround on this ?

ctsims: I think the core question is whether you want the engine to be able to differentiate between a non-existent input and a blank input via the XPath Variable itself. It sounds like you don't need to, and that the null value is relevant for the URL endpoint generation but not the expressions here. I think the cleanest solution would definitely be that the Query Screen *always* populates a value for all of its inputs into the XPath evaluation context, and it checks for null inputs and replaces them with empty strings. That would be consistent with what the normal engine would do with a reference which is valid (IE: The node exists in the form structure) but is currently irrelevant or empty.

ctsims (approved):
This looks good to me although I'm still a bit confused about the semantics of the allow_blank_value flag. I can bring those to the larger feature conversation.

src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -110,7 +111,7 @@ public String buildUrl(boolean skipDefaultPromptValues) {
             QueryPrompt prompt = userInputDisplays.get(key);
             for (String value : queryParams.get(key)) {
                 if (prompt != null && prompt.isSelect()) {
```
snopoke, line 113: @shubham1g5 I notice that in `FormplayerQueryScreen` you removed the `prompt.isSelect()` check (https://github.com/dimagi/formplayer/commit/4bc37f2bc6f87d6ad37f02c28ebb68ebbb0a0114#diff-d43851ef662a8e31af133ace2b4c60e87e8df0f14f8b70293ec884762f183518L33) should that have been done here too?

shubham1g5: yes, we should remove it from here as well.

PR #1018, opened 2021-07-26
Added session endpoints to CLI

orangejenny:
This adds an `-e` argument to the CLI to start at an endpoint.

Since endpoints can have an arbitrary number of arguments, it expects a comma-separated value like `endpoint_id,arg1,arg2,...`

I'm still thinking over instead implementing this as a navigation command you can use at any point, since an endpoint could be used to start any session. Open to feedback on whether or not this makes sense as a command-line arg.


orangejenny:
Closing this one - I'm going to open a PR with CLI fixes and then make this a PR on top of that.

PR #1019, opened 2021-07-27
Fixed case details and case search workflows in CLI

orangejenny:
The CLI doesn't seem to work properly for either case details or case search:
* Case details don't get shown at all. You see the case list, pick a case, and get sent straight to the next menu.
* On case search, QueryScreen hits HQ and gets results but never displays them, it just keeps asking you to enter search fields.

The main fix is to make the implementations of `handleInputAndUpdateSession` and `handleInputAndUpdateHost` consistent with their descriptions:
* [handleInputAndUpdateSession](https://github.com/dimagi/commcare-core/blob/2e6a96b551c3c4e8fd57e787a0e6884f824f10fa/src/cli/java/org/commcare/util/screen/Screen.java#L33-L44) is supposed to return true if the current screen is "done" and the app can move on...but `ApplicationHost` assigns the return value of `handleInputAndUpdateSession` to a `screenIsRedrawing` variable that tells the CLI to **stay** on the same screen ([link](https://github.com/dimagi/commcare-core/blob/2e6a96b551c3c4e8fd57e787a0e6884f824f10fa/src/cli/java/org/commcare/util/cli/ApplicationHost.java#L248)).
* [handleInputAndUpdateHost](https://github.com/dimagi/commcare-core/blob/2e6a96b551c3c4e8fd57e787a0e6884f824f10fa/src/cli/java/org/commcare/util/screen/Subscreen.java#L15-L32) is a similar function used by case list/detail subscreens to indicate to their host EntityScreen that they're "done" and the app can move on...but EntityScreen's implementation of `handleInputAndUpdateSession` returns the **inverse** of `handleInputAndUpdateHost` ([link](https://github.com/dimagi/commcare-core/blob/2e6a96b551c3c4e8fd57e787a0e6884f824f10fa/src/cli/java/org/commcare/util/screen/CompoundScreenHost.java#L35-L39)).

This PR should not change formplayer behavior, because formplayer never uses the return value of `handleInputAndUpdateSession`. Case details are handled by a separate request in formplayer, so the new `setCurrentScreenToDetail` calls shouldn't affect formplayer.

My only hesitation is it seems like these two workflows have basically never worked, even though they've been around for 4-6 years. @ctsims I'm curious if you have any intuition how likely that is.

This PR also fixes some spacing things that were bugging me and adds a couple of tests.


ctsims (approved):
I find it very plausible that those use cases were never implemented fully, so I wouldn't let that trouble you too much. A lot of the CLI was implemented as use cases came up that needed deeper testing, so detail pages were probably just not on that list.

The boolean returns I have less context on, honestly, but my only point to discuss and de-risk would have been the formpalyer overlap, so if you've validated that formplayer doesn't use the return value path at all, everything seems good to me.

PR #1020, opened 2021-07-27
Added CLI support for session endpoint

orangejenny:
(This is a redo of https://github.com/dimagi/commcare-core/pull/1018 with just the new functionality, no other changes.)

This adds an `-e` argument to the CLI to start at an endpoint.

Since endpoints can have an arbitrary number of arguments, it expects a comma-separated value like `endpoint_id,arg1,arg2,...`

I'm still thinking over instead implementing this as a navigation command you can use at any point, since an endpoint could be used to start any session. Open to feedback on whether or not this makes sense as a command-line arg.


src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -108,15 +110,39 @@ public void setRestoreToDemoUser() {
         mRestoreStrategySet = true;
     }
 
-    public void run() {
+    public void advanceSessionWithEndpoint(String[] endpointIdAndArgs) {
```
ctsims, line 113: In the other platforms, I believe we explicitly reset the session state before we begin any endpoint execution. Even if the CLI can't currently end up in that state it seems worth being consistent with the execution structure.

ctsims (approved):
One small comment to improve consistency between the platforms.

I think making endpoint execution a command line arg like this make a lot of sense. I'm not quite as sold on the serialization format for the args.

What do you think about emulating something like CURL where when there are multiple pairs to be passed in they come as their own args, like the following?

`cli -e ENDPOINT -A "parameter:value" -A "parameter2:value2" ...`

It's not obviously much better because we still need a fake delimeter (with curl, the header is actually one slug, not a parsed syntax), but it feels somewhat more consistent with other cli formats.

src/cli/java/org/commcare/util/cli/ApplicationHost.java
orangejenny: Makes sense. Does resetting the state here mean `clearVolatiles`, or something else?

ctsims: I think this is the android entrypoint

https://github.com/dimagi/commcare-android/blob/35627e56cbe693cf8951e0315ca2f33057b3fc5c/app/src/org/commcare/activities/HomeScreenBaseActivity.java#L222

which follows to 
https://github.com/dimagi/commcare-android/blob/master/app/src/org/commcare/models/AndroidSessionWrapper.java#L70

so I think those two steps are the right ones.

orangejenny: a777f9aa4

orangejenny:
@ctsims Added 03622c53a to update the args format. Endpoint arguments can be either an ordered list or name/value pairs, and the ordered list makes for a cleaner command line (no delimiter), so the format is now `cli -e ENDPOINT -A value1 -A value2 ...`

ctsims (approved):
Thanks! Out of curiosity

PR #1022, opened 2021-08-04
Fixed NullPointerException in IndexedFixtureXmlParser

orangejenny:
I've been hitting this in the CLI on the locations fixture, for users who aren't assigned to a location.

### DIFF
--- a/src/main/java/org/commcare/xml/IndexedFixtureXmlParser.java
+++ b/src/main/java/org/commcare/xml/IndexedFixtureXmlParser.java
@@ -90,7 +90,10 @@ public class IndexedFixtureXmlParser extends TransactionParser<StorageIndexedTre
                 processEntry(entry, indices);
             }
         } else {
-            sandbox.getIndexedFixtureStorage(fixtureName).removeAll();
+            IStorageUtilityIndexed storage = sandbox.getIndexedFixtureStorage(fixtureName);
+            if (storage != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/commcare/xml/IndexedFixtureXmlParser.java
```diff
@@ -90,7 +90,10 @@ private void processRoot(TreeElement root) throws IOException {
                 processEntry(entry, indices);
             }
         } else {
-            sandbox.getIndexedFixtureStorage(fixtureName).removeAll();
+            IStorageUtilityIndexed storage = sandbox.getIndexedFixtureStorage(fixtureName);
```
shubham1g5, line 93: Confirming that even if the user is not assigned any location, a fixture for that location is present with no children/rows in restore ?

shubham1g5 (approved):

src/main/java/org/commcare/xml/IndexedFixtureXmlParser.java
esoergel: That's right, it should look like this:
```xml
<fixture id="locations" indexed="true" user_id="c0bdfa6db8ef44b8af0c9fbb70e4e19e">
  <locations/>
</fixture>
```
Though the fixture isn't present for all domains - only those that actually use locations

PR #1023, opened 2021-08-04
Fixed StackOperation.buildClearFrame

orangejenny:
Found this while experimenting with clear operations in the CLI.

Clear ops [don't support children](https://github.com/dimagi/commcare-core/blob/16397aa3df78c238f6b6f9341ccb786cdc89d428/src/main/java/org/commcare/xml/StackOpParser.java#L40), so this constructor passes in null, but that ultimately breaks serialization:
```
Unhandled Fatal Error executing CommCare appjava.lang.NullPointerException
        at org.javarosa.core.util.externalizable.ExtWrapList.<init>(ExtWrapList.java:24)
        at org.javarosa.core.util.externalizable.ExtWrapList.<init>(ExtWrapList.java:19)
        at org.commcare.suite.model.StackOperation.writeExternal(StackOperation.java:122)
        ...
```

### DIFF
--- a/src/main/java/org/commcare/suite/model/StackOperation.java
+++ b/src/main/java/org/commcare/suite/model/StackOperation.java
@@ -60,7 +60,7 @@ public class StackOperation implements Externalizable {
     }
 
     public static StackOperation buildClearFrame(String ifCondition) throws XPathSyntaxException {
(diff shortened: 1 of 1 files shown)


shubham1g5 (approved):
nice find. I am not aware of any use case where we use a clear frame in suite.xml which is probably why this had not come up till now.

PR #1024, opened 2021-08-05
Added CLI support for case claim in session endpoints

orangejenny:
~~PR into https://github.com/dimagi/commcare-core/pull/1020~~ (changed to PR against master)

https://dimagi-dev.atlassian.net/browse/USH-1185

Discussed [here](https://docs.google.com/document/d/1w7oRuH9skVhNB1vuhIa8XWbSBoUWNj3f_N6lnU_6lnc/edit?disco=AAAANcwHc5Q) in spec. At the time of that conversation, I thought it would be sufficient to put a case claim command into the endpoint's stack and no CLI changes would be needed. However, the sync request is part of the SyncScreen's input handling, even though there's no user input, and executing stack operations updates the session, but it doesn't run any of the input logic.

My compromise is to set up the endpoint logic as a series of push operations, and after each operation, have the CLI run the sync screen if that's what's up next. This is similar to how we handle the case claim search-first, etc. workflows on formplayer, where on every loop of `advanceSessionWithSelections` we [check](https://github.com/dimagi/formplayer/blob/6ee6ea2370d90c0c8906be7d05439ce28af7a67d/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L247-L249) if there's an autolaunch or query screen that needs to be handled.


ctsims (approved):
This is a really tricky thing to work around. In particular it worries me a lot that this change will introduce a functional difference between how `executeStackOperations` is followed between the CLI and the core code. 

I don't think it's important enough to be a blocking issue, but I think it would be worth reviewing if it's possible to make the adjustment at a lower level such that "Operating a set of stack operations" is something that works without a walk, if for no other reason than to surface if that's not possible cleanly for a very real reason which should affect our reliance on that.

ctsims:
Hi @orangejenny,

Yeah, by walk I mean execution of the presentation layer directly as opposed to an in-place manipulation in the 'headless' session state layer.

Thinking through things again I realized that I thought the disparity before was the CLI model, but it occurs to me that this is actually almost certainly due to the fact that in every runtime environment the actual `<query>` part of a query is done in the presentation layer and the mobile app also wouldn't be able to execute this walk independently.

We absolutely could refactor to make the query datum mechanism be able to independently trigger the query with a callback, but I think it's unreasonable to require that in the scope of this PR since I think doing so would want to happen deeper down (by decoupling the presentation part of that request with the execution of the query) rather than injecting into the stack runner.

ctsims (approved):

PR #1030, opened 2021-08-20
cli remote query entry

snopoke:
Changes to the CLI to allow using a query in a form entry session to get data from remote:

Suite file entry:
```xml
  <entry>
    <form>http://openrosa.org/formdesigner/C6C1CFA4-6994-4CE9-8C56-16673A5EEED2</form>
    <command id="m0-f0">
      <text>
        <locale id="forms.m0f0"/>
      </text>
    </command>
    <instance id="commcaresession" src="jr://instance/session"/>
    <session>
      <datum id="case_id_new_case_0" function="uuid()"/>
      <query url="http://localhost:8000/a/test/phone/get_case/" storage-instance="registry" template="case" default_search="true">
        <data key="case_type" ref="'case'"/>
        <data key="case_id" ref="instance('commcaresession')/session/data/case_id_new_case_0"/>
        <data key="registry" ref="'test_registry'"/>
      </query>
    </session>
  </entry>
```

form XML:
```xml
<instance id="registry" src="jr://instance/remote"/>
<bind nodeset="/data/registry_result" calculate="instance('registry')/results/case[@case_id='2f462e82-1cea-4638-896b-f36b3dc73ee7']/case_name" />
```

I'm not 100% sure these are the right set of changes but they do work locally. Also unclear how to test this in a unit test.


shubham1g5 (approved):
Looks good to me. 

> Also unclear how to test this in a unit test.

We can probably write a parse test like [this](https://github.com/dimagi/commcare-core/blob/master/src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java) in core and then write a more through test in Formplayer to check for the remote instance population with the help of a query mock like [this one](https://github.com/dimagi/formplayer/blob/master/src/test/java/org/commcare/formplayer/tests/CaseClaimTests.java#L217).

snopoke:
@shubham1g5 I've updated the code to allow mocking the response so that I can add a more meaningful test which actually tests the code change I made to `CommCareInstanceInitializer`. Let me know what you think.

snopoke:
@shubham1g5 this change is PR'd into `master`. Is that the correct branch to merge it into. I assume then I'd need to apply the same changes to the `formplayer` branch. Are there any other branches that would need these changes?

shubham1g5:
@snopoke Probably best to PR this into `formplayer` for now if we only need this feature on Web Apps. Mobile team updates the master with `formplayer` before mobile releases go to QA so it should make it way through to `master` eventually

shubham1g5 (approved):

snopoke:
Rebasing off formplayer branch was the wrong move: new PR here: https://github.com/dimagi/commcare-core/pull/1038

PR #1035, opened 2021-08-31
Case Search: Override defaults params with user inputs

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1199

```
Evaluate filters to be applied to the search uri using following rules -
     * 1. If no defaults are specified for a property, add the user input for search property as it is
     * 2. If defaults are specified, only add the user input if it's also specified in defaults
     * 3. If no user input is specified, add the defaults for the property as it is
```

cross-request: https://github.com/dimagi/formplayer/pull/1003

### DIFF
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -117,32 +117,55 @@ public class RemoteQuerySessionManager {
     }
 
     /**
-     * @param skipDefaultPromptValues don't apply the default value expressions for query prompts
+     * Evaluate filters to be applied to the search uri using following rules -
+     * 1. If no defaults are specified for a property, add the user input for search property as it is
+     * 2. If defaults are specified, only add the user input if it's also specified in defaults
+     * 3. If no user input is specified, add the defaults for the property as it is
+     *
+     * @param skipUserInput don't populate user inputs for query prompts into search params
      * @return filters to be applied to case search uri as query params
      */
-    public Multimap<String, String> getRawQueryParams(boolean skipDefaultPromptValues) {
-        Multimap<String, String> params = ArrayListMultimap.create();
+    public Multimap<String, String> getRawQueryParams(boolean skipUserInput) {
+        Multimap<String, String> defaultParams = ArrayListMultimap.create();
         Multimap<String, XPathExpression> hiddenQueryValues = queryDatum.getHiddenQueryValues();
         for (String key : hiddenQueryValues.keySet()) {
             for (XPathExpression xpathExpression : hiddenQueryValues.get(key)) {
                 String evaluatedExpr = evalXpathExpression(xpathExpression, evaluationContext);
-                params.put(key, evaluatedExpr);
+                defaultParams.put(key, evaluatedExpr);
             }
         }
 
-        if (!skipDefaultPromptValues) {
+        Multimap<String, String> searchParams = ArrayListMultimap.create();
+
+        // Populate User Params
+        if (!skipUserInput) {
             for (Enumeration e = userAnswers.keys(); e.hasMoreElements(); ) {
                 String key = (String)e.nextElement();
                 String value = userAnswers.get(key);
-                if (!(params.containsKey(key) && params.get(key).contains(value))) {
-                    if (value != null) {
-                        params.put(key, userAnswers.get(key));
+
+                // if defaults are defined, make sure userParams are restricted to defaults
+                String[] choices = RemoteQuerySessionManager.extractMultipleChoices(value);
+                for (String choice : choices) {
+                    if (choice != null &&
+                            ((defaultParams.containsKey(key) && defaultParams.get(key).contains(choice))
+                                    || !defaultParams.containsKey(key)) &&
+                            !(searchParams.containsKey(key) && searchParams.get(key).contains(choice))) {
+                        searchParams.put(key, choice);
                     }
                 }
             }
         }
-        return params;
+
+        // Add defaults if no user inputs have been defined
+        for (String key : defaultParams.keySet()) {
+            if (!searchParams.containsKey(key)) {
+                searchParams.putAll(key, defaultParams.get(key));
+            }
+        }
+
+        return searchParams;
     }
+
 
     public static String evalXpathExpression(XPathExpression expr,
                                              EvaluationContext evaluationContext) {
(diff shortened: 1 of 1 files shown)


shubham1g5:

shubham1g5:

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -117,33 +117,56 @@ public URL getBaseUrl() {
...
-                    if (value != null) {
-                        params.put(key, userAnswers.get(key));
+
+                // if defaults are defined, make sure userParams are restricted to defaults
+                String[] choices = RemoteQuerySessionManager.extractMultipleChoices(value);
+                for (String choice : choices) {
+                    if (choice != null &&
+                            ((defaultParams.containsKey(key) && defaultParams.get(key).contains(choice))
+                                    || !defaultParams.containsKey(key)) &&
+                            !(searchParams.containsKey(key) && searchParams.get(key).contains(choice))) {
```
snopoke, line 152: hard to follow the logic here - would be useful to break it up into some separate vars:
```java
if (choice == null) {
    continue
}

boolean noDefault = !defaultParams.containsKey(key)
boolean validDefault = defaultParams.containsKey(key) && defaultParams.get(key).contains(choice)
boolean duplicateValue = searchParams.containsKey(key) && searchParams.get(key).contains(choice)

if ( (noDefault || validDefault) && !duplicateValue) {
    searchParams.put(key, choice);
}
```

Incidentally `noDefault || validDefault` (`!defaultParams.containsKey(key) || (defaultParams.containsKey(key) && defaultParams.get(key).contains(choice))`) can be simplified to `!defaultParams.containsKey(key) || defaultParams.get(key).contains(choice)` so we can make this:

```java
boolean validParam = !defaultParams.containsKey(key) || defaultParams.get(key).contains(choice)
boolean duplicateValue = searchParams.containsKey(key) && searchParams.get(key).contains(choice)

if (validParam && !duplicateValue) {
    searchParams.put(key, choice);
}
```

Explanation: `!A or (A & B)` -> `(!A or A) & (!A or B)` -> `True & (!A or B)` -> `!A or B`

snopoke (changes requested):

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -117,33 +117,56 @@ public URL getBaseUrl() {
...
+        if (!skipUserInput) {
             for (Enumeration e = userAnswers.keys(); e.hasMoreElements(); ) {
                 String key = (String)e.nextElement();
                 String value = userAnswers.get(key);
-                if (!(params.containsKey(key) && params.get(key).contains(value))) {
-                    if (value != null) {
-                        params.put(key, userAnswers.get(key));
+
+                // if defaults are defined, make sure userParams are restricted to defaults
+                String[] choices = RemoteQuerySessionManager.extractMultipleChoices(value);
```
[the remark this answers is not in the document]
shubham1g5: yes, this was done to support blank value searches in combination with a non empty value search - https://github.com/dimagi/formplayer/blob/blankValueSearch/src/test/java/org/commcare/formplayer/tests/CaseClaimTests.java#L114

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
shubham1g5: thanks this is so much better. https://github.com/dimagi/commcare-core/pull/1035/commits/6d82f12ce3803181464b6dca0c7340ecca73f645

snopoke (approved):

snopoke (commented):
@shubham1g5 just re-reading this again I'm a bit confused about it. Initially I had misunderstood what was going on here in terms of the difference between `hiddenQueryValues` and `userAnswers`.

After looking into it more I now understand that:
- `hiddenQueryValues` represents the 'data' elements in the query
- `userAnswers` represent the values collected from the user for 'prompts'.

Can you confirm that the values from `userAnswers` are already in the `evaluationContext` when we run the xpath expressions for `hiddenQueryValues`?

I'm a bit confuse why we add the `userAnswers` into the query params at all. Surely the `data` elements fully define what should be in the request?

Using this as an example:
```
<data key="key1" ref="xpath"/>
<prompt key="key1">
<prompt key="key2">
```

My read of the code before this changes is that the method would first evaluate the xpath for data[key1] and then it would not overwrite that with prompt[key1] but it would also add prompt[key2].

After this change it seems like the prompt values would take precedence over the data values. If that's correct I don't think that makes sense.

snopoke:
Looking at the tests in the formplayer PR I think there are 2 things which are clear to me:
1. I don't think should validate the 'prompt' values against the values from the 'data' elements
2. My intuition is that we should leave it the way it is right now where 'data' values take precedence over 'prompt' values. My main reason for saying that is because changing it has the potential to break existing apps which are relying on the current precedence.

If you agree with those then I don't think we should merge this PR.

shubham1g5:
>I'm a bit confuse why we add the userAnswers into the query params at all. Surely the data elements fully define what should be in the request?

the `userAnswers` incorporate the user inputs provided into the query params in addition to the hidden params set by 'data'. Without taking these into account into query params, user input will never be accounted in the case search query.  

> My read of the code before this changes is that the method would first evaluate the xpath for data[key1] and then it would not overwrite that with prompt[key1] but it would also add prompt[key2].

With your example, the query params passed before this change will be - `?key1=evaluated_xpath&key1=user_input_for_key1 &key2=user_input_for_key2` 

> I don't think should validate the 'prompt' values against the values from the 'data' elements

The reason I think it made sense is the model where `data` elements act as an overall filter/scope the user can search in and `userAnswers` are used to narrow down that scope of search. 

I don't think this PR would cause a breaking change because one can't define the same key for <data> elements and <prompt> on HQ except for the `case_type`. HQ gives an error when someone tries to add a search parameter which is also defined in the defaults or vice versa. Though because of that reason this PR also doesn't serve any purpose if the 'case_type` are already getting overriden as mentioned by Ethan [here](https://dimagi-dev.atlassian.net/browse/USH-1199?focusedCommentId=165149).  Therofore it's probably best to not merge it. Closing it for now and can reopen if needed.

snopoke:
@shubham1g5 thanks for the explanation - I had forgotten that it was a mult-value hashmap and so adding new values doesn't override old ones. 

> the userAnswers incorporate the user inputs provided into the query params in addition to the hidden params set by 'data'.

I don't quite understand this but I'll take your word for it.

Overall I agree we should still not merge this.

PR #1036, opened 2021-08-31
Revert "Revert "(formplayer) Fixed case details and case search workf…

orangejenny:
…lows in CLI""

Reverts https://github.com/dimagi/commcare-core/pull/1029, restoring https://github.com/dimagi/commcare-core/pull/1021

Part of fixing https://dimagi-dev.atlassian.net/browse/QA-3382

See https://github.com/dimagi/formplayer/pull/1004 for explanation.


shubham1g5:
Hey @orangejenny I am a bit confused in why this PR would not cause the same issues again. Is it not an exact revert of #1029 ? If not, is there any other change you plan to get with this so that the same issues don't appear again.

shubham1g5:
> Hey @orangejenny I am a bit confused in why this PR would not cause the same issues again. Is it not an exact revert of #1029 ? If not, is there any other change you plan to get with this so that the same issues don't appear again.

Nvm, missed this earlier -  https://github.com/dimagi/formplayer/pull/1004 .

shubham1g5 (approved):

orangejenny:
Hi @shubham1g5 , yeah, it's an exact revert, but the changes in https://github.com/dimagi/formplayer/pull/1004 should keep the previous bug from happening again.

shubham1g5:
We should make sure that this is merged simultaneously with the Formplayer smart linking work to avoid any issues.

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
```diff
@@ -236,15 +236,17 @@ public boolean handleInputAndUpdateHost(String input, EntityScreen host, boolean
             try {
                 int index = Integer.parseInt(input);
                 host.setHighlightedEntity(mChoices[index]);
+                host.setCurrentScreenToDetail();
```
shubham1g5, line 239: Would this not result into moving to the Details screen directly without showing the Entity List screen in the normal workflow ?

orangejenny: In the CLI? At this point we've already shown the entity list screen, so this tells the CLI to show details when it redraws the screen.

orangejenny:
Rebased and force pushed just to clean up the history (remove commit that was later reverted and remove merge commit).

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
shubham1g5: Sorry I meant the question for Web Apps UI, mostly trying to understand if this change can impact the web apps case list -> detail navigation.

orangejenny: I don't think it matters for web apps, because web apps uses the `get_details` endpoint, so I don't think it ever shows details except in response to `get_details`...but I don't know a lot about the `get_details` workflow, let me verify that.

orangejenny: Looked at the code a bit and am more confident: the only place formplayer creates an `EntityDetailListResponse` is when responding to the specific `get_details` request, in `MenuController.getDetails`. In response to navigation requests, `advancedSessionWithSelections` returns either a form entry response or the return value of `getNextMenu`, which when it gets an EntityScreen doesn't look at the current subscreen, it always sends back an EntityListResponse: https://github.com/dimagi/formplayer/blob/2c916265655569583a5a397dc2c98912196a8379/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L132-L153

shubham1g5: Makes sense to me, thanks for double checking this!

shubham1g5 (approved):

orangejenny:
Commented on this in https://github.com/dimagi/formplayer/pull/1004, but to make sure it's clear on this PR, this PR no longer depends on any formplayer code. Merging!

PR #1040, opened 2021-09-10
Adds Query Datum Support for a Form

shubham1g5:
Changes to retain the xml query data instances into the form. The major part of the PR deals with serializing the objects correctly in order to preserve the Session Frame with the instance information into the Form Session. 

cross-request: https://github.com/dimagi/formplayer/pull/1006


src/main/java/org/commcare/suite/model/StackFrameStep.java
```diff
@@ -110,6 +110,10 @@ public ExternalDataInstance getXmlInstance() {
         return xmlInstance;
     }
 
+    public void setXmlInstance(ExternalDataInstance xmlInstance) {
```
shubham1g5, line 113: I don't really like making this mutable but could not come up with a better way either.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -189,38 +189,40 @@ public void populateItemSetChoices(QueryPrompt queryPrompt) {
...
-                QueryPrompt queryPrompt = userInputDisplays.get(promptId);
-                if (queryPrompt.isSelect()) {
-                    String answer = userAnswers.get(promptId);
-                    populateItemSetChoices(queryPrompt);
-                    String[] selectedChoices = extractMultipleChoices(answer);
-                    ArrayList<String> validSelectedChoices = new ArrayList<>();
-                    for (String selectedChoice : selectedChoices) {
-                        if (checkForValidSelectValue(queryPrompt.getItemsetBinding(), selectedChoice)) {
-                            validSelectedChoices.add(selectedChoice);
+        if (userInputDisplays.size() > 0) {
```
shubham1g5, line 192: this `if` is the only change in this file

src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -949,14 +959,28 @@ public static CommCareSession restoreSessionFromStream(CommCarePlatform ccPlatfo
...
+        }
+
+        restoredSession.syncState();
         return restoredSession;
     }
 
     public void serializeSessionState(DataOutputStream outputStream) throws IOException {
         frame.writeExternal(outputStream);
         ExtUtil.write(outputStream, new ExtWrapList(frameStack));
+        for (StackFrameStep step : frame.getSteps()) {
```
shubham1g5, line 979: This might have potential impact on disk storage and Clayton suggested that we add some sort of a threshold check here before serializing the xml. I still need to add that into this PR.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1408,35 +1409,41 @@ public void readExternal(DataInputStream dis, PrototypeFactory pf) throws IOExce
...
+     * @param isReadOnly      If we are in read only mode and only wants to view form
+     * @param commCareSession Current commcare session the form is opened in
      */
     public void initialize(boolean newInstance, boolean isCompletedInstance,
-                           InstanceInitializationFactory factory, String locale, boolean isReadOnly) {
+                           InstanceInitializationFactory factory, String locale, boolean isReadOnly, CommCareSession commCareSession) {
         for (Enumeration en = formInstances.keys(); en.hasMoreElements(); ) {
             String instanceId = (String)en.nextElement();
             DataInstance instance = formInstances.get(instanceId);
             formInstances.put(instanceId, instance.initialize(factory, instanceId));
```
snopoke, line 1440: just confirming my understanding, this will and an 'empty' instance with the same ID which then get's overridden below with the real instance from the session correct?

src/main/java/org/commcare/session/CommCareSession.java
snopoke: why do we serialise this here with the session instead as part of the SessionFrame?

src/main/java/org/javarosa/core/model/FormDef.java
shubham1g5: This doesn't necessary have an instance with same id as the one present in the session frame xml instances. If it does, yes it will get overriden by the code below.

src/main/java/org/commcare/session/CommCareSession.java
shubham1g5: This serialization ideally belongs in [ExternalDataInstance](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java#L99), though because of [root](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java#L19) being an instance of  interface `AbstractTreeElement` it skips searializing it over there. Here I am assuming that the `root` is always an `TreeElement` to serialize it and thought since we are alreay handling some serialization here it would be the next best place to put this code here. Though this does look a bit sketchy and I would be open to ideas here. Also tagging @ctsims to get this thoughts on the same

shubham1g5: Another issue I am facing to implement the size thereshold is we don't seem to have a way to calculate the size of a `TreeElement` without serializing it and checking the byte size. So I am worried if we implement this we will need to serialize the instance xml twice ?

src/main/java/org/commcare/suite/model/StackFrameStep.java
ctsims: If mutability is the only practical way to get the workflow right, one option to make it safer is to control how mutable the data is in the lifecycle of the frame step. IE: If the instance should only be set once and in a certain order, after it is set you could "seal" the object so that future attempts to change it through the same interface fail loudly. 

That would let you continue to guarantee certain aspects of how the classes work. For most purposes it's still effectively immutable if when you reference it it's either a consistent value or it's uninitialized and errors.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1408,35 +1409,41 @@ public void readExternal(DataInputStream dis, PrototypeFactory pf) throws IOExce
...
+     * @param newInstance     true if the form is to be used for a new entry interaction,
+     *                        false if it is using an existing IDataModel
+     * @param locale          The default locale in the current environment, if provided. Can be null
+     *                        to rely on the form's internal default.
+     * @param isReadOnly      If we are in read only mode and only wants to view form
+     * @param commCareSession Current commcare session the form is opened in
      */
     public void initialize(boolean newInstance, boolean isCompletedInstance,
-                           InstanceInitializationFactory factory, String locale, boolean isReadOnly) {
+                           InstanceInitializationFactory factory, String locale, boolean isReadOnly, CommCareSession commCareSession) {
```
ctsims, line 1436: Why aren't the instances from the session here coming through the IIF? The purpose of that object is to be able to provide instance initializers from different frames of context, so I think we should be relying on it rather than extending initialization to include a new coupled dependency.

src/main/java/org/commcare/session/CommCareSession.java
ctsims: @shubham1g5 

I think it's a bit dangerous to serialize this way directly for the reasons you mentioned. It's hard to gauge the size impact, and if the instance is serialized as the mechanism of persistence it makes it harder to set multiple paths for an instance to recover the instance data. 

My sense is that what we want is to extend the abstraction for **providing** the external input that gets turned into the data instance, and make it possible to either re-fetch data, or store it.

This is the entrypoint which I was thinking of:
https://github.com/dimagi/commcare-core/blob/f15e37c497fd6a1fc4511b8902dc13634fe607e6/src/main/java/org/commcare/session/RemoteQuerySessionManager.java#L144

If we could wrap and defer "Here's the source of an instance" we could serialize that in a way which can be retrieved. 

Then as part of events during the session lifecycle we could push the datum elements to go from "source" to "populated" and during that process determine whether to temporarily cache that response, permanently cache that response, or re-request the response at the appropriate point in the lifecycle.

src/main/java/org/javarosa/core/model/FormDef.java
shubham1g5: Thanks this was a mix up on my part. [Reverted](https://github.com/dimagi/commcare-core/pull/1040/commits/52a35e1ec05aeb4077b882173efa8b157f216dfa ) to Simon's approach on initializing the remote instance from https://github.com/dimagi/commcare-core/pull/1038

src/main/java/org/commcare/session/CommCareSession.java
snopoke: Sounds good to me. @shubham1g5 do you have a sense of how to go about doing that?

shubham1g5: @ctsims @snopoke I am trying adding a [remoteUrl](https://github.com/dimagi/commcare-core/blob/dr_form_query/src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java#L33) field in the `ExternalDataInstance` and retrieiving it later from cache/web while [initializing the FormSession](https://github.com/dimagi/formplayer/blob/dr_form_query/src/main/java/org/commcare/formplayer/session/FormSession.java#L228).

I am struggling with how to best handle failures on retrieving the instance data though one possible way can be to pass a flag to HQ that simply moves user one step back and we ask the user to select the form again so that we can run through the `QueryScreen` code again ? 

Also @ctsims, I am not able to find any obvious hooks into the `CommCareInstanceInitializer` to put this code in that makes it easeir for us to maintain the query cycle. Would definitely appreciate any tips on how to best organise this code.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -149,27 +149,11 @@ public static String evalXpathExpression(XPathExpression expr,
         return FunctionUtils.toString(expr.eval(evaluationContext));
     }
 
-    /**
-     * @return Data instance built from xml stream or the error message raised during parsing
-     */
-    public Pair<ExternalDataInstance, String> buildExternalDataInstance(InputStream instanceStream) {
```
shubham1g5, line 155: removed as one can directly use `ExternalDataInstance.buildFromRemote` while the error handling is moved to top level classes.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1441,6 +1455,41 @@ public void initialize(boolean newInstance, boolean isCompletedInstance,
...
+                    ExternalDataInstance newExternalDataInstance = remoteInstanceFetcher.getRemoteDataInstance(instance.getInstanceId(),
+                            externalDataInstance.useCaseTemplate(),
+                            new URI(externalDataInstance.getRemoteUrl()));
+                    if (newExternalDataInstance != null && newExternalDataInstance.getRoot() != null) {
+                        replacedInstances.add(newExternalDataInstance);
+                    }
+                } catch (UnfullfilledRequirementsException | XmlPullParserException |
+                        InvalidStructureException | IOException | URISyntaxException e) {
+                    String errorMessage;
+                    if (e instanceof URISyntaxException) {
```
snopoke, line 1477: why not split these out into their own catch blocks instead of doing type checking?

src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -146,23 +152,31 @@ public String buildUrl(boolean skipDefaultPromptValues) {
...
         }
-        Pair<ExternalDataInstance, String> instanceOrError =
-                remoteQuerySessionManager.buildExternalDataInstance(responseData);
+        Pair<ExternalDataInstance, String> instanceOrError;
+        try {
+            ExternalDataInstance instance = ExternalDataInstance.buildFromRemote(getQueryDatum().getDataId(), responseData, url, getQueryDatum().useCaseTemplate());
+            instanceOrError =  new Pair<>(instance, "");
+        } catch (InvalidStructureException | IOException
+                | XmlPullParserException | UnfullfilledRequirementsException e) {
+            e.printStackTrace();
```
snopoke, line 176: nit: printStackTrace

src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -146,23 +152,31 @@ public String buildUrl(boolean skipDefaultPromptValues) {
...
         Request request = new Request.Builder()
-                .url(url)
+                .url(queryUrl)
                 .header("Authorization", credential)
                 .build();
         return client.makeRequest(request);
     }
 
-    public Pair<ExternalDataInstance, String> processResponse(InputStream responseData) {
+    public Pair<ExternalDataInstance, String> processResponse(InputStream responseData, String url) {
```
snopoke, line 165: in the formplayer PR it seems this method call was replaced with something else (https://github.com/dimagi/formplayer/pull/1006#discussion_r725955421). Is this method still needed?

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1428,15 +1426,21 @@ public void initialize(boolean newInstance, InstanceInitializationFactory factor
...
+    public void initialize(boolean newInstance, boolean isCompletedInstance, InstanceInitializationFactory factory,
+                           String locale, boolean isReadOnly, @Nullable RemoteInstanceFetcher remoteInstanceFetcher) throws RemoteInstanceFetcher.RemoteInstanceException {
         for (Enumeration en = formInstances.keys(); en.hasMoreElements(); ) {
             String instanceId = (String)en.nextElement();
             DataInstance instance = formInstances.get(instanceId);
             formInstances.put(instanceId, instance.initialize(factory, instanceId));
         }
 
+        if (remoteInstanceFetcher != null) {
+            tryAttachingRemoteInstances(remoteInstanceFetcher);
```
snopoke, line 1441: Just for my understanding, why is this happening in the formDef? Surely this should happen prior to this step?

src/main/java/org/javarosa/core/model/FormDef.java
shubham1g5: I was trying to avoid the repetition on line below  `throw new RemoteInstanceFetcher.RemoteInstanceException(errorMessage, e.getCause());` but open to either styles if this doesn't look neat.

snopoke:
The approach seems cleaner to me and encapsulates the data needed to re-create the instance after deserialzation. I am a little confused by all the different `instanceId` variables floating around though.

orangejenny:
@ctsims Is your feedback addressed? The rest of this functionality is ready to merge.

ctsims (approved):
No more feedback from my side.

Do we have a clear sense of what serializers this breaks and what the migration path will be? Trying to figure out if there's going to be a complex path for getting this into the android build

orangejenny:
Thanks, @ctsims 

Merging so we can get the cross-jurisdictional features out, leaving on followup on the serializers to @shubham1g5 / @snopoke

PR #1039, opened 2021-09-10
Applied changes from PR#1036 to master branch

orangejenny:
This should have no user-facing changes. It adjusts the fixes made in https://github.com/dimagi/commcare-core/pull/1019 so that they don't break formplayer.

PR #1041, opened 2021-09-16
Remove unreachable return statement 

ShivamPokhriyal:
The build is failing on jenkins with this error: 
```
16:06:44 /commcare-core/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java:252: error: unreachable statement
16:06:44         return false;
```
This PR fixes above build issue.

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
@@ -249,7 +249,6 @@ public class EntityListSubscreen extends Subscreen<EntityScreen> {
             host.setCurrentScreenToDetail();
             return true;
         }
-        return false;
     }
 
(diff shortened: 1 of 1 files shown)


orangejenny (approved):
Thank you for fixing - any idea why this didn't break the build on https://github.com/dimagi/commcare-core/pull/1039/?

ShivamPokhriyal:
> Thank you for fixing - any idea why this didn't break the build on #1039?

I've been wondering the same. That build seemed to have run `:check` gradle task which should've thrown the error, but not sure why it didn't. When I tried running `./gradlew :commcare-core:check` it gave the error successfully. 

It could perhaps be possible that I was using gradle wrapper to run these tasks and jenkins is using gradle(not sure which version). I think I can probably try running the command on the jenkins machine to get more info.

PR #1046, opened 2021-09-21
Smart linking in web apps

orangejenny:
https://dimagi-dev.atlassian.net/browse/USH-1292

See https://github.com/dimagi/formplayer/pull/1010 for background. This PR just adds command id to endpoint definitions, so that formplayer can identify which endpoint goes with the screen it is currently on (or about to navigate to).


orangejenny:
This isn't necessary anymore, closing out.

PR #1048, opened 2021-10-11
Smart linking in web apps: Stack queries

orangejenny:
Needed for https://github.com/dimagi/formplayer/pull/1014 but can be merged independently. This adds parsing for stack queries and has logic to turn a query into a `StackFrameStep`.

Also PRed against master: https://github.com/dimagi/commcare-core/pull/1056

Review full diff at once.

### DIFF
--- a/src/main/java/org/commcare/xml/StackFrameStepParser.java
+++ b/src/main/java/org/commcare/xml/StackFrameStepParser.java
@@ -10,6 +10,8 @@ import org.kxml2.io.KXmlParser;
 import org.xmlpull.v1.XmlPullParserException;
 
 import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
 
 /**
  * @author ctsims
@@ -38,6 +40,13 @@ class StackFrameStepParser extends ElementParser<StackFrameStep> {
             case "query":
                 String queryId = parser.getAttributeValue(null, "id");
                 String url = parser.getAttributeValue(null, "value");
+                try {
+                    new URL(url);
+                } catch (MalformedURLException e) {
+                    String errorMsg = "<query> element has invalid 'value' attribute (" + url + ").";
+                    throw new InvalidStructureException(errorMsg, parser);
+                }
+
                 StackFrameStep step = null;
                 try {
                     step = new StackFrameStep(SessionFrame.STATE_QUERY_REQUEST, queryId, url, false);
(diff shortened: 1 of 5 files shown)


src/main/java/org/commcare/xml/StackFrameStepParser.java
```diff
@@ -22,17 +23,41 @@
...
+                return parseValue(SessionFrame.STATE_UNKNOWN, datumId);
             case "rewind":
                 return parseValue(SessionFrame.STATE_REWIND, null);
             case "mark":
                 return parseValue(SessionFrame.STATE_MARK, null);
             case "command":
                 return parseValue(SessionFrame.STATE_COMMAND_ID, null);
+            case "query":
+                String queryId = parser.getAttributeValue(null, "id");
+                String url = parser.getAttributeValue(null, "value");
```
shubham1g5, line 40: we should probably do some validation like [here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/xml/SessionDatumParser.java#L98) on url

shubham1g5 (approved):
Nothing blocking but probably be good to add some tests around this probably[here](https://github.com/dimagi/commcare-core/blob/master/src/test/java/org/commcare/backend/session/test/SessionStackTests.java) just to test the parsing.

snopoke (commented):
+1 on @shubham1g5 comments. Also this should get a docs update.

ctsims (approved):
Agreed with Simon that we should make sure that docs are updated.

Specifically, since this is very similar to another construct elsewhere in the app, but with slightly different semantics, it would be useful for someone to be able to figure out what expectations to have about the query lifecycle here (might have more feedback about that in the formplayer half of the PR.

src/main/java/org/commcare/xml/StackFrameStepParser.java
orangejenny: Simon addressed in https://github.com/dimagi/commcare-core/pull/1048/commits/d1ea42af5e893cb33a097583fd40c21b89027ade

orangejenny:
@shubham1g5 What do you think of https://github.com/dimagi/commcare-core/pull/1048/commits/4ba1de63f0e53347257ab0e95293e1534dba8ef0 for tests? It doesn't really assert anything, but it it to execute the parsing code.

snopoke (approved):

shubham1g5 (approved):
yes, looks great!

orangejenny:
Updated docs: https://github.com/dimagi/commcare-core/wiki/SessionStack#jump

PR #1049, opened 2021-10-13
Suggested refactor for Query Datum support

ctsims:
(not intended to be pulled, provided for illustration)

Builds on 
https://github.com/dimagi/commcare-core/pull/1040

but moves instantiating the external instances to a decoupled pre-condition rather than an update to initialization

would still need to be updated for clarity (and hasn't been tested) and conciseness. Also has the weakness that this approach currently isn't capable of **only** ensuring that instances which are referenced by the form are made available during form entry.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -1406,17 +1406,17 @@ public class FormDef implements IFormElement, IMetaData,
      *                            (presumably in HQ) - so don't fire end of form event.
      */
     public void initialize(boolean newInstance, boolean isCompletedInstance,
-                           InstanceInitializationFactory factory) throws RemoteInstanceFetcher.RemoteInstanceException {
-        initialize(newInstance, isCompletedInstance, factory, null, false, null);
+                           InstanceInitializationFactory factory) {
+        initialize(newInstance, isCompletedInstance, factory, null, false);
     }
 
-    public void initialize(boolean newInstance, InstanceInitializationFactory factory) throws RemoteInstanceFetcher.RemoteInstanceException {
-        initialize(newInstance, false, factory, null, false, null);
+    public void initialize(boolean newInstance, InstanceInitializationFactory factory) {
+        initialize(newInstance, false, factory, null, false);
     }
 
     public void initialize(boolean newInstance, InstanceInitializationFactory factory, String locale,
-                           boolean isReadOnly, RemoteInstanceFetcher remoteInstanceFetcher) throws RemoteInstanceFetcher.RemoteInstanceException {
-        initialize(newInstance, false, factory, locale, isReadOnly, remoteInstanceFetcher);
+                           boolean isReadOnly) {
+        initialize(newInstance, false, factory, locale, isReadOnly);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java
```diff
@@ -50,39 +53,36 @@ public ExternalDataInstance(ExternalDataInstance instance) {
...
-        this.root = instance.getRoot();
+        //Copy constructor avoids check.
+        this.root = instance.root;
         this.mCacheHost = instance.getCacheHost();
-        this.remoteUrl = instance.getRemoteUrl();
+        this.source = instance.getSource();
         useCaseTemplate = CaseInstanceTreeElement.MODEL_NAME.equals(instanceid);
     }
 
     private ExternalDataInstance(String reference, String instanceId,
```
snopoke, line 63: should `instanceId` also come from `source`?

snopoke: There are a few of places where `instanceId` get's passed as arguments on methods for objects that have an `instanceId` field which I find confusing. e.g. `ExternalDataInstance.initialize`

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1455,41 +1450,6 @@ public void initialize(boolean newInstance, boolean isCompletedInstance, Instanc
...
-                            new URI(externalDataInstance.getRemoteUrl()));
-                    if (newExternalDataInstance != null && newExternalDataInstance.getRoot() != null) {
-                        replacedInstances.add(newExternalDataInstance);
-                    }
-                } catch (UnfullfilledRequirementsException | XmlPullParserException |
-                        InvalidStructureException | IOException | URISyntaxException e) {
-                    String errorMessage;
-                    if (e instanceof URISyntaxException) {
-                        errorMessage = "Invalid url found for the remote instance " + instance.getName() + ".";
-                    } else if (e instanceof IOException) {
```
shubham1g5, line 1479: Should we retain this error handling elsewhere aka try to give different error messages on invalid data vs a network error ?

src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceSource.java
```diff
@@ -0,0 +1,71 @@
...
+
+    protected TreeElement getRoot() {
+        if (needsInit()) {
+            throw new RuntimeException("Uninstantiated external instance source");
+        }
+        return root;
+    }
+
+    public void init(TreeElement root) {
+        //TODO: One-time init
```
shubham1g5, line 52: is there still something pending here in terms of initialisation ?

src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java
```diff
@@ -101,6 +101,11 @@ public InstanceBase getBase() {
 
     @Override
     public AbstractTreeElement getRoot() {
+        if (needsInit()) {
+            throw new RuntimeException("Attempt to use instance " + instanceid + " without inititalization.");
+        }
+        //TODO: We are just presuming this is lined up, but we should probably be pulling this
```
shubham1g5, line 107: agree with this.

src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java
ctsims: I 100% agree about `instanceId`. I didn't get a chance to go back and refactor around, but I think I accidentally added at least one new unnecessary instance of it, and its role is generally quite obtuse. 

One thing Shubham and I chatted about this morning was refactoring more substantially to replace the use of ExternalDataInstance with the new `Source` object in the `StackFrameStep` itself, which I *think* would fix a lot of the callstack ordering confusion here.

PR #1052, opened 2021-10-22
preserve value of useCaseTemplate when initializing instances

snopoke:
For instances inside forms there is no way to declare that the instance uses a template. This will preserve the value of the instance from the session.

Using the 'case' template allows referencing case properties that don't exist in the case without causing errors.

PR into https://github.com/dimagi/commcare-core/pull/1040

### DIFF
--- a/src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
+++ b/src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
@@ -71,7 +71,7 @@ public class CommCareInstanceInitializer extends InstanceInitializationFactory {
     }
 
     @Override
-    public AbstractTreeElement generateRoot(ExternalDataInstance instance) {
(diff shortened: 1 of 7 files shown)


ctsims:
Hey Simon,

This is a pretty deep change, so I'm wondering if there's a way to restructure in a lighter way.

I'm not clear I understood the need to add the new `InstanceRoot` concept. Is it not sufficient to just provide the External Instance with the serialization of its initialized 'use case instance' property?

snopoke:
>  Is it not sufficient to just provide the External Instance with the serialization of its initialized 'use case instance' property?

The issue is with instances that are initialised in forms where you can't specify the template. The only way to get the 'use case instance' value it to copy it from the instance in the session. 

An alternative approach would be to add an attribute to the instance element in the form.

snopoke:
closing in favour of https://github.com/dimagi/commcare-core/pull/1053

PR #1053, opened 2021-11-02
Case Template persistence refactor

ctsims:
Proposed / illustration of alternative to

https://github.com/dimagi/commcare-core/pull/1052

### Context from #1052:

For instances inside forms there is no way to declare that the instance uses a template. This will preserve the value of the instance from the session.

Using the 'case' template allows referencing case properties that don't exist in the case without causing errors.

PR into #1040

@snopoke Still kind of convoluted, but allows compartmentalizing a tiny bit more. I'm not convinced this approach is better, but figured I'd share the code. I'm happy to take whichever path you'd prefer.

cross-request: https://github.com/dimagi/formplayer/pull/1024


snopoke:
> Still kind of convoluted, but allows compartmentalizing a tiny bit more. I'm not convinced this approach is better, but figured I'd share the code. I'm happy to take whichever path you'd prefer.

Looks good to me. I've added a few more commits to clean things up.

src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java
```diff
@@ -139,8 +139,21 @@ public void writeExternal(DataOutputStream out) throws IOException {
     @Override
     public DataInstance initialize(InstanceInitializationFactory initializer, String instanceId) {
         base = new InstanceBase(instanceId);
-        root = initializer.generateRoot(this);
-        base.setChild(root);
+        InstanceRoot instanceRoot = initializer.generateRoot(this);
+        // this indirectly calls `this.copyFromSource` via the InstanceRoot so that we call the
+        // correct method based on the type
+        instanceRoot.setupNewCopy(this);
```
shubham1g5, line 145: I am not able to grasp how the `InstanceRoot` is helping persisiting `useCaseTemplate` .  `setupNewCopy` only seems to be called from here and should result in executing the block - 

````
 root = instanceRoot.getRoot();
 base.setChild(root);
````

How is it then the `useCaseTemplate` getting copied here ?

**Review by shubham1g5 — COMMENTED** [review]
Is it possible to add tests on this that can assert for the value of `useCaseTemplate` for form instances ?

File: src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java (hunk 1 above)
Comment (snopoke) [reply src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java:145]: If the `instanceRoot` is an `ExternalDataInstanceSource` then it will call also copy the `source` value:

```
     public void copyFromSource(ExternalDataInstanceSource source) {
        //parent copy
        copyFromSource((InstanceRoot)source);
        this.source = source;
    }
```
https://github.com/dimagi/commcare-core/pull/1053/files/095d42695f3ecab008366af0294ca10b3a6ed084#diff-d872b3df15633847b778eaa24e6797f9a6339f68c9981ac04ed16d9eb5ade1c4R154-R158

snopoke:
> Is it possible to add tests on this that can assert for the value of `useCaseTemplate` for form instances ?

👍🏻

snopoke:
> > Is it possible to add tests on this that can assert for the value of `useCaseTemplate` for form instances ?
> 
> 👍🏻

2594214c

src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java
shubham1g5: this makes sense, thanks!

shubham1g5 (approved):

PR #1061, opened 2021-12-13
support multiple extras with the same key in stack frames

snopoke:
This converts `StackFrameStep.extras` from a single value map to a multi-map allowing it to contain multiple entries with the same key. This is needed for instances where the stack query contains duplicate data keys:

```xml
<query url="http://localhost:8000/a/skelly-1/phone/registry_case/3dc223faa388dc16b8441194c001cec2/" storage-instance="registry" template="case" default_search="true">
  <data key="case_type" ref="'patient'"/>
  <data key="case_type" ref="'contact'"/>
  <data key="case_id" ref="instance('commcaresession')/session/data/case_id"/>
  <data key="case_id" ref="instance('casedb')/casedb/case[@case_id = instance('commcaresession')/session/data/case_id]/potential_duplicate_id"/>
</query>
```

This will break all serialised sessions.

### DIFF
--- a/src/main/java/org/commcare/session/CommCareSession.java
+++ b/src/main/java/org/commcare/session/CommCareSession.java
@@ -1,5 +1,6 @@
 package org.commcare.session;
 
+import com.google.common.collect.Multimap;
 import org.commcare.suite.model.ComputedDatum;
 import org.commcare.suite.model.Detail;
 import org.commcare.suite.model.EntityDatum;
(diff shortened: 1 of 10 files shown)


orangejenny (approved):
Do you think it's worth also updating `testSettingAndGettingSessionExtras`? The new test seems fine for coverage, it just looks like the existing test is currently an overview of extra behavior, might be nice for it to include a multiple key example.

snopoke:
> testSettingAndGettingSessionExtras

3cd7dbbb3

orangejenny (approved):
thanks!

PR #1062, opened 2022-01-04
Replaces use of internal `sun.*` APIs

shubham1g5:
I am no longer able to resolve `sun.misc.IOUtils.readAllBytes` on my machine. So replacing it with an alternate implementation of ours.


ctsims (approved):
Probably a JDK thing (Open JDK v. Sun JDK). Makes sense to use a consistent method, thanks!

PR #1063, opened 2022-01-06
Added tracing

orangejenny:
See https://github.com/dimagi/formplayer/pull/1076

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,9 @@ dependencies {
     compile 'com.google.code.findbugs:jsr305:3.0.2'
     compile 'io.reactivex.rxjava2:rxjava:2.1.1'
     compile 'com.google.guava:guava:30.1.1-jre'
+    compile 'io.opentracing:opentracing-api:0.33.0'
+    compile 'io.opentracing:opentracing-util:0.33.0'
+    compile 'com.datadoghq:dd-trace-api:0.64.0'
 
     cliCompile 'org.json:json:20140107'
     cliCompile group: 'regexp', name: 'regexp', version: '1.3'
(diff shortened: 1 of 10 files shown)


shubham1g5 (approved):

ctsims:
@orangejenny does this creative any sort of passive tracing on these methods without some sort of top level span setting?

A bunch of the areas traces have been added are are in the most performance-sensitive parts of the code (expandEntityReferenceSet | initAllTriggerables | answerQuestion) which makes sense since that's presumably why you want to trace them. I just wanted to confirm whether it adds overhead without some other trigger, since that would cause pretty widespread impact.

orangejenny:
@ctsims

The datadog tracing shouldn't add any overhead - it's controlled by the usage of `dd-java-agent.jar` ([docs](https://docs.datadoghq.com/tracing/setup_overview/setup/java/?tab=containers#trace-annotation)), which we only use on staging.

For OpenTracing, I didn't find a built-in way to disable the tracing globally, so fcdceca3c81eb4be9b529e45cf9e86b81a59a807 adds a system property which I can then add to [formplayer_command_args](https://github.com/dimagi/commcare-cloud/blob/fae1d030bc553c1e3beb15d28716262937fd055b/environments/staging/app-processes.yml#L1).

ctsims:
@orangejenny  Great, that change looks like a good way to control it to me.

orangejenny:
Cool, added https://github.com/dimagi/commcare-cloud/pull/5096 to enable the new system property to staging.

ctsims:
Last question (sorry) - have we tested the impact of the tracing on some high volume apps? I think the AE team has some good performance baseline testbeds they can use to compare differences (although I'm not sure if they are functional on staging). 

I ask because these traces will get *deep* very quickly and I'm guessing even on the formplayer side there will be a pretty meaningful performance implication.

orangejenny:
@ctsims Fair question, but we've had this code on staging pretty consistently for over a year. USH is likely the biggest user of high-volume apps on staging, and they're pretty invested in having the tracing code running.

This PR ought to make it easier to turn off the tracing code, since previously it was on by default and required a code deploy to turn off. Now it'll only take a settings update.

ctsims:
ah, got it. That does seem really helpful, it would be interesting at some point to try something like splitting the webworkers into having the property on/off and seeing if there's a meaningful change, which'll be much easier with things controlled through this mechanism

PR #1065, opened 2022-01-26
Gradle 7 api changes for Formplayer

shubham1g5:
https://github.com/dimagi/formplayer/pull/1086

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,7 +73,6 @@ dependencies {
     // Added so Android Studio recognizes libs in util jar projects
     implementation 'org.json:json:20140107'
     implementation 'commons-cli:commons-cli:1.3.1'
-    implementation 'xpp3:xpp3:1.1.4c'
     implementation 'joda-time:joda-time:2.9.4'
     implementation 'com.carrotsearch:hppc:0.7.2'
     api 'com.squareup.retrofit2:retrofit:2.3.0'
@@ -89,7 +88,6 @@ dependencies {
     cliImplementation group: 'regexp', name: 'regexp', version: '1.3'
     cliImplementation 'com.github.stefanhaustein:kxml2:2.4.1'
     cliImplementation 'commons-cli:commons-cli:1.3.1'
-    cliImplementation 'xpp3:xpp3:1.1.4c'
     cliImplementation 'joda-time:joda-time:2.9.4'
     cliImplementation 'io.reactivex.rxjava2:rxjava:2.1.1'
     cliImplementation 'com.squareup.retrofit2:retrofit:2.3.0'
@@ -166,7 +164,10 @@ task cliJar(type: Jar, dependsOn: cliClasses) {
     manifest {
         attributes 'Main-Class': 'org.commcare.util.cli.CliMain'
     }
-    duplicatesStrategy = DuplicatesStrategy.INCLUDE
+    duplicatesStrategy = DuplicatesStrategy.FAIL
+    filesMatching(["**/LICENSE.txt","**/NOTICE.txt"]) {
+        duplicatesStrategy = DuplicatesStrategy.INCLUDE
+    }
 }
 
 task ccapiJar(type: Jar, dependsOn: ccapiClasses) {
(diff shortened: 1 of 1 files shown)


shubham1g5:
Updated Gradle version on core PR builder. 

shubham1g5:
changed `api` -> `ccapi` in Jenkins PR builder. 

build.gradle
```diff
@@ -146,62 +153,47 @@ jar {
...
     from {
-        configurations.apiCompile.collect {
-            it.isDirectory() ? it : zipTree(it)
-        }
-        configurations.apiRuntime.collect {
+        configurations.ccapi.collect {
             it.isDirectory() ? it : zipTree(it)
         }
     }
+    duplicatesStrategy = DuplicatesStrategy.INCLUDE
```
snopoke, line 181: What's the implication of doing this? Do you know which files are duplicated?

shubham1g5: All of these duplicates are caused by meta data files present in `META-INF` folder inside a jar ([example](https://github.com/dimagi/commcare-android/blob/master/app/build.gradle#L249-L255)). `INCLUDE` seems more flexible to me since it allows to keep duplicate files if possible otherwise override the files. `EXCLUDE` on other hand doesn't allow override but simply skips any duplicates. So it's a matter of choosing the first one vs last one in our case and don't think it doesn't really matter for these meta files. Probably `WARN` Is the best one to use since it behaves as `INCLUDE` but also logs a warning ?

snopoke: My concern is that if there are functional files like `META-INF/services/java.sql.Driver` then overwriting them could be problematic.

Is it possible to create an allow list so that the default is to fail except for files we allow. e.g.

```
duplicatesStrategy = DuplicatesStrategy.FAIL
filesMatching(["**/LICENSE*", "**/NOTICE*", "**/DEPENDENCIES*"]) {
    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
}
```

shubham1g5: Your concern is valid though I do not think there is any option to provide exceptions to the `duplicateStrategy` in [gradle dsl](https://docs.gradle.org/current/dsl/org.gradle.jvm.tasks.Jar.html#N22C03). Adding `INCLUDE` shouldn't result in any changes since previosuly without explicity specifying the duplicate strategy, it  would be defaulting to `[INHERIT](https://docs.gradle.org/current/javadoc/org/gradle/api/file/DuplicatesStrategy.html#INHERIT)` which should have been resolving to `INCLUDE` strategy as well.

snopoke: I tested the approach I suggested and it does work (note the logic below is reversed to test that it's applying the rule to the matching files only):
```
task cliJar(type: Jar, dependsOn: cliClasses) {
    ....
    duplicatesStrategy = DuplicatesStrategy.INCLUDE
    filesMatching("**/LICENSE.txt") {
        duplicatesStrategy = DuplicatesStrategy.FAIL
    }
}
```

Running `./gradlew commcare:cliJar --stacktrace` fails with the error: `Encountered duplicate path "META-INF/LICENSE.txt" during copy operation configured with DuplicatesStrategy.FAIL`.

I also tested with WARN to get a list of duplicates and came up with the following list in the CLI Jar:
```
META-INF/services/org.xmlpull.v1.XmlPullParserFactory
org/xmlpull/v1/XmlPullParser.class
org/xmlpull/v1/XmlPullParserException.class
org/xmlpull/v1/XmlPullParserFactory.class
org/xmlpull/v1/XmlSerializer.class
```

If you say the previous strategy was INCLUDE them I'm fine to go ahead as is in this PR but I would like to re-visit this later.

snopoke (approved):

build.gradle
shubham1g5: Thanks a bunch. I didn't realise we could provide config in `filesMatching`. I made some changes in https://github.com/dimagi/commcare-core/pull/1065/commits/db786b0278972ab37d39b845f86bc45dd283a1cb. It turned out only the `translate` really needs the `xmlpull` dependency. I went down a rabbit hole in order to identify from where `xmlpull` classes are duplicating but could not identify the cause of it after surveying all dependencies. Even when I remove the `xpp3` dependency from other configs, the `xmlpull` classes are present in the jar. So it definitely seems like some other dependency is bringing these classes. I individually tried adding `exclude module 'xmlpull`` to each of the dependency but that didn't solve the issue as well. So in the end I am just specifying these as exceptions in duplicate strategy as well.

PR #1070, opened 2022-02-08
Removed unused private var from QueryScreen

orangejenny:


snopoke (approved):

esoergel (approved):

shubham1g5:
FYI that  PRs to `master` currently fails in building due to gradle version mismatch between FP and Mobile. We can only point the Jenkins PR builder to one specific Gradle version as of now and it's configured for Gradle 7 now aka for FP PRs. We can only update mobile Gradle build along with Java 11 upgrade on mobile (unlike FP where we just did gradle update without Java). But mobile Java 11 upgrade needs to happen along with FP. TLDR: the PR builds for `master` will fail until we merge Java 11 upgrade on FP.

shubham1g5:
master builds are now resolved. Merging.

orangejenny:
thanks!

PR #1073, opened 2022-03-02
Added case search user inputs to EC when evaling search filters

orangejenny:
https://dimagi-dev.atlassian.net/browse/USH-1732


shubham1g5 (commented):
code looks good to me, though is it possible to add a test case either in core or FP to display the required use case around dynamic filtering.

snopoke (approved):
+1 for a test

orangejenny:
Closing out since we're going in a different direction (virtual instance containing user input).

PR #1074, opened 2022-03-16
make case search user input available to default filters

snopoke:
Jira: https://dimagi-dev.atlassian.net/browse/USH-1749

These changes make the user input from case search available to the xpath expressions contained in the default filters.

The user input is made available via the `search-input` instance: `instance('search-input')/input/field[@name='patient_id']`

**Contrived example**
The `_xpath_query` data element references the value from the `patient_id` prompt:

```
<query url="https://www.fake.com/patient_search/" storage-instance="patients">
  <data key="_xpath_query" ref="if(count(instance('search-input')/input/field[@name='patient_id']) > 0, concat('external_id = ', instance('search-input')/input/field[@name='patient_id']), '')"/>
  <prompt key="patient_id">
    <display>
      <text>
        <locale id="query.name"/>
      </text>
    </display>
  </prompt>
</query>
```

* If the value of `patient_id` is "123" the XPath evaluates to `external_id = 123`
* If the value of `patient_id` is `null` the XPath evaluates to and empty string

**Notes for reviewer**
In addition to normal review I'm particularly interested in how I've set up the new `VirtualDataInstance` class. I'm also interested in figuring out how we could make this instance available in the case list / detail screen that results from the search (See https://dimagi-dev.atlassian.net/browse/USH-1754).

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -225,7 +225,7 @@ public class QueryScreen extends Screen {
     }
 
     public void refreshItemSetChoices() {
-        remoteQuerySessionManager.refreshItemSetChoices(remoteQuerySessionManager.getUserAnswers());
+        remoteQuerySessionManager.refreshItemSetChoices();
     }
 
     public URL getBaseUrl() {
--- a/src/main/java/org/commcare/data/xml/SimpleNode.java
+++ b/src/main/java/org/commcare/data/xml/SimpleNode.java
@@ -1,0 +1,57 @@
+package org.commcare.data.xml;
+
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Metadata class used in conjunction with {@link TreeBuilder} to allow creating simple
+ * {@link org.javarosa.core.model.instance.TreeElement} trees.
+ *
+ * SimpleNode.parentNode("node", ImmutableMap.of("attr1", "v1"), ImmutableList.of(
+ *     SimpleNode.textNode("child", Collections.emptyMap(), "some text"),
+ *     SimpleNode.textNode("child", Collections.emptyMap(), "other text")
+ * ))
+ */
+public class SimpleNode {
+    private String name;
+    private Map<String, String> attributes;
+    private List<SimpleNode> children;
+    private String text;
+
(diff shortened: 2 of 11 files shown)


orangejenny (commented):
This is awesome.

I don't have much insight into making this available to the case list. The first thing that comes to mind is caching, but it would have to be 100% reliable.

`VirtualDataInstance` itself makes sense to me. I'm a little surprised by `VirtualInstances` - I'd probably have made the search instance a subclass of `VirtualDataInstance` instead, and given `VirtualDataInstance` a build method. But I don't have an argument for that being better.

What would you think of having type information in this instance? I'm assuming date ranges will show up in this with delimiters, like `2022-03-05__range__2022-03-07` and `1#,#2#,#3#,#`. Although app builders already need to know about these if they use default values for these field types, so this PR doesn't expose anything that isn't already exposed.

shubham1g5 (approved):
Looks great to me! 

Agree with Jenny that we would need to store the instance locally to make it available outside the search screen. That probably means we would like to define a [reference](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java#L28) as well for the VirtualDataInstance that can be used to reinitialize it later on.

snopoke:
> Agree with Jenny that we would need to store the instance locally to make it available outside the search screen. That probably means we would like to define a [reference](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java#L28) as well for the VirtualDataInstance that can be used to reinitialize it later on.

I'm going to leave that for a future PR.

snopoke:
@shubham1g5 I was going to PR this into `master` as well but there are a number of changes in `RemoteQuerySessionManager` which I'm unsure about.

shubham1g5:
@snopoke Any change in particular ? This looks quite safe to me to merge into master as we are only adding another virtual instance to the eval context which is unlikely to interfere with any exisiting xpath evaluations. I would support opening up a PR and I can test that Android integration tests passes with those changes before merging.

snopoke:
> @snopoke Any change in particular ? This looks quite safe to me to merge into master as we are only adding another virtual instance to the eval context which is unlikely to interfere with any exisiting xpath evaluations. I would support opening up a PR and I can test that Android integration tests passes with those changes before merging.

Will do

PR #1075, opened 2022-03-18
Bump max line length to 115

shubham1g5:


orangejenny (approved):

millerdev:
The HQ [max line length is 115](https://github.com/dimagi/commcare-hq/blob/master/.flake8#L2). It seems like there would be value in standardizing so developers who switch between the two have a more consistent experience.

PR #1076, opened 2022-03-22
include null inputs when populating itemset choices

snopoke:
Followup to https://github.com/dimagi/commcare-core/pull/1074

I discovered failing tests when updating the commcare-core ref in `formplayer`.

I've added relevant tests to commcare-core and fixed the bug.

### DIFF
--- a/src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java
+++ b/src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java
@@ -55,7 +55,8 @@ public class CaseClaimModelTests {
         testPopulateItemsetChoices(Collections.emptyMap(), Collections.emptyList());
     }
 
-    private void testPopulateItemsetChoices(Map<String, String> userInput, List<String> expected) throws Exception {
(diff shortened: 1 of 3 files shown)


src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java
```diff
@@ -35,6 +45,43 @@ public void testRemoteQueryDatum() throws Exception {
...
+
+    private void testPopulateItemsetChoices(Map<String, String> userInput, List<String> expected)
+            throws Exception {
+        MockApp mApp = new MockApp("/case_claim_example/");
+
+        SessionWrapper session = mApp.getSession();
+        session.setCommand("patient-search");
+
+        VirtualDataInstance districtInstance = buildDistrictInstance();
+        EvaluationContext context = session.getEvaluationContext().spawnWithCleanLifecycle(ImmutableMap.of(
```
shubham1g5, line 66: Quite cool that we can now inject data instances without relying on loading them through restore.

shubham1g5 (approved):

PR #1079, opened 2022-03-28
FormInstance should have formVersion set from the version attribute

gherceg:
When serialiizing/deserializing a FormDef, we use the writeExternal and readExternal methods on a FormDef, which eventually calls down to the readExternal and writeExternal methods on a FormInstance.

The form version was just not being read or write at this point so this change ensures that happens.


calellowitz (approved):

src/main/java/org/javarosa/core/model/instance/FormInstance.java
```diff
@@ -175,6 +175,7 @@ public FormInstance clone() {
     public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
         super.readExternal(in, pf);
         schema = (String)ExtUtil.read(in, new ExtWrapNullable(String.class), pf);
+        formVersion = (String)ExtUtil.read(in, new ExtWrapNullable(String.class), pf);
```
shubham1g5, line 178: can we also do the same for `uiVersion` as that also seems to be missing from here.

gherceg:
I'm realizing this change is very naïve since we are reading in/writing from a data stream. If I look at a typical <data> field, I see: `<data xmlns:jrm="http://dev.commcarehq.org/jr/xforms" xmlns="http://openrosa.org/formdesigner/5BE25716-192B-4554-BD2A-3409ECFFACF1" uiVersion="1" version="7" name="Followup Form">`. My assumption would be after the schema is read in, read in uiVersion, then version. I'm not sure where date saved fits into that.

shubham1g5 (approved):
The order in xml doesn't matter as long as we read attributes in `readExternal` in the same order as we wrote them in `writeExternal`

gherceg:
Thanks @shubham1g5 that clears it up 👍🏻

PR #1080, opened 2022-03-28
simplify list query data

snopoke:
The nested `list` element is unnecessary and makes parsing more complicated. Adding the `exclude` attribute on the `data` element also allows us to use it for other purposes.

This also adds support for the `excludes` attribute on value data elements:
```xml
<data key="external_id" ref="'abc'" exclude="instance('casedb')/casedb/case[@case_type='commcare-user'][hq_user_id=instance('commcaresession')/session/context/userid]/role = 'admin'" />
```

### DIFF
--- a/src/test/java/org/javarosa/test_utils/ReflectionUtils.java
+++ b/src/test/java/org/javarosa/test_utils/ReflectionUtils.java
@@ -6,7 +6,7 @@ import java.lang.reflect.Modifier;
 /**
  * ReflectionUtils is a collection of reflection-based utility methods for use in
  * unit testing scenarios.
- *
+ * <p>
  * Code adapter from org.springframework.test.util.ReflectionTestUtils
  */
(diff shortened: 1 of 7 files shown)


src/test/java/org/javarosa/test_utils/ReflectionUtils.java
```diff
@@ -0,0 +1,62 @@
+package org.javarosa.test_utils;
```
shubham1g5, line 1: you can probaby just do `org.javarosa.utils` as a package.

snopoke: Since that package already exists I think it makes sense to use it

shubham1g5: ahh didn't realise it already exists. makes sense.

PR #1082, opened 2022-03-30
Session navigation for Multi Select Case List 

shubham1g5:
- Adds support for [Multi Select Case Lists](https://github.com/dimagi/formplayer/wiki/Multi-Select-Case-Lists)

- Handles initialization for [Virtual Instances](https://github.com/dimagi/commcare-core/wiki/Virtual-Data-Instances)

https://github.com/dimagi/formplayer/pull/1109


src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -149,7 +149,7 @@ public void advanceSessionWithEndpoint(String endpointId, String[] endpointArgs)
             if (s instanceof SyncScreen) {
                 try {
                     s.init(mSession);
-                    s.handleInputAndUpdateSession(mSession, "", false);
+                    s.handleInputAndUpdateSession(mSession, "", false, null);
```
shubham1g5, line 152: I am not very faimilar with CLI at this point, so I have parked a ticket instead - https://dimagi-dev.atlassian.net/browse/USH-1838 to get back to.

src/cli/java/org/commcare/util/screen/EntityScreen.java
```diff
@@ -192,8 +192,13 @@ public Subscreen getCurrentScreen() {
         return mCurrentScreen;
     }
 
+    protected String getReturnValueFromSelection(TreeReference contextRef) {
+        return getReturnValueFromSelection(contextRef, mNeededDatum, evalContext);
```
orangejenny, line 194: Pre-existing nitpick, but I wish this class were consistent about using the `m` prefix for class attributes.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,84 @@
...
+            EntitiesSelectionCache entitiesSelectionCache)
+            throws CommCareSessionException, SQLException {
+        super(handleCaseIndex, full, session);
+        this.entitiesSelectionCache = entitiesSelectionCache;
+        processSelectedValues(selectedValues, session);
+    }
+
+    private void processSelectedValues(String[] selectedValues, SessionWrapper session)
+            throws SQLException, CommCareSessionException {
+        if (selectedValues != null) {
```
orangejenny, line 56: So if all values are to be unselected, this function would get an empty array passed in?

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,84 @@
...
+
+            String[] evaluatedValues = new String[selectedValues.length];
+            for (int i = 0; i < selectedValues.length; i++) {
+                TreeReference currentReference = getEntityReference(selectedValues[i]);
+                if (currentReference == null) {
```
orangejenny, line 60: Would it be worth gathering up all erroring entities instead of stopping on the first one?

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Override
+    protected void setSession(SessionWrapper session) throws CommCareSessionException {
+        super.setSession(session);
+        maxSelectValue = ((MultiSelectEntityDatum)mNeededDatum).getMaxSelectValue();
+    }
+
+    @Override
+    public void setHighlightedEntity(String id) throws CommCareSessionException {
+        String[] cachedSelction = entitiesSelectionCache.read(id);
```
orangejenny: `cachedSelction` => `cachedSelection`

src/cli/java/org/commcare/util/screen/EntityScreen.java
```diff
@@ -127,7 +126,7 @@ public void init(SessionWrapper session) throws CommCareSessionException {
             evalContext.addFunctionHandler(new ScreenUtils.HereDummyFunc(-23.56, -46.66));
 
             if (mNeededDatum.isAutoSelectEnabled() && references.size() == 1) {
-                this.setHighlightedEntity(references.firstElement());
+                this.setSelectedEntity(references.firstElement());
```
orangejenny, line 129: Nice naming change.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public void setSelectedEntities(String input, String[] selectedValues)
+            throws CommCareSessionException {
+        if (input.contentEquals(USE_SELECTED_VALUES)) {
+            try {
+                processSelectedValues(selectedValues);
+            } catch (SQLException throwables) {
+                throw new CommCareSessionException(
+                        "An error occurred tryign to process selections on this screen. " +
```
orangejenny, line 40:
```suggestion
                        "An error occurred trying to process selections on this screen. " +
```

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -11,30 +11,49 @@
...
+                                " If this error persists please report a bug to CommCareHQ.", throwables);
+            }
+        } else {
+            String[] cachedSelction = entitiesSelectionCache.read(input);
```
orangejenny, line 44: `cachedSelction` => `cachedSelection`

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -16,7 +16,7 @@
  */
 public class MultiSelectEntityScreen extends EntityScreen {
 
-    private static final CharSequence USE_SELECTED_VALUES = "use_selected_values";
+    public static final CharSequence USE_SELECTED_VALUES = "use_selected_values";
```
orangejenny, line 19: Why does this need to be public? Looks like it's only referenced within this class.

orangejenny (approved):

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
shubham1g5: I am confused, can we navigate forward on Entity Screen with all values unselected ? I was supposing that can't happen.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
shubham1g5: I could not think of a compelling reason to do so, I think just the single entity should be sufficient in order to point out what's wrong to a developer. For users, Showing all those entities can be overwhelming and cluttery when there are more of those present.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,84 @@
...
+                                    " If this error persists please report a bug to CommCareHQ.");
+                }
+
+                evaluatedValues[i] = getReturnValueFromSelection(currentReference);
+            }
+
+
+            String guid = PropertyUtils.genGUID(10);
+            entitiesSelectionCache.cache(guid, evaluatedValues);
+            session.setDatum(STATE_MULTIPLE_DATUM_VAL, mNeededDatum.getDataId(), guid);
```
snopoke: shouldn't this happen in `updateSession`?

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
orangejenny: oh, no, you're right - I was narrowly focused on the code and not thinking about the context

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
orangejenny: sounds good

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,97 @@
...
+                if (currentReference == null) {
+                    throw new CommCareSessionException(
+                            "Could not select case " + selectedValues[i] + " on this screen. " +
+                                    " If this error persists please report a bug to CommCareHQ.");
+                }
+                evaluatedValues[i] = getReturnValueFromSelection(currentReference);
+            }
+
+            String guid = PropertyUtils.genGUID(10);
+            entitiesSelectionCache.cache(guid, evaluatedValues);
```
snopoke, line 69: when does this get cleared?

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
```diff
@@ -235,22 +238,31 @@ public boolean handleInputAndUpdateHost(String input, EntityScreen host, boolean
...
 
+    private void updateSelectedEntities(EntityScreen host, String input, String[] selectedValues)
+            throws CommCareSessionException {
+        if (host instanceof MultiSelectEntityScreen) {
+            ((MultiSelectEntityScreen)host).setSelectedEntities(input, selectedValues);
+        } else {
+            host.setSelectedEntity(input);
+            // Set entity screen to show detail and redraw
+            host.setCurrentScreenToDetail();
+        }
```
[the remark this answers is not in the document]
shubham1g5: I thought of that but didn't want to expose `selectedValues` in `EntityScreen` when it will only ever be used in `MultiSelectEntityScreen`.

src/cli/java/org/commcare/util/screen/EntityScreen.java
```diff
@@ -117,31 +115,35 @@ public void init(SessionWrapper session) throws CommCareSessionException {
 
         if (full || references.size() == 1) {
             referenceMap = new Hashtable<>();
-            EntityDatum needed = (EntityDatum) session.getNeededDatum();
-            for(TreeReference reference: references) {
-                referenceMap.put(getReturnValueFromSelection(reference, needed, evalContext), reference);
+            EntityDatum needed = (EntityDatum)session.getNeededDatum();
+            for (TreeReference reference : references) {
+                referenceMap.put(getReturnValueFromSelection(reference, needed, evalContext),
+                        reference);
```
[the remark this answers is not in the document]
shubham1g5: The line limit was bumped recenty from 100 to 115, so there are some inconsistencies in this PR regarding that.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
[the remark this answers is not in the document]
shubham1g5: No particular reason other than the variable was auto-created by IntelliJ based on it's usage in `String.contentEquals(CharSequence cs)`. Though I think using `String` and `CharSequence` is synonymous for this statement.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,84 @@
...
+            throws SQLException, CommCareSessionException {
+        if (selectedValues != null) {
+
+            String[] evaluatedValues = new String[selectedValues.length];
+            for (int i = 0; i < selectedValues.length; i++) {
+                TreeReference currentReference = getEntityReference(selectedValues[i]);
+                if (currentReference == null) {
+                    throw new CommCareSessionException(
+                            "Could not select case " + selectedValues[i] + " on this screen. " +
+                                    " If this error persists please report a bug to CommCareHQ.");
```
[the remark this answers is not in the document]
shubham1g5: yes.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
shubham1g5: Ahh good point, I didn't think about it. I will look into it now.

shubham1g5:
@snopoke Noted your point regrading formatting changes. I did do a PR separately to remove some of the formatting changes from [PR diff](https://github.com/dimagi/formplayer/pull/1110) for Formplayer. I can do more of that going forward. Curious if you think that be a good strategy ? That would not remove the formatting changes from individual commits but the PR diff should be more conscise. Also these changes should become insignificant as we have more code correctly formatted with time though do agree they are a pain for now.

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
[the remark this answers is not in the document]
snopoke: I think not having this logic in the sub screen would be of more benefit.

snopoke:
> @snopoke Noted your point regrading formatting changes. I did do a PR separately to remove some of the formatting changes from [PR diff](https://github.com/dimagi/formplayer/pull/1110) for Formplayer. I can do more of that going forward. Curious if you think that be a good strategy ? That would not remove the formatting changes from individual commits but the PR diff should be more conscise. Also these changes should become insignificant as we have more code correctly formatted with time though do agree they are a pain for now.

I don't mind if the formatting changes are in the same PR, just not the same commits as functional changes. Do you have autoformatting configured in your IDE? If so I would suggest turning off.

shubham1g5:
@snopoke Cool I will do that. I see there is an option in IntelliJ to reformat only the selected code. So I will switch to that instead. Thanks!

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+                String[] cachedSelection = entitiesSelectionCache.read(input);
+                if (cachedSelection == null) {
+                    throw new CommCareSessionException(
+                            "Could not make selection with reference id " + input + " on this screen. " +
+                                    " If this error persists please report a bug to CommCareHQ.");
+                }
+                storageReferenceId = input;
+            }
+        } catch (SQLException throwables) {
```
snopoke, line 61: is this exception still relevant with the move to PostgreSQL?

snopoke (commented):
Looking good

src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
```diff
@@ -89,10 +90,35 @@ public InstanceRoot generateRoot(ExternalDataInstance instance) {
...
+        throw new RuntimeException(
+                "Instances with reference to 'selected_cases' are not supported on this platform");
+    }
+
+    @Nullable
+    protected String getGuidForSelectedCasesInstance(ExternalDataInstance instance) {
+        for (StackFrameStep step : session.getFrame().getSteps()) {
+            if (step.getId().equals(instance.getInstanceId()) && step.getType().equals(
+                    SessionFrame.STATE_MULTIPLE_DATUM_VAL)) {
+                return step.getValue();
```
snopoke, line 116: I was just looking at this code again and realised we're taking a different approach with storing this instance than we are with the results instance from query datums (where we put the instance source directly into the step).

I'm curious if we couldn't use the same mechanism for this?

shubham1g5: Hmm Good point, I didn't think about it until now and think we could use the same mechanism without any downsides. Although it seems like the wiring around that is pretty specific to `ExternalDataInstance` and `ExternalDataInstanceSource`. So we probably will need to make `VirtualDataInstance` an implementation of `ExternalDataInstance` and make a new type of `ExternalDataInstanceSource` which is more DB based rather than HTTP based.

snopoke: Yes that's all true but I think it's better design for the long run. I don't think it's necessary to do immediately but I expect we'd need to then do it for the 'case search input' work.

shubham1g5: yeah that makes sense. Since I am on PTO soon I will hold it off for now and we can revisit it next sprint and see whether to include it with case search input or multi-select work.

src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -309,7 +309,7 @@ private boolean loopSession() throws IOException {
                         }
                     }
 
-                    screenIsRedrawing = !s.handleInputAndUpdateSession(mSession, input, false);
+                    screenIsRedrawing = !s.handleInputAndUpdateSession(mSession, input, false, null);
```
shubham1g5, line 312: @stephherbers this is the place where I anticipate changes for the multi-select referrals (You will need to work on top off this branch), I think we need to check here if we are on the `MultiSelectEntityScreen`. If yes, do following - 
 - Devise a strategy on how to take multiple cases as input from user from command prompts (probably ask user to enter all case ids at once in a tokenised string). Form a `selectedValues[]` with the user input
 - Check if we are on a `MultiSelectEntityScreen` and passing the input as `selectedValues[]` to the `handleInputAndUpdateSession` call.
 - print the `guid` corresponding to the saved `selectedValues[]` back on command prompt for user reference. 
 
 I have never ran CLI before myself so I might be slightly off in specifics of how it works but hopefully should not be far off from the general idea. Let me know if you have any questions and I would be happy to collab as well.

src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
shubham1g5: @snopoke I think I have time this week to do this work though it seems like you have already started on 'case search input' work on FP which also touches this area. You have any suggestions on how to best manage the interlap between the two. I can start on it today and probably PR by tomorrow EOD but if you wanna take it with 'case search input' work that's also fine with me.

snopoke: @shubham1g5 I haven't gotten to that point yet. I'm currently just working on making it possible to have an entry with a post request work. So it should be fine for you to go ahead.

src/main/java/org/commcare/core/interfaces/VirtualDataInstanceCache.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.sql.SQLException;
+import java.util.UUID;
+
+import javax.annotation.Nullable;
+
+/**
+ * Read and write operations for entity selections made on a mult-select Entity Screen
+ */
+public interface VirtualDataInstanceCache {
```
snopoke, line 13: how about renaming this to `VirtualDataInstanceStorage` since it is a primary store and not a cache?

ctsims: Agreed. I think 'cache' sets the wrong expectation for how this is used.

src/main/java/org/commcare/core/interfaces/VirtualDataInstanceCache.java
```diff
@@ -0,0 +1,20 @@
...
+
+    UUID write(VirtualDataInstance dataInstance);
```
ctsims, line 15: Is there a specific reason why we anchor on UUID objects here in particular?

For better or worse, we are fairly consistent in the other parts of the app in simply keying on strings that happen to be uuid's, and I'm worried about mixing the datatypes resulting in a sprawling set of implicit casts.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,117 @@
...
+    public void updateSelection(String input, @Nullable String[] selectedValues) throws CommCareSessionException {
+        setSelectedEntities(input, selectedValues);
+    }
+
+    private void setSelectedEntities(String input, @Nullable String[] selectedValues)
+            throws CommCareSessionException {
+        if (input.contentEquals(USE_SELECTED_VALUES)) {
+            processSelectedValues(selectedValues);
+        } else {
+            UUID inputId = UUID.fromString(input);
```
ctsims, line 53: I'm fairly confused about what this clause and the `USE_SELECTED_VALUES` structure does exactly It might be worth refactoring into separate methods whcih more clearly articulate why the same call path has very different behaviors.

src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
ctsims: Strongly agree with Simon that these two mechanisms seem very similar in structure and requirements and that it would be great to see the caching and loading unified. 

I think there's still some question in my mind about schema management expectations. Since we're defining the structure of these virtual docs, I think we'll want quite clear documentation about what the different kinds of virtual docs consist of, and we'll want to confirm whether we are going to provide versioning or graceful back/forwards compatibility with the types of changes we expect.

ctsims (commented):
One thing I want to be mindful of with this PR:

We are making a pretty big addition / change to the engine processing, but the exact usage specs (document schemas, uri reference patterns, session lifecycle descriptions, etc) aren't linked in the PR. I think it would be helpful for those specs and approaches available on the PR, both to facilitate review (and ensure design review) and so that we have a clear trail later about the scheme that this was **intended** to implement

src/main/java/org/commcare/core/interfaces/VirtualDataInstanceCache.java
shubham1g5: This has been carried forward from the actual [postgres model](https://github.com/dimagi/formplayer/blob/multiSelectList/src/main/java/org/commcare/formplayer/db/migration/V23__init_virtual_data_instance.java#L13) itself. Are you suggesting that we do a type conversion before writing to the Postgres or change the datatype in db itself ?

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
shubham1g5: will do, there are some details on this behaviour [here](https://github.com/dimagi/formplayer/wiki/Multi-Select-Case-Lists#http-lifecycle-bw-web-apps-and-formplayer)

src/main/java/org/commcare/core/interfaces/VirtualDataInstanceCache.java
shubham1g5: Implemented in the other PR to avoid conflicts - https://github.com/dimagi/commcare-core/pull/1109/commits/5ebe896802bc4a4b0e7e3b4edcaf5d5b8d8c566b

src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
shubham1g5: >  we'll want to confirm whether we are going to provide versioning or graceful back/forwards compatibility

That's a good point, I would assume that we need to be backward compatible here in general. Possibly a good idea to add versions to the virtual docs right away to facilitiate smoother migrations in future ?

Corresponding changes are in https://github.com/dimagi/commcare-core/pull/1109

src/main/java/org/commcare/core/interfaces/VirtualDataInstanceCache.java
ctsims: My inclination would be to change datatype itself and just key with a unique indexed string to stay consistent with other GUID indexes in the system. 

It's less efficient, for sure, than a proper DB UUID but I think the odds of us running into an issue with having a mix of string represented UUID's and typed UUID's are pretty high. It might be an interesting experiment to try to use strongly typed UUID's here to see if we can start shifting, but it would be a bad outcome if we only saw the edges late enough that we ended up casting UUID's <-> Strings

shubham1g5: https://github.com/dimagi/commcare-core/pull/1109/commits/e888ecf7e6afe5285d7be8d84d725c691c47462f and https://github.com/dimagi/formplayer/pull/1139/commits/1f3c5b392be59fa69aca53391dfdb56e73a95b44

shubham1g5:
@ctsims @snopoke FYI that I have added a wiki page on Virtual Instances [here](https://github.com/dimagi/commcare-core/wiki/Virtual-Data-Instances), let me know if you have suggestions on any specific pointers that we should add there.

snopoke:
> @ctsims @snopoke FYI that I have added a wiki page on Virtual Instances [here](https://github.com/dimagi/commcare-core/wiki/Virtual-Data-Instances), let me know if you have suggestions on any specific pointers that we should add there.

Thanks @shubham1g5, generally looks good. Adding comments here since you can't comment on the wiki:

* `selected_cases`: I still think this should be `selected-cases` to be consistent with other instances where a dash is used instead of an `_` (maybe too late to change now?)

* typo:
```diff
- <instance id="search-input" src="jr://instance/search_input"/>
+ <instance id="search-input" src="jr://instance/search-input"/>
```
* is this page referenced anywhere else? I think it should at least be referenced in the query datum docs: https://github.com/dimagi/commcare-core/wiki/Suite20#query-datum (and in the 'instance-datum' docs when we put those up).

src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java
```diff
@@ -23,17 +23,21 @@
  * @author ctsims
  */
 public class ExternalDataInstance extends DataInstance {
-    private String reference;
 
+    public static final String JR_SESSION_REFERENCE = "jr://instance/session";
+    public static final String JR_CASE_DB_REFERENCE = "jr://instance/casedb";
+    public static final String JR_LEDGER_DB_REFERENCE = "jr://instance/ledgerdb";
+    public static final String JR_SEARCH_INPUT_REFERENCE = "jr://instance/search_input";
```
snopoke, line 30:
```suggestion
    public static final String JR_SEARCH_INPUT_REFERENCE = "jr://instance/search-input";
```
This got changed from a '-' to an '_' here: https://github.com/dimagi/commcare-core/commit/b1d80eed3e07dd8f5f2e039dcef493d6b43a9b26#diff-c023914dd76eddfd2ac9ce91bd463205b5258e85f05dffbc1826fa0548c32fd8R23

shubham1g5:
>  I still think this should be selected-cases to be consistent with other instances where a dash is used instead of an _ (maybe too late to change now?)

Sounds good, I have changes the instance src to dashed in wiki and [code](https://github.com/dimagi/commcare-core/pull/1082/commits/3b44e986b4afb08ca3bbb12b20f6fa1a2eb34f4e). @orangejenny we would need to change the instance connector `src` on HQ as well from `jr://instance/selected_cases ` to `jr://instance/selected-cases `

> is this page referenced anywhere else? I think it should at least be referenced in the query datum docs: https://github.com/dimagi/commcare-core/wiki/Suite20#query-datum (and in the 'instance-datum' docs when we put those up).

Added that to the [query datum](https://github.com/dimagi/commcare-core/wiki/Suite20#query-datum) and [multi-select wiki ](https://github.com/dimagi/formplayer/wiki/Multi-Select-Case-Lists#selected-cases-instance)

shubham1g5:
By the way, it seems like we also already call the usual external data instance as 'virtual' data instance atleast in [one place](https://github.com/dimagi/commcare-core/wiki/Performance:-Virtual-Instances).  So the terminology is still a bit confusing here. There is also a potential change we can make where we can store the external instance like case search results instance in the `virtual_data_instance` table. And if we think about CommCare Session instance, it's also actually formed based on user inputs. That is to say I don't think there is a meaningful distinction here that `virtual` keyword is making and how about we just call these all external data instances and change the terminology in code accordingly as well.

ctsims:
Hi Shubham,

Just reviewing some of the docs and I'm still coming away with a couple of questions about lifecycle management and how the new virtual instances are injected.

I didn't see how the [search-input](https://github.com/dimagi/commcare-core/wiki/Virtual-Data-Instances#search-input-instance) or [selected-cases](https://github.com/dimagi/commcare-core/wiki/Virtual-Data-Instances#selected-cases-instance) instances delineate **which** user input their virtual schema should connect to. 

If the current design is that the session populates at most one of these, and that there's no way to correlate it with a specific sessions step, I think that's going to cause problems sooner rather than later. We fairly regularly need to perform multiple operations of the same type in a single session, so I think the connector needs a way to refer to a specific selection.

I'm not sure if I've mentioned this but I also ahve lingering concerns about the anchoring on `selected-cases` v. `selected-entities` since entity lists are used to select items from fixtures, organization trees, etc. As far as I can tell, the selection output would be valid for any vector list, is there a reason we're restricting the name to cases?

snopoke:
> I didn't see how the [search-input](https://github.com/dimagi/commcare-core/wiki/Virtual-Data-Instances#search-input-instance) or [selected-cases](https://github.com/dimagi/commcare-core/wiki/Virtual-Data-Instances#selected-cases-instance) instances delineate **which** user input their virtual schema should connect to.
> 
> If the current design is that the session populates at most one of these, and that there's no way to correlate it with a specific sessions step, I think that's going to cause problems sooner rather than later. We fairly regularly need to perform multiple operations of the same type in a single session, so I think the connector needs a way to refer to a specific selection.

This has been an ongoing discussion. For this use case the instance ID matches the datum ID of the `instance-datum`.

For another use case where we are adding an instance for query prompt data (https://github.com/dimagi/commcare-core/pull/1112) it's less clear what the instance ID should be since queries don't have an ID so to speak. The query result data creates an instance with the ID specified in the `storage-instance` attribute. We could also use that as a base for the user input instance and append a qualifier e.g. `results-user-input`.

The naming is also a bit uncertain, should it be `search-input`, `user-input`, `query-prompts`, `query-data`?

shubham1g5:
Thanks @ctsims for inputs,  made a few changes in response - 

1. Renamed the existing virtual instance wiki as [external data instance](https://github.com/dimagi/commcare-core/wiki/Performance:-External-Data-Instances) to avoid the confusion
2. Added small info in the new [virtual instance wiki](https://github.com/dimagi/commcare-core/wiki/Virtual-Data-Instances) around it's lifecycle being tied to a specific stack frame step in the session. Let me know if this still looks incomplete and needs more detail around the same. 
3. [Renamed instance](https://github.com/dimagi/commcare-core/pull/1082/commits/7f4dd703e67b19302dc2abe4dc38ff3e68feabe7) as 'selected-entities'. It only changes the instance reference to `jr://instance/selected-entities` and not the instance id itself. Instance id is connected to the `instance-datum` id and I guess it's fine for that to be `selected-cases` in correspondence with the normal `case_id` datum we have for single select case lists.

ctsims:
@snopoke 

> This has been an ongoing discussion. For this use case the instance ID matches the datum ID of the instance-datum.

If this is the case, I think we should be differentiating the root instance type in the src URI, IE:

`jr://instance/selected-entitites/<datum-id>`

the ID of the instance in the connector (ie: `<instance id="<instance id>" src=""`) is intended to be exclusively part of the application's definition, not part of the platform's definition, so we should avoid restricting what values can go into it. 

> For another use case where we are adding an instance for query prompt data (https://github.com/dimagi/commcare-core/pull/1112) it's less clear what the instance ID should be since queries don't have an ID so to speak.

Query data is still part of an individual step, right? I'd presume the ID of the session step would be the easiest to anchor on.

ctsims:
> Thanks @ctsims for inputs, made a few changes in response -
> 
> 1. Renamed the existing virtual instance wiki as [external data instance](https://github.com/dimagi/commcare-core/wiki/Performance:-External-Data-Instances) to avoid the confusion

That's perfect, thanks! 

> 2. Added small info in the new [virtual instance wiki](https://github.com/dimagi/commcare-core/wiki/Virtual-Data-Instances) around it's lifecycle being tied to a specific stack frame step in the session. Let me know if this still looks incomplete and needs more detail around the same.

Thanks for sharing, I'll take a look and might add some details or suggest some areas which would help clarify. 

> 3. [Renamed instance](https://github.com/dimagi/commcare-core/pull/1082/commits/7f4dd703e67b19302dc2abe4dc38ff3e68feabe7) as 'selected-entities'. It only changes the instance reference to `jr://instance/selected-entities` and not the instance id itself. Instance id is connected to the `instance-datum` id and I guess it's fine for that to be `selected-cases` in correspondence with the normal `case_id` datum we have for single select case lists.

Agreed that the `id` and the `src` titles can be totally unrelated. It's fine for HQ to inject `selected-entities` refs and refer to them with other ids. We'll need a diverse way to reference those anyway, since we will need to be able to refer to multiple sets of selected cases within a single session.

snopoke:
> @snopoke
> 
> > This has been an ongoing discussion. For this use case the instance ID matches the datum ID of the instance-datum.
> 
> If this is the case, I think we should be differentiating the root instance type in the src URI, IE:
> 
> `jr://instance/selected-entitites/<datum-id>`
> 
> the ID of the instance in the connector (ie: `<instance id="<instance id>" src=""`) is intended to be exclusively part of the application's definition, not part of the platform's definition, so we should avoid restricting what values can go into it.

Interesting, that's not how I've been thinking about it but I think that makes sense.

I think app builder will continue to use the datum ID for the instance IDs as a convention though, but that seems fine.

> 
> > For another use case where we are adding an instance for query prompt data (#1112) it's less clear what the instance ID should be since queries don't have an ID so to speak.
> 
> Query data is still part of an individual step, right? I'd presume the ID of the session step would be the easiest to anchor on.

What do you mean by the step ID? Query's don't have an ID attribute do they?

One other point is that it wouldn't be something opaque but rather something predictable that app builders can determine.

ctsims:
> Interesting, that's not how I've been thinking about it but I think that makes sense.
> 
> I think app builder will continue to use the datum ID for the instance IDs as a convention though, but that seems fine.

Yeah, seems 100% ok for the app builder to use the same ID's. 

One last quick note is that it might be worth going one step further on the uri scheme and denoting both mechanism and schema, ie:

```
jr://instance/virtual/<instance-schema>/<instance-id>
jr://instance/virtual/selected-entitites/selected-cases`
jr://instance/virtual/selected-entitites/selected-cases-patient`
jr://instance/virtual/search-input/results`

```

The `virtual` instances are something we think we'll have a number of and they all share common semantics (a virtual instance is created by a session step and available until the end of that session). 

> 
> > > For another use case where we are adding an instance for query prompt data (#1112) it's less clear what the instance ID should be since queries don't have an ID so to speak.
> > 
> > 
> > Query data is still part of an individual step, right? I'd presume the ID of the session step would be the easiest to anchor on.
> 
> What do you mean by the step ID? Query's don't have an ID attribute do they?
> 
> One other point is that it wouldn't be something opaque but rather something predictable that app builders can determine.

Right, I totally forgot that there wasn't a datum or step ID for [<query> datums](https://github.com/dimagi/commcare-core/wiki/Remote-Request-Query-and-Post-Configurations). I think the most straightforward thing do do would be to query off of the `@storage-instance` key from the [datum definition](https://github.com/dimagi/commcare-core/wiki/Suite20#query-datum) That's essentially already used by a different virtual instance connector to qualify an instance-id, so it would make sense to connect `<instance id="search-input:results" src="jr://instance/virtual/search-input/results">` to map to the existing `instance('results')\` output from that step.

snopoke:
> > @snopoke
> > > This has been an ongoing discussion. For this use case the instance ID matches the datum ID of the instance-datum.
> > 
> > 
> > If this is the case, I think we should be differentiating the root instance type in the src URI, IE:
> > `jr://instance/selected-entitites/<datum-id>`
> > the ID of the instance in the connector (ie: `<instance id="<instance id>" src=""`) is intended to be exclusively part of the application's definition, not part of the platform's definition, so we should avoid restricting what values can go into it.
> 
> Interesting, that's not how I've been thinking about it but I think that makes sense.
> 

On second thoughts I don't think this works. Instances are looked up by their ID, not the ref. The ref is only used to determine how to initialize them. It might work for standard instances like casedb but I don't think it does for dynamic ones. The query instances have their ID set in the XML definition and must be referred to by that ID.

It might be different once you are in a form and you can specify the instance declaration, but even then I don't think the current code will work if you change the ID (I'm need to dig into the code to confirm).

ctsims:
@snopoke What do you mean? We definitely load fixtures by an ID that's defined in the src uri ref. 

https://github.com/dimagi/commcare-core/blob/ce29258248030b230743b55f9501e5abd22f37f9/src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java#L111

snopoke:
> We definitely load fixtures by an ID that's defined in the src uri ref.

Yes I see that now (I was replying from my phone before). So I think what we are saying is that the ref should uniquely identify a specific instance in the given context (e.g. a session). The instance ID is purely for defining the name of that instance in the evaluation context for future reference by the `instance(id)` xpath expression.

That all makes sense to me but we still have an issue with these dynamic instances since there is no instance declaration available in the suite file that allows the user to define what the ID should be. For instances that are used in entries the instance ID's are taken from the declaration in the entry (similarly in forms). For example, if the app builder references the 'selected-value' instance in a query datum xpath expression, there is nowhere to declare the instance in order to define what it's ID should be. It can't be declared in the entry since it won't exist until after the user has given their input.

So either we need to declare them in the entry but make them lazy or else we need to set pre-defined IDs which seems to go against the design of instance declarations.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,152 @@
...
+            return;
+        }
+        if (storageReferenceId != null) {
+            if (selectedValuesInstance == null) {
+                selectedValuesInstance = virtualDataInstanceStorage.read(storageReferenceId);
+            }
+            ExternalDataInstanceSource externalDataInstanceSource = buildSelectedValuesInstanceSource(
+                    selectedValuesInstance, storageReferenceId);
+            session.setDatum(STATE_MULTIPLE_DATUM_VAL, mNeededDatum.getDataId(),
+                    storageReferenceId.toString(), externalDataInstanceSource);
```
snopoke, line 112: nit: storageReferenceId is already a string
```suggestion
                    storageReferenceId, externalDataInstanceSource);
```

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -0,0 +1,152 @@
...
+public class MultiSelectEntityScreen extends EntityScreen {
+
+    public static final String USE_SELECTED_VALUES = "use_selected_values";
+    private int maxSelectValue = -1;
+
+    private final VirtualDataInstanceStorage virtualDataInstanceStorage;
+    private String storageReferenceId;
+    private ExternalDataInstance selectedValuesInstance;
+
+    public MultiSelectEntityScreen(boolean handleCaseIndex, boolean full,
```
snopoke, line 29: It seems like a MultiSelectEntityScreen will never want to `handleCaseIndex` so I think this could be removed and `false` passed to the super constructor

ctsims:
@snopoke I definitely see the challenge on ordering, here.

> declare them in the entry but make them lazy

My sense is that this is the most practical leading option, and I think the instance infrastructure may already actually make this somewhat straightforward since I think instance init is a two-part process.

Thinking it over I see a couple of different ways to make this work. One is an implicit lazy init where we failfast, and the other is treating the instance structure as "valid" but empty / data free until session context is provided. I'm generally inclined to fail as early as possible if that's practical, but if the schema of the instances is essentially static, we could alternatively conceptualize the abstraction as an accessor (almost like the casedb) which is getting filled in and changed by session context. That might make it less awkward for the structure to exist out of order.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
shubham1g5: I am assuming that for CLI we would want this to be `true` similar to the [EntityScreen](https://github.com/dimagi/commcare-core/blob/formplayer/src/cli/java/org/commcare/util/cli/ApplicationHost.java#L433) ?

snopoke (approved):
Some small comments but nothing blocking

shubham1g5:
Going to merge this PR for tomorrow's deploy, though would followup on pending discussions on this PR in separate smaller PRs.

src/main/java/org/commcare/session/SessionInstanceBuilder.java
```diff
@@ -39,7 +39,7 @@ private static void addSessionNavData(TreeElement sessionRoot, SessionFrame fram
      */
     private static void addDatums(TreeElement sessionData, SessionFrame frame) {
         for (StackFrameStep step : frame.getSteps()) {
-            if (SessionFrame.STATE_DATUM_VAL.equals(step.getType()) ||
+            if (SessionFrame.isEntitySelectionDatum(step.getType()) ||
```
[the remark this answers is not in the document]
shubham1g5: One particular use case is during mark and rewind for case claim. We define the rewind as -
````
<stack>
 <push>
     <rewind value="instance('commcaresession')/session/data/selected_cases"/>
  </push>
     </stack>
````

And the value from this rewind gets inserted into the instance-datum for the primary case list.

PR #1084, opened 2022-04-01
Initializes Selected cases instance For Multi Select Case Lists 

shubham1g5:
https://github.com/dimagi/formplayer/pull/1112

### DIFF
--- a/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
@@ -87,6 +87,17 @@ public class MultiSelectEntityScreen extends EntityScreen {
         }
     }
 
+    @Override
+    public boolean referencesContainStep(String stepValue) {
+        try {
+            String[] cachedSelection = entitiesSelectionCache.read(stepValue);
+            return cachedSelection != null;
+        } catch (SQLException throwables) {
+            throwables.printStackTrace();
+        }
+        return false;
+    }
+
     public int getMaxSelectValue() {
         return maxSelectValue;
     }
--- a/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
@@ -90,12 +90,10 @@ public class MultiSelectEntityScreen extends EntityScreen {
     @Override
     public boolean referencesContainStep(String stepValue) {
         try {
-            String[] cachedSelection = entitiesSelectionCache.read(stepValue);
-            return cachedSelection != null;
+            return entitiesSelectionCache.contains(stepValue);
         } catch (SQLException throwables) {
-            throwables.printStackTrace();
+            throw new RuntimeException("An error occurred trying to read entity selections", throwables);
         }
-        return false;
     }
 
     public int getMaxSelectValue() {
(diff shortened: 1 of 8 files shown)


src/main/java/org/commcare/data/xml/VirtualInstances.java
```diff
@@ -24,4 +28,15 @@ public static VirtualDataInstance buildSearchInputInstance(Map<String, String> u
...
+
+    public static VirtualDataInstance buildSelectedValuesInstance(
+            ExternalDataInstance instance, String[] selectedValues) {
+        List<SimpleNode> nodes = new ArrayList<>();
+        for (String selectedValue : selectedValues) {
+            nodes.add(SimpleNode.textNode(SELCTED_CASES_INSTANCE_NODE_NAME, selectedValue));
+        }
+        TreeElement root = TreeBuilder.buildTree(instance.getInstanceId(), SELCTED_CASES_INSTANCE_ROOT_NAME,
+                nodes);
+        return new VirtualDataInstance(SEARCH_INSTANCE_ID, root);
```
snopoke, line 40: This needs a different instance ID
```suggestion
        return new VirtualDataInstance("search-input", root);
```

shubham1g5: thanks for catching, [Corrects instanceid](https://github.com/dimagi/commcare-core/pull/1084/commits/6ea2a14df19497d67239586b8992aee0fa4a6d39)

src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
```diff
@@ -89,10 +90,27 @@ public InstanceRoot generateRoot(ExternalDataInstance instance) {
             return setupRemoteData(instance);
         } else if (ref.contains("migration")) {
             return setupMigrationData(instance);
+        } else if (ref.contains("selected_cases")) {
+            return setupSelectedCases(instance);
         }
         return ConcreteInstanceRoot.NULL;
     }
 
+    protected InstanceRoot setupSelectedCases(ExternalDataInstance instance) {
```
snopoke, line 99: Would be good to add javadoc with a note that this should be implemented by subclasses.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -87,6 +87,17 @@ protected void updateSession(CommCareSession session) {
         }
     }
 
+    @Override
+    public boolean referencesContainStep(String stepValue) {
+        try {
+            String[] cachedSelection = entitiesSelectionCache.read(stepValue);
+            return cachedSelection != null;
+        } catch (SQLException throwables) {
+            throwables.printStackTrace();
```
snopoke, line 96: Does this end up being logged anywhere? I would have though that using a logger would be better.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -87,6 +87,15 @@ protected void updateSession(CommCareSession session) {
         }
     }
 
+    @Override
+    public boolean referencesContainStep(String stepValue) {
```
snopoke, line 91: Where will this get used?

shubham1g5: This gets called when [we rebuild the menu session from session frame](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L90) after EOF navigation or case claim to establish the validity of the datum in the session frame.

snopoke (approved):

PR #1085, opened 2022-04-04
added exclude from search functionality

Robert-Costello:
Add a new Search Property configured as "Exclude From Search Filters"
[Spec for this feature](https://docs.google.com/document/d/1FeHaa1JoA6NCRJfPX8Nt3joOAlqzFEgpT0NKlZkgy1I/edit#)
[Related HQ PR](https://github.com/dimagi/commcare-hq/pull/31205)
Screen Shot 2022-04-04 at 9 28 35 AM


src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -48,6 +48,8 @@
     @Nullable
     private ItemsetBinding itemsetBinding;
 
+    private boolean exclude;
```
snopoke, line 51:
```suggestion
    @Nullable
    private XPathExpression exclude;
```

src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -62,6 +64,7 @@ public QueryPrompt(String key, String appearance, String input, String receive,
         this.display = display;
         this.itemsetBinding = itemsetBinding;
         this.defaultValueExpr = defaultValueExpr;
+        this.exclude = exclude;
```
snopoke, line 67: this needs to be added as a parameter to the constructor

src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -74,6 +77,7 @@ public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOExcep
         display = (DisplayUnit)ExtUtil.read(in, DisplayUnit.class, pf);
         itemsetBinding = (ItemsetBinding)ExtUtil.read(in, new ExtWrapNullable(ItemsetBinding.class), pf);
         defaultValueExpr = (XPathExpression)ExtUtil.read(in, new ExtWrapNullable(new ExtWrapTagged()), pf);
+        exclude = ExtUtil.readBool(in);
```
snopoke, line 80:
```suggestion
        exclude = (XPathExpression)ExtUtil.read(in, new ExtWrapNullable(new ExtWrapTagged()), pf);
```

src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -86,6 +90,7 @@ public void writeExternal(DataOutputStream out) throws IOException {
         ExtUtil.write(out, display);
         ExtUtil.write(out, new ExtWrapNullable(itemsetBinding));
         ExtUtil.write(out, new ExtWrapNullable(defaultValueExpr == null ? null : new ExtWrapTagged(defaultValueExpr)));
+        ExtUtil.writeBool(out, exclude);
```
snopoke, line 93:
```suggestion
        ExtUtil.write(out, new ExtWrapNullable(exclude == null ? null : new ExtWrapTagged(exclude)));
```

src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -126,6 +131,10 @@ public XPathExpression getDefaultValueExpr() {
         return defaultValueExpr;
     }
 
+    public boolean getExclude() {
```
snopoke, line 134:
```suggestion
    public XPathExpression getExclude() {
```

src/main/java/org/commcare/xml/QueryPromptParser.java
```diff
@@ -46,6 +47,7 @@ public QueryPrompt parse() throws InvalidStructureException, IOException, XmlPul
         DisplayUnit display = null;
         ItemsetBinding itemsetBinding = null;
         XPathExpression defaultValue = null;
+        boolean exclude = "true".equals(parser.getAttributeValue(null, ATTR_EXCLUDE));
```
snopoke, line 50: needs to be updated to handle xpath expressions;
```java
String excludeValueString = parser.getAttributeValue(null, ATTR_EXCLUDE);
XPathExpression exclude = null;
if (excludeValueString != null) {
    // needs error handling
    exclude = XPathParseTool.parseXPath(excludeValueString);
}
```

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -124,7 +124,9 @@ public URL getBaseUrl() {
             for (Enumeration e = userAnswers.keys(); e.hasMoreElements(); ) {
                 String key = (String)e.nextElement();
                 String value = userAnswers.get(key);
-                if (!StringUtils.isEmpty(value)) {
+                QueryPrompt prompt = queryDatum.getUserQueryPrompts().get(key);
+                boolean exclude = prompt.getExclude();
+                if (!StringUtils.isEmpty(value) && !exclude) {
```
snopoke, line 129: evaluate the xpath to determine what the value will be
```suggestion
                XPathExpression excludeExpr = prompt.getExclude();
                if (excludeExpr == null || !(boolean) excludeExpr.eval(evaluationContext)) {
```

snopoke (commented):
Feedback based on the spec discussion of switching to an xpath expression for the exclude attribute.

Also there appear to be a number of changes that need to be merged in from the `formplayer` branch.

src/main/java/org/commcare/suite/model/QueryPrompt.java
Robert-Costello: Thanks, missed that when transferring changes from testing in formplayer.

snopoke (commented):
Looking good. How about a test? You can add some to `org.commcare.backend.suite.model.test.CaseClaimModelTests` which already has a few for `getRawQueryParams`

Robert-Costello:
Is there a way to run only commcare-core tests? I'm finding a bit difficult to troubleshoot running tests in formplayer. I'm going to push up the test I've been working on. I could use some suggestions on changes to the test suite.xml.

src/main/java/org/commcare/xml/QueryPromptParser.java
```diff
@@ -47,9 +48,21 @@ public QueryPrompt parse() throws InvalidStructureException, IOException, XmlPul
...
+            try {
+                exclude = XPathParseTool.parseXPath(excludeValueString);
+            } catch (XPathSyntaxException e) {
+                InvalidStructureException toThrow = new InvalidStructureException(String.format(
+                        "Invalid XPath Expression in QueryPrompt %s",
+                        e.getMessage()), parser);
+                toThrow.initCause(e);
+                throw toThrow;
+            }
+        }
```
snopoke, line 65: This block is almost identical to the one below and could be extracted into a method. I would even go so far as to say it could be added to the parent class since there is similar code in other parses but that could be a separate PR.

```java
XPathExpression exclude = this.parseXPath(excludeValueString, "prompt[@exclude]");

...

private XPathExpression parseXPath(String value, String context) throws InvalidStructureException {
    if (value == null) {
        return null;
    }
    try {
        return XPathParseTool.parseXPath(value);
    } catch (XPathSyntaxException e) {
        InvalidStructureException toThrow = new InvalidStructureException(String.format(
                "Invalid XPath Expression for %s: %s",
                context, e.getMessage()), parser);
        toThrow.initCause(e);
        throw toThrow;
    }
}

File: src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java [hunk 9]
```diff
@@ -112,6 +120,23 @@ private void testGetRawQueryParamsWithUserInput(Map<String, String> userInput, L
...
+        session.setCommand("patient-search");
+
+        RemoteQuerySessionManager remoteQuerySessionManager = RemoteQuerySessionManager.buildQuerySessionManager(
+                session, session.getEvaluationContext(), new ArrayList<>());
+
+        userInput.forEach(remoteQuerySessionManager::answerUserPrompt);
+
+        Multimap<String, String> params = remoteQuerySessionManager.getRawQueryParams(false);
+
+        Assert.assertEquals(expected, params.get("_xpath_query_exclude"));
```
snopoke, line 137: This should be looking at the `exclude_patient_id` param. Also I would expect the param to be missing entirely rather than have an empty value:

```
Assert.assertFalse(params.containsKey("exclude_patient_id"))
```

src/main/java/org/commcare/xml/QueryPromptParser.java
Robert-Costello: That makes sense. I can work on that PR after this one.

snopoke (approved):

PR #1086, opened 2022-04-05
Build with github action

shubham1g5:
Experimenting for now to see how it fits in with Jenkins build.


.github/workflows/build.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Build with Gradle
+        uses: gradle/gradle-build-action@v2
+        with:
+          gradle-version: 7.3
+          arguments: build
+      - name: Cli Jar
+        uses: gradle/gradle-build-action@v2
+        with:
+          gradle-version: 7.3
+          arguments: cliJar
```
snopoke, line 28: this won't run the tests so it will only check that the code compiles right?

snopoke (commented):
Great to see this!

.github/workflows/build.yml
shubham1g5: the `build` step above compiles the code and runs the tests as well. The `cliJar` step just makes the `jar` using compiled classes.

snopoke: ah, right, I missed that step.

.github/workflows/build.yml
```diff
@@ -0,0 +1,33 @@
+name: CommCare Core Builds for Formplayer
+
+on:
+  push:
+    branches: [formplayer]
+  pull_request:
+    branches: [formplayer]
```
snopoke, line 7: wouldn't we also want this on the `master` branch?

snopoke: That seems OK to me but then let's rename this file to `build_formplayer.yml`

.github/workflows/build_formplayer.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: gradle/gradle-build-action@v2
+        with:
+          gradle-version: 7.3
+          arguments: build
+      - name: Cli Jar
+        uses: gradle/gradle-build-action@v2
+        with:
+          gradle-version: 7.3
+          arguments: cliJar
+      - name: Upload Archives
```
snopoke, line 29: What's the use case for storing the archives?

shubham1g5: Not sure, mostly adding it here since Jenkins build was storing it. Though I imagine the idea being if someone wants to use CLI directly without having to built FP locally, they can use these jars.

snopoke: I would punt this for now until we have a need.

snopoke (approved):
Looks great, thanks for the extra effort to make this support both branches!

PR #1097, opened 2022-04-16
Support for Post Relevancy based on post body params

shubham1g5:
Adds ability to reference post params in the relevancy condition for post. This allows us to not fire case claim request when all selected cases are owned by the user. Example - 

````
<remote-request>
    <post url="http://localhost:8000/a/test-3/phone/claim-case/"  relevant="$case_id != ‘’”>
      <data key=”case_id”
      nodeset=”instance(‘search_cases’)/results/value” exclude=”count(instance(‘casedb’)/casedb/case[@case_id = current()/.]) = 1” ref=”.”/>
      </data>
   </post>
````
formplayer: https://github.com/dimagi/formplayer/pull/1122/commits/b996aed2562b36482dcfe0b2e0d1de976f0a43d1

### DIFF
--- a/src/main/java/org/commcare/suite/model/PostRequest.java
+++ b/src/main/java/org/commcare/suite/model/PostRequest.java
@@ -48,7 +48,12 @@ public class PostRequest implements Externalizable {
     public Multimap<String, String> getEvaluatedParams(EvaluationContext evalContext) {
         Multimap<String, String> evaluatedParams = ArrayListMultimap.create();
         for (QueryData queryData : params) {
-            evaluatedParams.putAll(queryData.getKey(), queryData.getValues(evalContext));
+            Iterable<String> val = queryData.getValues(evalContext);
+            if (val.iterator().hasNext()) {
+                evaluatedParams.putAll(queryData.getKey(), val);
+            } else {
+                evaluatedParams.put(queryData.getKey(), "");
+            }
         }
         return evaluatedParams;
     }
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/main/java/org/commcare/suite/model/PostRequest.java [hunk 1]
```diff
@@ -48,7 +48,12 @@ public URL getUrl() {
     public Multimap<String, String> getEvaluatedParams(EvaluationContext evalContext) {
         Multimap<String, String> evaluatedParams = ArrayListMultimap.create();
         for (QueryData queryData : params) {
-            evaluatedParams.putAll(queryData.getKey(), queryData.getValues(evalContext));
+            Iterable<String> val = queryData.getValues(evalContext);
+            if (val.iterator().hasNext()) {
+                evaluatedParams.putAll(queryData.getKey(), val);
+            } else {
+                evaluatedParams.put(queryData.getKey(), "");
```
snopoke, line 55: This is changing the value that get's put in right? Before if there were no values then the key would not be present in `evaluatedParams` but now the key will be present with an empty string as the value. Making that change will also affect the actual POST request and I think it will require changes on HQ's side.

I think there are 2 options to avoid this:
1. Pass in a boolean arg to determine if blank keys should be included
    - This is simpler but may not be the best long term option if we end up needing to support empty strings as post param values (maybe that's unlikely).
3. Use a virtual instance which would then support using the `count` function to determine if the key was present.
    - More complex and would require changing the 'relevant' expression but is the most flexible and future proof.

shubham1g5: Ahh right! I do think it's quite unlikely that we need to support blank values in post params for now so I am inclined towards the simpler option 1 here.

shubham1g5: f26f65dd60129d2cddb3d8a21f67944064b5cff2

snopoke (approved):

PR #1098, opened 2022-04-19
Switch entities selection to a generic xml instance interface

shubham1g5:
https://github.com/dimagi/formplayer/pull/1125

### DIFF
--- a/src/main/java/org/javarosa/core/reference/ReferenceHandler.java
+++ b/src/main/java/org/javarosa/core/reference/ReferenceHandler.java
@@ -39,4 +39,9 @@ public class ReferenceHandler {
             return staticManager;
         }
     }
+
+    public static void clearInstance() {
+        threadLocalManager.set(null);
+        staticManager = null;
+    }
 }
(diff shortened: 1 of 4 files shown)


shubham1g5:
@snopoke bumping this for review as well. Thanks!

src/main/java/org/javarosa/core/reference/ReferenceHandler.java
```diff
@@ -39,4 +39,9 @@ public static ReferenceManager instance() {
             return staticManager;
         }
     }
+
+    public static void clearInstance() {
+        threadLocalManager.set(null);
+        staticManager = null;
+    }
```
snopoke, line 46: how is this change related?

snopoke (approved):
One question but otherwise looks good

src/main/java/org/javarosa/core/reference/ReferenceHandler.java
shubham1g5: This is to [fix tests in Formplayer](https://github.com/dimagi/formplayer/pull/1125/commits/e9d372a96180272184c9dd05984f52da5a512990).  For test cases using unarchived ccz file, we need to clear references to staticManager for app to correctly install.

PR #1101, opened 2022-04-19
Dont wait for case detail if user selected action (master)

AddisonDunn:
Same as #1093 but for the master branch (I was able to test w/ the build on master).


AddisonDunn:
Realizing I need to actually make the same commit, built off master. Sorry.

PR #1034, opened 2021-08-26
Updates to AGP 7 

ShivamPokhriyal:
cross-request: https://github.com/dimagi/commcare-android/pull/2526

PR #1104, opened 2022-04-25
Updates for new Java Runtimes

ctsims:
When updating my Android Studio environment to Bumblebee I ran across a few issues that seem to be associated with updated Java runtimes

1) Resource Context Issues

When running tests from a fresh checkout, tests which relied on Resources were failing with FileNotFoundExceptions

```
logger> resources: Install attempt unsuccessful from: jr://resource/app_for_text_tests/profile.ccpr|File not found when opening resource as stream
```
After some digging it looks like the default static classloader in the runtime can end up not being connected, so this change redirects from the static (`System.class`) classloader to the class specific one.

2) PrettyXML issues

We have a helper method in a little-used library for pretty printing XML which has relied on deprecated XML libraries for a long time. This just updates to use modern built-in xml transforms, and replaces the associated test output.

### DIFF
--- a/src/main/java/org/javarosa/core/reference/ResourceReference.java
+++ b/src/main/java/org/javarosa/core/reference/ResourceReference.java
@@ -31,7 +31,7 @@ public class ResourceReference implements Reference {
     public boolean doesBinaryExist() throws IOException {
         //Figure out if there's a file by trying to open
         //a stream to it and determining if it's null.
-        InputStream is = System.class.getResourceAsStream(URI);
+        InputStream is = this.getClass().getResourceAsStream(URI);
         if (is == null) {
             return false;
         } else {
@@ -42,7 +42,7 @@ public class ResourceReference implements Reference {
 
     @Override
     public InputStream getStream() throws IOException {
-        InputStream is = System.class.getResourceAsStream(URI);
+        InputStream is = this.getClass().getResourceAsStream(URI);
         if (is == null) {
             throw new FileNotFoundException("File not found when opening resource as stream");
         }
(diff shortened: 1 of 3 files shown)


snopoke (approved):

shubham1g5:
@ctsims Thanks for raising this. There is a similar PR open here - https://github.com/dimagi/commcare-core/pull/1067/files for Java 11 changes. One notable difference between the two is that other PR uses `LSSerializer` to prettify xml. It seems like `LSSerializer` is preferred method over the other because of performance reasons. 

Except that we had not really merged these PRs before on `master` because Jenkins could only build `commcare-core` jobs with a specific java version. Though now with the tranisition to github action we do have the capability to build jobs differently on different branches. I will make that change in https://github.com/dimagi/commcare-core/pull/1067/ and we can probably close this one out if you agree with the use of `LSSerializer`

ctsims:
Hi Shubham,

Got it, thanks for the catch. I didn't pick the XML serializer with any particular intent, so I'm happy to use `LSSerializer` instead.

One thing: Your PR is targeting the `formplayer` branch. What's the realistic timeframe on getting everything merged back up? I'm a bit concerned about ending up back in the world where `formplayer` and `master` are regularly far apart or where we are blocking on build environment fixes by pushing to formplayer before back up into master. Maybe I should adapt just the common changes and get master buildable today and we can defer on needing to get everything back into sync

shubham1g5:
@ctsims Sorry I linked wrong PR. This is the [original PR](https://github.com/dimagi/commcare-core/pull/1034/files) that already targets master. I am going to make some changes there for consistency with `formplayer` and make it up to date.

shubham1g5:
closing in favour of https://github.com/dimagi/commcare-core/pull/1034

PR #1108, opened 2022-05-04
Support 'post' requests from form entry

snopoke:
This adds a 'post' attribute to the FormEntry class and adjusts navigation code to support executing the post request right before entering the form.

The design doc suggests merging the `Entry`, `ViewEntry`, `FormEntry` and `RemoteRequestEntry` classes. I consider this as an incremental step towards that.

There may be additional changes required when I start working on the Formplayer side. So far I've only tested this with the CLI.

I recommend reviewing by commit.

Design doc: https://docs.google.com/document/d/1Mmx1FrYZrcEmWidqSkNjC_gWSJ6xzRFKoP3Rn_xSaj4/edit#
Jira: https://dimagi-dev.atlassian.net/browse/USH-1946


src/main/java/org/commcare/suite/model/FormEntry.java
```diff
@@ -41,6 +41,11 @@ public FormEntry(String commandId, DisplayUnit display,
         this.post = post;
     }
 
+    @Override
+    public boolean hasForm() {
+        return xFormNamespace != null;
```
shubham1g5: 💯

src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -204,10 +205,16 @@ public String getNeededData(EvaluationContext evalContext) {
         } else if (entriesForCurrentCommand.isEmpty()) {
             // No entries available directly within the current command, so we must need to select another menu
             return SessionFrame.STATE_COMMAND_ID;
-        } else if (entriesForCurrentCommand.size() == 1
-                && entriesForCurrentCommand.elementAt(0).getPostRequest() != null
-                && entriesForCurrentCommand.elementAt(0).getPostRequest().isRelevant(evalContext)) {
-            return SessionFrame.STATE_SYNC_REQUEST;
+        } else if (entriesForCurrentCommand.size() == 1) {
+            Entry entry = getPlatform().getEntry(currentCmd);
+            if (entry == null) {
```
shubham1g5, line 212: this should never happen since we have already established `entriesForCurrentCommand.size() == 1` ?

src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -220,12 +220,10 @@ && getCurrentFrameStepExtra(EXTRA_POST_SUCCESS) == null
             } else {
                 return null;
             }
-        } else if (entriesForCurrentCommand.size() > 1 || !entriesForCurrentCommand.elementAt(0).getCommandId().equals(currentCmd)) {
+        } else {
             //the only other thing we can need is a form command. If there's
             //still more than one applicable entry, we need to keep going
             return SessionFrame.STATE_COMMAND_ID;
-        } else {
```
shubham1g5, line 216: would this result in situations where we would be returning `STATE_COMMAND_ID` instead of `null` in the new code ? `null` is also symbolic to getting into a Xform so I am a bit concerned with this change.

src/cli/java/org/commcare/util/screen/SyncScreen.java
```diff
@@ -34,6 +24,9 @@ public class SyncScreen extends Screen {
     private PrintStream printStream;
     private boolean syncSuccessful;
 
+    public SyncScreen() {
```
shubham1g5, line 28: is this needed ?

src/cli/java/org/commcare/util/screen/SyncScreen.java
```diff
@@ -56,64 +49,30 @@ private void parseMakeRequest() throws CommCareSessionException {
...
                     printStream);
 
-                printStream.println(String.format("Sync successful with response %s", response));
+                printStream.println("Sync successful with response");
             } else {
                 printStream.println("Did not sync because case was already claimed.");
             }
             printStream.println("Press 'enter' to continue.");
-        } catch (IOException e) {
+        } catch (Exception e) {
```
shubham1g5, line 71: I think there is value in keeping this as `IOException` so that we don't gulp any code errors like NPE. `IOException` is more anticipated because of a network request involved and we should throw the underlying exception on unsuccesful response code as `IOException` as well.

src/cli/java/org/commcare/util/screen/SessionUtils.java
```diff
@@ -20,9 +23,18 @@
...
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+
 public class SessionUtils {
+    public SessionUtils() {}
 
-    public static void restoreUserToSandbox(UserSandbox sandbox,
+    public void restoreUserToSandbox(UserSandbox sandbox,
```
shubham1g5, line 37: It seems like we only changed it to non-static to handle tests. Curious if it's not possible to mock the static methods using Mockito ? That might make these changes redundant.

shubham1g5 (commented):
Looks great overall, left some minor queries.

src/main/java/org/commcare/session/CommCareSession.java
snopoke: This can happen if the command is a menu command that only has one entry.

src/cli/java/org/commcare/util/screen/SyncScreen.java
snopoke: I plan to use this in Formplayer instead of passing in nulls.

I would also like to refactor this class into a base class that can have a CLI extension and a Formplayer extension since most of the functionality here is for the CLI. Having one class makes it hard to know what is used where.

src/cli/java/org/commcare/util/screen/SessionUtils.java
snopoke: It is possible to do with Mockito but it's not a great design pattern to rely on static functions.

src/cli/java/org/commcare/util/screen/SyncScreen.java
snopoke: 8fc8d2d02

src/main/java/org/commcare/session/CommCareSession.java
snopoke: both of these if clause checks are handled in the other places above:
* if entries.isEmpty()
* if entries.size() == 1
  * Entry entry = getPlatform().getEntry(currentCmd);
  * if entry == null: // this handles the 2nd test `(!entries.elementAt(0).getCommandId().equals(currentCmd)`
* else  # this handles the 1st test (`entries.size() > 1`)

The only way we should return `null` (start an xform) is if there is a single entry with a command ID that matches the current command AND there isn't a post request to perform.

I think the updated logic makes this decision clearer.

src/main/java/org/commcare/session/CommCareSession.java
shubham1g5: I don't think I am understanding it correctly. Why would `entry` be null if the command menu has only one entry ?

snopoke: Because if the command is for a menu then it can't also be for an entry right? So `getPlatform().getEntry()` won't find an Entry with that command ID.

shubham1g5: yes, makes sense

src/main/java/org/commcare/session/CommCareSession.java
shubham1g5: great, agree that this is a better logic flow.

shubham1g5 (approved):

src/cli/java/org/commcare/util/screen/SyncScreen.java
```diff
@@ -129,6 +103,13 @@ public boolean prompt(PrintStream printStream) throws CommCareSessionException {
     @Override
     public boolean handleInputAndUpdateSession(CommCareSession commCareSession, String s, boolean allowAutoLaunch) throws CommCareSessionException {
         if (syncSuccessful) {
+            updateSessionOnSuccess();
+            Entry commandEntry = sessionWrapper.getCurrentEntry();
```
snopoke, line 107: @shubham1g5 with reference to this comment here: https://github.com/dimagi/commcare-core/pull/1109#discussion_r870999599, do you think this needs to use the passed in session rather than the one in the screen?

`getCurrentEntry` callsed `getCommand` on the session so it relies on the command in the session being correct. I THINK that the session the screen is initialised with will have the correct command.

shubham1g5: I think it's ok to use the passed session object to perform any READ operations as it tend to give the same result as using the screen's copy of session in almost all cases, Though the safest thing to do will be to rely on the passed instance of session just in case it was updated before this method was called.

src/cli/java/org/commcare/util/screen/SessionUtils.java
```diff
@@ -138,7 +138,7 @@ public int doPostRequest(
             PrintStream printStream
     ) throws IOException {
         String url = buildUrl(syncPost.getUrl().toString());
-        Multimap<String, String> params = syncPost.getEvaluatedParams(session.getEvaluationContext());
+        Multimap<String, String> params = syncPost.getEvaluatedParams(session.getEvaluationContext(), true);
```
orangejenny, line 141: I'm not 100% sure whether this should be passing true or false.

snopoke: Based on the usage here: https://github.com/dimagi/formplayer/blob/671ea55f7e3780e5b4e0dc2395d96812ce3af18f/src/main/java/org/commcare/formplayer/screens/FormplayerSyncScreen.java#L37-L38 I would say it should be `false`.

However, looking back in the history I'm not 100% sure what it should be. We [introduced the `exclude` attribute on post data elements](https://github.com/dimagi/commcare-core/commit/0a6be291a021c59941a13350437b7d5dc6bca6f2) and for a simple value data it's clear that if `exclude` evaluates to `true` then the entire param should be excluded. For data elements have a nodeset, the `exclude` expression applies to [each individual item in the list](https://github.com/dimagi/commcare-core/blob/3265df37fce0bcbecc40ab6c8c1bc4b72a8504ef/src/main/java/org/commcare/suite/model/ListQueryData.java#L64-L66). So the question is, if the list is empty, should the param still be included or not?

For now I'll change this param should be `false`.

PR #1109, opened 2022-05-10
Refactors VirtualDataInstance to extend from ExternalDataInstance

shubham1g5:
- VirtualDataInstance now extends from `ExternalDataInstance` 
- Bunch of changes are initialising instance from ExternalDataInstanceSource
- Instance initialisation for selected cases instance now is similar to the case search results instance initialisation aka they both makes use of stack frame's 'xmlInstanceSource' object to initialise the instance from. 

Better to review PR as a whole as there is some back and forth between commits.

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityScreen.java
@@ -370,4 +370,8 @@ public class EntityScreen extends CompoundScreenHost {
         }
         return false;
     }
+
+    public void updateDatum(CommCareSession session, String input) {
+        session.setEntityDatum(session.getNeededDatum(), input);
+    }
 }
--- a/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
@@ -7,6 +7,8 @@ import org.commcare.data.xml.VirtualInstances;
 import org.commcare.modern.session.SessionWrapper;
 import org.commcare.session.CommCareSession;
 import org.commcare.suite.model.MultiSelectEntityDatum;
+import org.javarosa.core.model.instance.ExternalDataInstanceSource;
+import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.core.model.instance.VirtualDataInstance;
 
@@ -24,6 +26,7 @@ public class MultiSelectEntityScreen extends EntityScreen {
 
     private final VirtualDataInstanceCache virtualDataInstanceCache;
     private UUID storageReferenceId;
+    private VirtualDataInstance selectedValuesInstance;
 
     public MultiSelectEntityScreen(boolean handleCaseIndex, boolean full,
             SessionWrapper session,
@@ -74,7 +77,7 @@ public class MultiSelectEntityScreen extends EntityScreen {
                 }
                 evaluatedValues[i] = getReturnValueFromSelection(currentReference);
             }
-            VirtualDataInstance selectedValuesInstance =
+            selectedValuesInstance =
                     VirtualInstances.buildSelectedValuesInstance(getSession().getNeededDatum().getDataId(),
                             selectedValues);
             UUID guid = virtualDataInstanceCache.write(selectedValuesInstance);
@@ -94,7 +97,14 @@ public class MultiSelectEntityScreen extends EntityScreen {
             return;
         }
         if (storageReferenceId != null) {
-            session.setDatum(STATE_MULTIPLE_DATUM_VAL, mNeededDatum.getDataId(), storageReferenceId.toString());
+            ExternalDataInstanceSource externalDataInstanceSource = ExternalDataInstanceSource.buildStorageBackedDataInstanceSource(
+                    selectedValuesInstance.getInstanceId(),
+                    (TreeElement)selectedValuesInstance.getRoot(),
+                    VirtualInstances.JR_SELECTED_VALUES_REFERENCE,
+                    selectedValuesInstance.useCaseTemplate(),
+                    storageReferenceId);
+            session.setDatum(STATE_MULTIPLE_DATUM_VAL, mNeededDatum.getDataId(),
+                    storageReferenceId.toString(), externalDataInstanceSource);
         }
     }
 
--- a/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
@@ -7,10 +7,10 @@ import org.commcare.data.xml.VirtualInstances;
 import org.commcare.modern.session.SessionWrapper;
 import org.commcare.session.CommCareSession;
 import org.commcare.suite.model.MultiSelectEntityDatum;
+import org.javarosa.core.model.instance.ExternalDataInstance;
 import org.javarosa.core.model.instance.ExternalDataInstanceSource;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.model.instance.TreeReference;
-import org.javarosa.core.model.instance.VirtualDataInstance;
 
 import java.util.UUID;
 
@@ -26,7 +26,7 @@ public class MultiSelectEntityScreen extends EntityScreen {
 
     private final VirtualDataInstanceCache virtualDataInstanceCache;
     private UUID storageReferenceId;
-    private VirtualDataInstance selectedValuesInstance;
+    private ExternalDataInstance selectedValuesInstance;
 
     public MultiSelectEntityScreen(boolean handleCaseIndex, boolean full,
             SessionWrapper session,
@@ -54,7 +54,7 @@ public class MultiSelectEntityScreen extends EntityScreen {
             processSelectedValues(selectedValues);
         } else {
             UUID inputId = UUID.fromString(input);
-            VirtualDataInstance cachedInstance = virtualDataInstanceCache.read(inputId);
+            ExternalDataInstance cachedInstance = virtualDataInstanceCache.read(inputId);
             if (cachedInstance == null) {
                 throw new CommCareSessionException(
                         "Could not make selection with reference id " + input + " on this screen. " +
@@ -97,6 +97,9 @@ public class MultiSelectEntityScreen extends EntityScreen {
             return;
         }
         if (storageReferenceId != null) {
+            if (selectedValuesInstance == null) {
+                selectedValuesInstance = virtualDataInstanceCache.read(storageReferenceId);
+            }
             ExternalDataInstanceSource externalDataInstanceSource = ExternalDataInstanceSource.buildStorageBackedDataInstanceSource(
                     selectedValuesInstance.getInstanceId(),
                     (TreeElement)selectedValuesInstance.getRoot(),
--- a/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
@@ -139,8 +139,8 @@ public class MultiSelectEntityScreen extends EntityScreen {
         return virtualDataInstanceCache.contains(UUID.fromString(stepValue));
     }
 
-    public VirtualDataInstance getVirtualInstance() {
-        return virtualDataInstanceCache.read(getStorageReferenceId());
+    public ExternalDataInstance getVirtualInstance() {
+        return selectedValuesInstance;
     }
 
     public int getMaxSelectValue() {
--- a/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
@@ -1,6 +1,7 @@
 package org.commcare.util.screen;
 
 import static org.commcare.session.SessionFrame.STATE_MULTIPLE_DATUM_VAL;
+import static org.javarosa.core.model.instance.ExternalDataInstanceSource.buildStorageBackedDataInstanceSource;
 
 import org.commcare.core.interfaces.VirtualDataInstanceCache;
 import org.commcare.data.xml.VirtualInstances;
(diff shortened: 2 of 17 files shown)


src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceSource.java
```diff
@@ -83,52 +105,60 @@ public void init(TreeElement root) {
...
     }
 
     @Override
     public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
-        sourceUri = ExtUtil.readString(in);
         instanceId = ExtUtil.readString(in);
         useCaseTemplate = ExtUtil.readBool(in);
+        sourceUri = ExtUtil.nullIfEmpty(ExtUtil.readString(in));
         requestData = (Multimap<String, String>)ExtUtil.read(in, new ExtWrapMultiMap(String.class), pf);
+        Object nullableRefId = ExtUtil.read(in, new ExtWrapNullable(String.class), pf);
```
shubham1g5, line 124: These serialisation changes are now resulting in breaking the deserialisation for a serialised form def (https://github.com/dimagi/formplayer/pull/1126/files#r868345569) and we will need to implement a fallback mechanism against such errors before deploying it to production.

src/main/java/org/javarosa/core/model/instance/VirtualDataInstance.java
```diff
@@ -11,69 +11,23 @@
 /**
  * Data instance that is created during execution based on user input of other live data
  */
-public class VirtualDataInstance extends DataInstance {
-    private TreeElement root = new TreeElement();
-    private InstanceBase base;
+public class VirtualDataInstance extends ExternalDataInstance {
```
shubham1g5, line 14: This is now exactly same as `ExternalDataInstance` . Should we now remove the concept of `VirtualDataInstance` completely ? Not sure if it might still be potentially useful to keep this distinction.

src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceSource.java
```diff
@@ -1,64 +1,86 @@
...
+                                       String sourceUri, Multimap requestData, UUID storageReferenceId) {
         this.instanceId = instanceId;
-        this.sourceUri = sourceUri;
-        this.requestData = requestData;
         this.root = root;
+        this.reference = reference;
         this.useCaseTemplate = useCaseTemplate;
+        this.sourceUri = sourceUri;
+        this.requestData = requestData;
+        this.storageReferenceId = storageReferenceId;
```
snopoke, line 53: Maybe validate that one of `sourceUri` and `storageReferenceId` is not null and that both are not supplied?

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -94,17 +104,43 @@ protected void updateSession(CommCareSession session) {
...
+        session.setDatum(STATE_MULTIPLE_DATUM_VAL, session.getNeededDatum().getDataId(),
+                input, externalDataInstanceSource);
+    }
+
+    private static ExternalDataInstanceSource buildSelectedValuesInstance(
+            ExternalDataInstance selectedValuesInstance, UUID storageReferenceId){
+        return buildStorageBackedDataInstanceSource(
+                selectedValuesInstance.getInstanceId(),
+                (TreeElement)selectedValuesInstance.getRoot(),
+                VirtualInstances.JR_SELECTED_VALUES_REFERENCE,
```
snopoke, line 132:
```suggestion
                selectedValuesInstance.getReference()
```

src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceSource.java
```diff
@@ -1,64 +1,86 @@
...
+    public static ExternalDataInstanceSource buildRemoteDataInstanceSource(
+            String instanceId, @Nullable TreeElement root,
+            boolean useCaseTemplate, String sourceUri,
+            Multimap requestData) {
+        return new ExternalDataInstanceSource(instanceId,root, JR_REMOTE_REFERENCE,
+                useCaseTemplate, sourceUri, requestData, null);
+    }
+
+
+    public static ExternalDataInstanceSource buildStorageBackedDataInstanceSource(
```
snopoke, line 78: I think this could always take an `ExternalDataInstance` and a reference like `org.commcare.util.screen.MultiSelectEntityScreen#buildSelectedValuesInstance`

snopoke (commented):
I've done a pass. So far looks generally OK. I plan to do another pass tomorrow.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -74,11 +78,17 @@ private void processSelectedValues(String[] selectedValues)
...
+            ExternalDataInstance instance = VirtualInstances.buildSelectedValuesInstance(
+                    getSession().getNeededDatum().getDataId(),
+                    selectedValues);
+            UUID guid = virtualDataInstanceCache.write(instance);
             storageReferenceId = guid;
+
+            // rebuild instance with the source
+            ExternalDataInstanceSource instanceSource = buildStorageBackedDataInstanceSource(
+                    instance.getInstanceId(), (TreeElement)instance.getRoot(),
+                    instance.getReference(), instance.useCaseTemplate(), storageReferenceId);
```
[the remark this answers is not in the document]
shubham1g5: Not sure if I rightly understand you though if it's regardign rebuilding the instance, the issue is `storageReferenceId` is only available once you build and save the instance without which we can't correctly build the source beforehand. 

> Also won't instance.useCaseTemplate() always be false?

it is for now, though I was not sure if there can be potential dynamic case instances which we might want to store in furture.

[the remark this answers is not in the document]
shubham1g5: > Not sure if I rightly understand you though if it's regardign rebuilding the instance, the issue is `storageReferenceId` is only available once you build and save the instance without which we can't correctly build the source beforehand. 

Ahh I got confused with `VirtualInstances.buildSelectedValuesInstance` , I got your point now 👍

src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceSource.java
shubham1g5: That would mean that we would always need to create instance twice once without source and then with the source returned from here, one such place for ex  would be in [VirtualDataInstanceService](https://github.com/dimagi/formplayer/blob/refactorMultiSelect/src/main/java/org/commcare/formplayer/services/VirtualDataInstanceService.java#L63) where having this definition is more convenient.

snopoke: Ah true, I did look at that but of course it's not the same type. Fine to leave as is.

src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java
```diff
@@ -66,17 +70,16 @@ private ExternalDataInstance(String reference, String instanceId,
         base.setChild(root);
     }
 
+    public static ExternalDataInstance buildInstance(ExternalDataInstanceSource source) {
```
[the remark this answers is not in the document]
snopoke: I also wonder if this shouldn't be in `ExternalDataInstanceSource` and called something like `toInstance` or `buildInstance`. That would make it quite a bit easier to get the instance from the source in a number of places.

src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceSource.java
```diff
@@ -1,64 +1,90 @@
...
+            String instanceId, @Nullable TreeElement root,
+            String reference, boolean useCaseTemplate,
+            UUID storageReferenceId) {
+        return new ExternalDataInstanceSource(instanceId, root, reference,
+                useCaseTemplate, null, ArrayListMultimap.create(), storageReferenceId);
```
snopoke, line 87: Why change this from `ImmutableMultimap.of()` to `ArrayListMultimap.create()`? It should never be mutated should it?

src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceSource.java
```diff
@@ -1,64 +1,90 @@
...
         this.useCaseTemplate = externalDataInstanceSource.useCaseTemplate();
+        this.sourceUri = externalDataInstanceSource.sourceUri;
         this.requestData = externalDataInstanceSource.requestData;
+        this.storageReferenceId = externalDataInstanceSource.storageReferenceId;
+    }
+
+    public static ExternalDataInstanceSource buildRemoteDataInstanceSource(
+            String instanceId, @Nullable TreeElement root,
+            boolean useCaseTemplate, String sourceUri,
+            Multimap requestData) {
```
snopoke, line 76:
```suggestion
            Multimap<String, String> requestData) {
```

src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceSource.java
```diff
@@ -1,64 +1,90 @@
...
-    public ExternalDataInstanceSource(String instanceId, TreeElement root, String sourceUri, boolean useCaseTemplate) {
-        this(instanceId, root, sourceUri, ImmutableMultimap.of(), useCaseTemplate);
+    public final static String JR_REMOTE_REFERENCE = "jr://instance/remote";
+
+    public ExternalDataInstanceSource() {
     }
 
-    public ExternalDataInstanceSource(String instanceId, TreeElement root, String sourceUri, Multimap<String, String> requestData, boolean useCaseTemplate) {
+    private ExternalDataInstanceSource(String instanceId, TreeElement root, String reference, boolean useCaseTemplate,
+                                       String sourceUri, Multimap requestData, UUID storageReferenceId) {
```
snopoke, line 46: I notice that you removed the type parameterization of requestData. Was that on purpose?

(also in the constructor below)

src/test/java/org/commcare/xml/TestInstances.java
```diff
@@ -31,38 +32,38 @@ public static Hashtable<String, DataInstance> getInstances() {
...
     }
 
-    public static VirtualDataInstance buildSessionInstance() {
+    public static ExternalDataInstance buildSessionInstance() {
         List<SimpleNode> nodes = ImmutableList.of(
                 SimpleNode.textNode("case_id", Collections.emptyMap(), "bang")
         );
         TreeElement root = TreeBuilder.buildTree(SESSION, SESSION, nodes);
-        return new VirtualDataInstance(SESSION, root);
+        return new ExternalDataInstance("jr://instance/session", SESSION, root);
```
snopoke, line 40: could we make these refs constants?

src/cli/java/org/commcare/util/screen/EntityScreen.java
```diff
@@ -370,4 +370,8 @@ public boolean referencesContainStep(String stepValue) {
         }
         return false;
     }
+
+    public void updateDatum(CommCareSession session, String input) {
+        session.setEntityDatum(session.getNeededDatum(), input);
```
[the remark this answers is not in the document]
shubham1g5: We need to update the `MenuSession` copy of the CommCareSession here which is responsible to advance the navigation rather than the internal EntityScreen which mostly snapshots the session state for which the `EntityScreen` was created for.

src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceSource.java
shubham1g5: thanks for catching, it was unintentional.

src/main/java/org/commcare/data/xml/VirtualInstances.java
```diff
@@ -19,24 +19,27 @@ public class VirtualInstances {
     public static final String SELCTED_CASES_INSTANCE_ROOT_NAME = "results";
     public static final String SELCTED_CASES_INSTANCE_NODE_NAME = "value";
 
-    public static VirtualDataInstance buildSearchInputInstance(Map<String, String> userInputValues) {
+    public final static String JR_SEARCH_INPUT_REFERENCE = "jr://instance/search_input";
+    public final static String JR_SELECTED_VALUES_REFERENCE = "jr://instance/selected_cases";
```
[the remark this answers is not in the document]
shubham1g5: yes that make sense to me, how about sticking to `jr://instance/remote` in that case since we already have that one utilised heavily in current apps ?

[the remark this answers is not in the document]
shubham1g5: And confirming if it matches your understanding that instanceId check is suffcient and `stepType` check [here](https://github.com/dimagi/commcare-core/blob/refactorMultiSelect/src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java#L204) is redundant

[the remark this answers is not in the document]
snopoke: > yes that make sense to me, how about sticking to jr://instance/remote in that case since we already have that one utilised heavily in current apps ?

I think it makes sense to use a separate reference since it does communicate something about the instance. `remote` reference are fetched from and external system while `virtual` ones are created during session navigation. I don't think there's a downside to introducing a new one.

I have also been thinking about if we can standardise the 'virtual' instance schema or otherwise communicate what the expected schema is. With 'remote' instances the schema is always the same (at least so far). But with these 'virtual' instances we already have 2 different schemas. This is probably a problem for another time but I was thinking of adding an optional 'schema' attribute to the instance def which could be used to indicate what the expected schema should be.


> And confirming if it matches your understanding that instanceId check is suffcient and `stepType` check [here](https://github.com/dimagi/commcare-core/blob/refactorMultiSelect/src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java#L204) is redundant

Yes, if necessary we could prevent adding instances into the stack with duplicate IDs.

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -1,14 +1,17 @@
 package org.commcare.util.screen;
 
 import static org.commcare.session.SessionFrame.STATE_MULTIPLE_DATUM_VAL;
+import static org.javarosa.core.model.instance.ExternalDataInstanceSource.buildVirtual;
```
snopoke, line 4: with the shorter name I makes sense to always call this with the class `ExternalDataInstanceSource.buildVirtual`

src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
```diff
@@ -86,37 +87,24 @@ public InstanceRoot generateRoot(ExternalDataInstance instance) {
             return setupFixtureData(instance);
         } else if (instance.getReference().contains("session")) {
```
snopoke, line 88: why are some of these 'startswith' and some 'contains'? Shouldn't they always be an exact match anyway?

src/cli/java/org/commcare/util/screen/EntityScreen.java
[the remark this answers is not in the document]
shubham1g5: yes I find it confusing as well, I generally use a mental model of performing any writes to the MenuSession copy and then it doesn't generally matter in practice what you use for reading from the session. Probably we can rename the screen's copy to something like `readOnlySession` to make it more evident ?

src/main/java/org/commcare/data/xml/VirtualInstances.java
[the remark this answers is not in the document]
shubham1g5: that distinction makes sense to me and I am happy with using `virtual`.

Though your second point makes me think there is value in keeping these as separate references because they have different schemas. Like why not use the reference itself as an indication of a distinct schema it represents instead of adding a separate field to the instance ?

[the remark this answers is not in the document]
snopoke: > Though your second point makes me think there is value in keeping these as separate references because they have different schemas. Like why not use the reference itself as an indication of a distinct schema it represents instead of adding a separate field to the instance ?

Maybe, though I still think it would be nice to namespace them.

snopoke (approved):

PR #1110, opened 2022-05-10
Query.init Search Screen Title value change

dominicochoa-dimagi:
This change is part of a series of PRs regarding the _Case Claim title translation per module_ feature change. 

Previously the Search Screen Title was retrieved from the localization values.
```Java
mTitle = Localization.get("case.search.title");
```

This new change instead retrieves the title value from the current session's command value.
```Java
CommCarePlatform platform = sessionWrapper.getPlatform();
mTitle = platform.getEntry(sessionWrapper.getCommand()).getText().evaluate();
```

Spec: https://docs.google.com/document/d/10PVDEk4MjQH0rPwLJSe5Fgi5AEG7GI2_2tC9A1qHTOY/edit#heading=h.t19aq1hpqun1
Jira: https://dimagi-dev.atlassian.net/browse/USH-1790


shubham1g5 (commented):
Looks good, you can complement it with a test in [here](https://github.com/dimagi/commcare-core/blob/formplayer/src/test/java/org/commcare/backend/suite/model/test/QueryModelTests.java#L32).

snopoke (commented):
Looks OK but I've just realized that while this works for remote request entries, it will display the form name for query screens that are part of a form entry.

FYI @orangejenny

.gitignore
```diff
@@ -15,7 +15,7 @@ backend/dist/*
 key.properties
 glob:application/custom-devices.xml
 cases/bin/
-application/custom-devices.xml
```
orangejenny, line 18: Is this changing line break characters or something? Might as well remove this file from this PR.

src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -110,7 +110,8 @@ public void init(SessionWrapper sessionWrapper) throws CommCareSessionException
         }
 
         try {
-            mTitle = Localization.get("case.search.title");
+            mTitle = Localization.get(getQueryDatum().getTitleLocaleId());
```
orangejenny, line 113: This should still fall back to the old style for now, until existing apps are migrated.

src/main/java/org/commcare/xml/SessionDatumParser.java
```diff
@@ -132,10 +133,11 @@ private RemoteQueryDatum parseRemoteQueryDatum()
             } else if ("prompt".equals(tagName)) {
                 String key = parser.getAttributeValue(null, "key");
                 userQueryPrompts.put(key, new QueryPromptParser(parser).parse());
+            } else if ("locale".equals(tagName)) {
```
orangejenny, line 136: Was it not feasible to do this before the loop, as we chatted about?

I was thinking something like [this](https://github.com/dimagi/commcare-core/blob/ce29258248030b230743b55f9501e5abd22f37f9/src/main/java/org/commcare/xml/DetailParser.java#L43-L51), only without the `checkNode` since the title isn't guaranteed to be there (until apps are migrated).

.gitignore
dominicochoa-dimagi: Not sure what happened here but this is fixed.

src/cli/java/org/commcare/util/screen/QueryScreen.java
orangejenny: Have you confirmed this returns null if nothing is found? [This code](https://github.com/dimagi/commcare-core/blob/d9066ea561f4ce85717b25263f9ddad695c98b29/src/main/java/org/javarosa/xml/util/ActionableInvalidStructureException.java#L28-L32) makes me think it throws an exception.

orangejenny: (as does the surrounding code in this file)

src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -114,6 +116,8 @@ private String getTitleLocaleStringLegacy() {
             mTitle = Localization.get("case.search.title");
         } catch (NoLocalizedTextException nlte) {
             mTitle = "Case Claim";
+        } catch (NullPointerException npe) {
```
orangejenny, line 119: Nit: you can make a single `catch` block handle both exceptions. Check out [this code](https://github.com/dimagi/formplayer/blob/4cd1ed39f4c58e2b1b2c5a9f00e38ed3923be2c9/src/main/java/org/commcare/formplayer/util/RequestUtils.java#L68-L73) for an example.

Same comment for the function above.

src/main/java/org/commcare/suite/model/RemoteQueryDatum.java
```diff
@@ -26,7 +29,7 @@ public class RemoteQueryDatum extends SessionDatum {
     private OrderedHashtable<String, QueryPrompt> userQueryPrompts;
     private boolean useCaseTemplate;
     private boolean defaultSearch;
-    private String titleLocaleId;
+    private Text titleLocaleId;
```
orangejenny, line 32: Since this is a now a `Text` instead of a locale id, `titleLocaleId` doesn't make sense as a name. How about just `title`? (This goes for the various related function names, too.)

orangejenny (commented):
Updates look great. I left two minor, hopefully final comments.

orangejenny (approved):

src/main/java/org/commcare/xml/SessionDatumParser.java
```diff
@@ -140,6 +134,9 @@ private RemoteQueryDatum parseRemoteQueryDatum()
             } else if ("prompt".equals(tagName)) {
                 String key = parser.getAttributeValue(null, "key");
                 userQueryPrompts.put(key, new QueryPromptParser(parser).parse());
+            } else if ("title".equals(tagName)) {
```
orangejenny, line 137: @snopoke This commit is similar to how @dominicochoa-dimagi originally wrote the code, and then I asked him to move the title parsing before the loop because I thought that made sense conceptually (since there's at most one title and it'll show up first). Can you say more about why the having the title parsing before the loop doesn't work?

snopoke: It does work as long as the title is present. If it is not present then the parser skips over the next element because the while loop calls `nextTagInBlock` again.

If we want to keep the title handling separate then we could change the while loop into a 'do while' loop.

> since there's at most one title
We could add a check in the loop to make sure the 'title' var is null.

> it'll show up first
It showing up first is purely an artifact of how HQ builds the XML, there's no reason it couldn't come later. I would prefer if the parser did not make assumptions about how the XML is structured where it's not functionally different.

orangejenny: ah, I see, the double `nextTagInBlock`. And fair point about the ordering not being a functional part of the suite spec.

(sorry @dominicochoa-dimagi for the misdirection)

PR #1111, opened 2022-05-17
Only create new storage if needed

gherceg:
This might be excessive, and we can instead just check:
```
if (storageRegistry.containsKey(key) && storageRegistry.get(key).getClass() == type) {
  return;
}
```
but I figured it might be nice to log a scenario where the type differs for the same key. Open to suggestions.

### DIFF
--- a/src/main/java/org/javarosa/core/services/storage/StorageManager.java
+++ b/src/main/java/org/javarosa/core/services/storage/StorageManager.java
@@ -46,7 +46,7 @@ public class StorageManager {
             throw new RuntimeException("No storage factory has been set; I don't know what kind of storage utility to create. Either set a storage factory, or register your StorageUtilitys directly.");
         }
         if (storageRegistry.containsKey(key)) {
-            if (storageRegistry.get(key).getClass() == type) {
+            if (storageRegistry.get(key).getPrototype() == type) {
                 return;
             } else {
                 throw new RuntimeException("Attempting to change storage type for key " + key + " from type " + storageRegistry.get(key).getClass() + " to type " + type + ".");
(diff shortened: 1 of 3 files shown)


src/main/java/org/javarosa/core/services/storage/StorageManager.java
```diff
@@ -45,6 +45,13 @@ public void registerStorage(String key, Class type) {
         if (storageFactory == null) {
             throw new RuntimeException("No storage factory has been set; I don't know what kind of storage utility to create. Either set a storage factory, or register your StorageUtilitys directly.");
         }
+        if (storageRegistry.containsKey(key)) {
+            if (storageRegistry.get(key).getClass() == type) {
+                return;
+            } else {
+                Logger.log("Warning", "Overwriting storage type " + storageRegistry.get(key).getClass() + " with type " + type + " for key " + key + ".");
```
shubham1g5, line 52: probably throw a `RuntimeException` from here ? I would think it almost always a code mistake if the storage is keyed with 2 different types in the same request lifecycle.

snopoke: +1 for throwing an exception here.

snopoke (commented):
Looks good pending the exception change.

gherceg:
First, I didn't take into account that the storage class type was wrapped with `IStorageUtilityIndexed`, so there is always a class mismatch with the current logic. The prototype class within `IStorageUtilityIndexed` is private. Do we think it is worth adding a getter for the prototype to be able to add this check, or just scratch this and don't bother checking?

Also, I'll need to also create a new PR as I branched off of master, not formplayer.

snopoke:
> First, I didn't take into account that the storage class type was wrapped with `IStorageUtilityIndexed`, so there is always a class mismatch with the current logic. The prototype class within `IStorageUtilityIndexed` is private. Do we think it is worth adding a getter for the prototype to be able to add this check, or just scratch this and don't bother checking?

I don't see an issue with exposing the prototype field as a read only field.

gherceg:
Replaced by https://github.com/dimagi/commcare-core/pull/1113

PR #1112, opened 2022-05-18
store case search user input in the session for later access

snopoke:
Part 2 of https://docs.google.com/document/d/1Mmx1FrYZrcEmWidqSkNjC_gWSJ6xzRFKoP3Rn_xSaj4/edit#

This builds off the virtual instance storage in https://github.com/dimagi/commcare-core/pull/1082 and adds support for:

1. storing multiple instances per stack frame step
2. saving the query prompt values as an instance and making that available via the session

Cross request: https://github.com/dimagi/formplayer/pull/1141

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -191,7 +191,7 @@ public class QueryScreen extends Screen {
         return instanceOrError;
     }
 
-    public void setQueryDatum(ExternalDataInstance dataInstance) {
+    public void updateSession(ExternalDataInstance dataInstance) {
         if (dataInstance != null) {
             sessionWrapper.setQueryDatum(dataInstance);
         }
@@ -278,7 +278,7 @@ public class QueryScreen extends Screen {
         Multimap<String, String> requestData = getRequestData(false);
         InputStream response = makeQueryRequestReturnStream(url, requestData);
         Pair<ExternalDataInstance, String> instanceOrError = processResponse(response, url, requestData);
-        setQueryDatum(instanceOrError.first);
+        updateSession(instanceOrError.first);
         if (currentMessage != null) {
             out.println(currentMessage);
         }
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -111,6 +111,7 @@ public class QueryScreen extends Screen {
         for (Map.Entry<String, QueryPrompt> queryPromptEntry : userInputDisplays.entrySet()) {
             fields[count] = queryPromptEntry.getValue().getDisplay().getText().evaluate(
                     sessionWrapper.getEvaluationContext());
+            count++;
         }
 
         try {
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -9,7 +9,9 @@ import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.Multimap;
 
 import org.commcare.cases.util.StringUtils;
+import org.commcare.core.encryption.CryptUtil;
 import org.commcare.core.interfaces.VirtualDataInstanceStorage;
+import org.commcare.data.xml.VirtualInstances;
 import org.commcare.modern.session.SessionWrapper;
 import org.commcare.modern.util.Pair;
 import org.commcare.session.CommCareSession;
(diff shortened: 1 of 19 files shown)


src/main/java/org/commcare/suite/model/StackFrameStep.java
```diff
@@ -102,12 +100,42 @@ public String getValue() {
         return value;
     }
 
-    public boolean hasXmlInstance() {
-        return xmlInstanceSource != null;
+    public void addDataInstanceSources(ExternalDataInstanceSource source) {
+        if (source == null) {
+            return;
```
shubham1g5, line 105: should we throw an exception here instead ?

shubham1g5 (approved):
Looks great 🥇

src/main/java/org/commcare/session/CommCareSession.java
```diff
@@ -465,20 +466,25 @@ public void setDatum(String action, String keyId, String value) {
...
+    public void setQueryDatum(ExternalDataInstance queryResultInstance, ExternalDataInstance... extras) {
         SessionDatum datum = getNeededDatum();
         if (datum instanceof RemoteQueryDatum) {
-            StackFrameStep step =
-                    new StackFrameStep(SessionFrame.STATE_QUERY_REQUEST,
-                            datum.getDataId(), datum.getValue(), queryResultInstance.getSource());
+            StackFrameStep step = new StackFrameStep(
+                    SessionFrame.STATE_QUERY_REQUEST, datum.getDataId(), datum.getValue());
+            step.addDataInstanceSources(queryResultInstance.getSource());
+            Arrays.stream(extras).forEach((instance) -> {
```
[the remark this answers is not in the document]
snopoke: Hmm, not sure. That does make sense. I can never remember what iteration methods apply to different classes and just default to trying 'forEach'. I'll update.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -143,13 +143,17 @@ public Multimap<String, String> getRawQueryParams(boolean skipDefaultPromptValue
...
-        ExternalDataInstance userInputInstance = VirtualInstances.buildSearchInputInstance(userQueryValues);
+        ExternalDataInstance userInputInstance = getUserInputInstance();
         return evaluationContext.spawnWithCleanLifecycle(
                 ImmutableMap.of(userInputInstance.getInstanceId(), userInputInstance)
         );
     }
 
+    public ExternalDataInstance getUserInputInstance() {
+        Map<String, String> userQueryValues = getUserQueryValues(false);
+        return VirtualInstances.buildSearchInputInstance(userQueryValues);
```
snopoke, line 154: TODO: the instance name needs to be unique for the session - one option could be to use `query/@storage-instance` with a prefix e.g. `resutls-search-input`.

On the topic of naming, is `search-input` right? What about `user-input`, `query-prompts`, `query-data`?

src/cli/java/org/commcare/util/screen/QueryScreen.java
```diff
@@ -191,12 +196,23 @@ public Pair<ExternalDataInstance, String> processResponse(InputStream responseDa
         return instanceOrError;
     }
 
-    public void setQueryDatum(ExternalDataInstance dataInstance) {
+    public void updateSession(ExternalDataInstance dataInstance) {
         if (dataInstance != null) {
-            sessionWrapper.setQueryDatum(dataInstance);
+            ExternalDataInstance userInputInstance = getUserInputInstance();
+            sessionWrapper.setQueryDatum(dataInstance, userInputInstance);
```
snopoke, line 202: I realize that this is going to create a new instance on every request until the form session starts. Will need to think about how to prevent that

snopoke: Also see this discussion about instance IDs & refs: https://github.com/dimagi/commcare-core/pull/1082#issuecomment-1134448655

snopoke: [f2d6616](https://github.com/dimagi/commcare-core/pull/1112/commits/f2d6616c292b09ff2ad86bd7804e60f710cfdccd)

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
snopoke: Also see the discussion here: https://github.com/dimagi/commcare-core/pull/1082#issuecomment-1134448655

src/main/java/org/commcare/suite/model/StackFrameStep.java
snopoke: [dbc3541](https://github.com/dimagi/commcare-core/pull/1112/commits/dbc35417d37bcebb867b317eebbce7c2911760b9)

orangejenny:
(force pushed to remove a merge I had done)

shubham1g5 (approved):

PR #1115, opened 2022-05-27
Added dates to supported case search prompts

orangejenny:
Companion to https://github.com/dimagi/commcare-hq/pull/31671


shubham1g5:
I also see a time widget in https://github.com/dimagi/commcare-hq/pull/31671 , does that not have a different prompt type than `date` ?

shubham1g5 (approved):
great, thanks for clarification!

PR #1118, opened 2022-06-14
Add Multi Select Case Lists to CLI

stephherbers:
Ticket: [USH-2131](https://dimagi-dev.atlassian.net/browse/USH-2131) 

This add the multi select case list functionality to the CLI. For Multi Select screens, the user can select one or more entries (separated by a comma if more than one in the input) from the case list.

### DIFF
--- a/src/cli/java/org/commcare/util/cli/ApplicationHost.java
+++ b/src/cli/java/org/commcare/util/cli/ApplicationHost.java
@@ -324,8 +324,6 @@ public class ApplicationHost {
                     boolean waitForCaseDetail = false;
                     if (screen instanceof MultiSelectEntityScreen) {
                         String[] selectedValues = input.split(",");
-                        System.out.println("selectedValues");
-                        System.out.println(selectedValues);
                         screen.handleInputAndUpdateSession(mSession, USE_SELECTED_VALUES, false, selectedValues);
                     } else {
                         if (screen instanceof EntityScreen) {
--- a/src/cli/java/org/commcare/util/cli/ApplicationHost.java
+++ b/src/cli/java/org/commcare/util/cli/ApplicationHost.java
@@ -324,7 +324,7 @@ public class ApplicationHost {
                     boolean waitForCaseDetail = false;
                     if (screen instanceof MultiSelectEntityScreen) {
                         String[] selectedValues = input.split(",");
-                        screen.handleInputAndUpdateSession(mSession, USE_SELECTED_VALUES, false, selectedValues);
+                        screenIsRedrawing = screen.handleInputAndUpdateSession(mSession, USE_SELECTED_VALUES, false, selectedValues);
                     } else {
                         if (screen instanceof EntityScreen) {
                             boolean isAction = input.startsWith("action "); // Don't wait for case detail if action
--- a/src/cli/java/org/commcare/util/cli/ApplicationHost.java
+++ b/src/cli/java/org/commcare/util/cli/ApplicationHost.java
@@ -56,6 +56,8 @@ import java.io.InputStreamReader;
 import java.io.PrintStream;
 import java.util.*;
 
+import static org.commcare.util.screen.MultiSelectEntityScreen.USE_SELECTED_VALUES;
+
 /**
  * CLI host for running a commcare application which has been configured and instatiated
  * for the provided user.
@@ -322,7 +324,9 @@ public class ApplicationHost {
                     boolean waitForCaseDetail = false;
                     if (screen instanceof MultiSelectEntityScreen) {
                         String[] selectedValues = input.split(",");
-                        screen.handleInputAndUpdateSession(mSession, input, false, selectedValues);
+                        System.out.println("selectedValues");
+                        System.out.println(selectedValues);
+                        screen.handleInputAndUpdateSession(mSession, USE_SELECTED_VALUES, false, selectedValues);
                     } else {
                         if (screen instanceof EntityScreen) {
                             boolean isAction = input.startsWith("action "); // Don't wait for case detail if action
--- a/src/cli/java/org/commcare/util/cli/ApplicationHost.java
+++ b/src/cli/java/org/commcare/util/cli/ApplicationHost.java
@@ -26,6 +26,7 @@ import org.commcare.util.screen.CommCareSessionException;
 import org.commcare.util.screen.EntityListSubscreen;
 import org.commcare.util.screen.EntityScreen;
 import org.commcare.util.screen.MenuScreen;
+import org.commcare.util.screen.MultiSelectEntityScreen;
 import org.commcare.util.screen.QueryScreen;
 import org.commcare.util.screen.Screen;
 import org.commcare.util.screen.SessionUtils;
@@ -64,7 +65,7 @@ import java.util.*;
 public class ApplicationHost {
     private final CommCareConfigEngine mEngine;
     private final CommCarePlatform mPlatform;
-    private UserSandbox mSandbox;
+    private MockUserDataSandbox mSandbox;
     private CLISessionWrapper mSession;
 
     private boolean mUpdatePending = false;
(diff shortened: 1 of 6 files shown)


src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -319,15 +320,20 @@ private boolean loopSession() throws IOException {
                     // which ultimately updates the session, so getNextScreen will move onto the form list,
                     // skipping the entity detail. To avoid this, flag that we want to force a redraw in this case.
                     boolean waitForCaseDetail = false;
-                    if (screen instanceof EntityScreen) {
-                        boolean isAction = input.startsWith("action "); // Don't wait for case detail if action
-                        EntityScreen eScreen = (EntityScreen)screen;
-                        if (!isAction && eScreen.getCurrentScreen() instanceof EntityListSubscreen) {
-                            waitForCaseDetail = true;
+                    if (screen instanceof MultiSelectEntityScreen) {
+                        String[] selectedValues = input.split(",");
```
shubham1g5, line 324: The semantics on `input` doesn't seem entirely clear to me here. I think we want to support 2 kinds of input `use_selected_values` or a guid to `MultiSelectEntityScreen` based on [this request lifecycle](https://github.com/dimagi/formplayer/wiki/Multi-Select-Case-Lists#http-lifecycle-bw-web-apps-and-formplayer). One idea is to take the selected values as - `use_selected_values val1 val2` while guid would just be a singular value like normal Entity Screen `guid_value` .  We should do input validation here to make sure input is one of these two types and throw an error in case of a mismatch, and parse the input so that we make either of these 2 calls depending on the input type  `screen.handleInputAndUpdateSession(mSession, USE_SELECTED_VALUES, false, selectedValues);` or for guid as input `screen.handleInputAndUpdateSession(mSession, input, false, null);`

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
```diff
@@ -236,10 +239,24 @@ public boolean handleInputAndUpdateHost(String input, EntityScreen host, boolean
...
+                } else {
+                    int index = Integer.parseInt(input);
+                    TreeReference[] selectedRefs = new TreeReference[1];
+                    selectedRefs[0] = mChoices[index];
+                    if (host instanceof MultiSelectEntityScreen) {
+                        host.updateSelection(USE_SELECTED_VALUES, selectedRefs);
+                    } else {
+                        host.updateSelection(input, selectedRefs);
+                    }
+                }
```
shubham1g5, line 259: In correlation to my comment above on  semantics of `input` and `use_selected_values`, this should get modified as below to handle input correctly for multi-select. 
```suggestion              
                TreeReference[] selectedRefs = null;
                if (host instanceof MultiSelectEntityScreen) {
                    if (input.contentEquals(USE_SELECTED_VALUES)) {
                        if (selectedValues == null) {
                            throw new IllegalArgumentException("selected values can't be null");
                        }
                        selectedRefs = new TreeReference[selectedValues.length];
                        for (int i = 0; i < selectedValues.length; i++) {
                            int index = Integer.parseInt(selectedValues[i]);
                            selectedRefs[i] = mChoices[index];
                        }
                    }
                } else {
                    int index = Integer.parseInt(input);
                    selectedRefs = new TreeReference[1];
                    selectedRefs[0] = mChoices[index];
                }
                host.updateSelection(input, selectedRefs);   
```

shubham1g5 (changes requested):
looks like we need some more work on this to get the request lifecycle correspond to the Web Apps.

src/cli/java/org/commcare/util/cli/ApplicationHost.java
snopoke: I don't think that level of complexity is necessary in the CLI since we aren't rebuilding the session on each navigation request. I don't think we should ever have to use the the GUID in the CLI and any input would always be the selected values.

src/cli/java/org/commcare/util/mocks/MockUserDataSandbox.java
```diff
@@ -104,6 +109,10 @@ public IStorageUtilityIndexed<FormInstance> getAppFixtureStorage() {
         return appFixtureStorage;
     }
 
+    public VirtualDataInstanceStorage getExternalInstanceStorage() {
```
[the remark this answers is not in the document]
snopoke: +1

src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -319,15 +320,20 @@ private boolean loopSession() throws IOException {
...
                     // skipping the entity detail. To avoid this, flag that we want to force a redraw in this case.
                     boolean waitForCaseDetail = false;
-                    if (screen instanceof EntityScreen) {
-                        boolean isAction = input.startsWith("action "); // Don't wait for case detail if action
-                        EntityScreen eScreen = (EntityScreen)screen;
-                        if (!isAction && eScreen.getCurrentScreen() instanceof EntityListSubscreen) {
-                            waitForCaseDetail = true;
+                    if (screen instanceof MultiSelectEntityScreen) {
+                        String[] selectedValues = input.split(",");
+                        screen.handleInputAndUpdateSession(mSession, input, false, selectedValues);
```
snopoke, line 325: This should pass in the `USE_SELECTED_VALUES` const along with the list of values instead of the raw input.
```suggestion
                        screen.handleInputAndUpdateSession(mSession, USE_SELECTED_VALUES, false, selectedValues);

```

src/cli/java/org/commcare/util/cli/ApplicationHost.java
snopoke: Is that what the CLI is for? I don't think about it like that. If that is the goal then we should really also be rebuilding the session on every 'request' and probably push a lot of the FP code down into the CLI.

I always thought the CLI was for testing apps.

shubham1g5: Ahh Interesting, looks like I probably am incorrect in my assumption here in that case. Makes sense to not use `guid` ever for CLI input.

src/cli/java/org/commcare/util/mocks/MockUserDataSandbox.java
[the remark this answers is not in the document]
stephherbers: ah very nice https://github.com/dimagi/commcare-core/pull/1118/commits/d2e7ed5926d623e6e521b454c2e5889e979d324b

src/cli/java/org/commcare/util/cli/ApplicationHost.java
```diff
@@ -319,15 +320,20 @@ private boolean loopSession() throws IOException {
...
+                    }
+                    else {
+                        if (screen instanceof EntityScreen) {
+                            boolean isAction = input.startsWith("action "); // Don't wait for case detail if action
+                            EntityScreen eScreen = (EntityScreen)screen;
+                            if (!isAction && eScreen.getCurrentScreen() instanceof EntityListSubscreen) {
+                                waitForCaseDetail = true;
+                            }
                         }
+                        screenIsRedrawing = !screen.handleInputAndUpdateSession(mSession, input, false, null);
```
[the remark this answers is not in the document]
stephherbers: Ah good point changed here: https://github.com/dimagi/commcare-core/pull/1118/commits/d655cf39844cbabb5b108be3f96888132cf14ad5
One thing is I did have to remove `final` from the handleInputAndUpdateSession in `CompundScreenHost` which felt a bit ironic, and I wanted to flag

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -181,4 +181,12 @@ public int getMaxSelectValue() {
     public String getStorageReferenceId() {
         return storageReferenceId;
     }
+
+    @Override
+    public boolean handleInputAndUpdateSession(CommCareSession session, String input,
+            boolean allowAutoLaunch, String[] selectedValues) throws CommCareSessionException {
+        getCurrentScreen().handleInputAndUpdateHost(input, this, allowAutoLaunch, selectedValues);
+        this.updateSession(session);
```
shubham1g5, line 189: you can just call `super.handleInputAndUpdateSession` here.

shubham1g5 (changes requested):
left one minor comment, but looks great overall!

shubham1g5 (approved):

snopoke (approved):

PR #1119, opened 2022-06-15
reverts exact instance matching due to errors with existing apps

shubham1g5:
I recently changed instance initialisation to check the exact instance references. This seem to cause issues with existing apps which are unfortunately using incorrect instance references. 

https://dimagi-dev.atlassian.net/browse/SAAS-13658

If we are going to make this change, we should give an advance notice to app builders to make this change.


ctsims (approved):

snopoke:
wow, that's horrible!

shubham1g5:
yeah! Wondering if we can have validations on HQ for instance references as a first step to this ? That would probably be more safer way to eliminate incorrect references before we implement the same on Formplayer/Mobile

PR #1122, opened 2022-06-17
get json property by name

Robert-Costello:
Adds a function to get a property from a stringified json object. To be used to support hidden values as geocoder receivers.
spec: https://docs.google.com/document/d/1v2AziWCpVI8MO8FTmEapUYwoN9rmYTQbjt-MlEWJcdU/edit#
ticket: https://dimagi-dev.atlassian.net/browse/USH-2137


src/main/java/org/commcare/util/GetJsonProperty.java
```diff
@@ -0,0 +1,24 @@
...
+import org.json.JSONObject;
+
+/** 
+ * @author rcostello
+ * @return A String value for the property name passed in if that property exists else a blank String
+ */
+
+public class GetJsonProperty {
+
+    public static String getJsonProperty(String stringifiedJsonOblect, String propertyName) {
```
orangejenny, line 14:
```suggestion
    public static String getJsonProperty(String stringifiedJsonObject, String propertyName) {
```

src/main/java/org/commcare/util/GetJsonProperty.java
```diff
@@ -0,0 +1,24 @@
...
+        JSONObject parsedObject = new JSONObject(stringifiedJsonOblect);
+        String value = "";
+        try {
+            value = parsedObject.getString(propertyName);
```
orangejenny, line 18: Is this where an exception would get thrown if the given data isn't valid json or isn't an object?

src/test/java/org/commcare/util/GetJsonPropertyTest.java
```diff
@@ -0,0 +1,18 @@
...
+import org.junit.Test;
+import org.junit.Assert;
+
+import static org.junit.Assert.assertTrue;
+
+public class GetJsonPropertyTest {
+
+    @Test
+    public void getJsonProperty() {
+        String testObj1 = "{name: Sam}";
```
orangejenny, line 17: I'm surprised this allows the object key to not be quoted. Are there any quote-related options in JSONObject? I'm guessing the data coming from web apps will have the keys quoted.

src/test/java/org/commcare/util/GetJsonPropertyTest.java
```diff
@@ -0,0 +1,18 @@
...
+        String testVal1 = GetJsonProperty.getJsonProperty(testObj1, "name");
```
orangejenny, line 18: A couple of negative tests would be nice documentation here: invalid json, a more deeply nested object.

src/main/java/org/commcare/util/GetJsonProperty.java
```diff
@@ -0,0 +1,24 @@
+package org.commcare.util;
```
orangejenny, line 1: This function needs to be available as an xpath expression, not only a java function, so that app builders have access to it. That means moving this into `src.main.java.org.javarosa.xpath.expr` and adding some boilerplate-type code to handle arguments. Check out the other classes in that package for examples, I think `XPathDistanceFunc` is the one most recently added.

src/main/java/org/commcare/util/GetJsonProperty.java
Robert-Costello: thanks

src/main/java/org/commcare/util/GetJsonProperty.java
Robert-Costello: Yeah, it throws a JSONException if the key doesn't exist on the object. Going by the spec it seemed likely that an attempt to get a non-existent property would be made at some point.

src/test/java/org/commcare/util/GetJsonPropertyTest.java
Robert-Costello: I can check on that. I did wonder about it. Here I was going by example is the code.

src/main/java/org/commcare/util/GetJsonProperty.java
Robert-Costello: Ok I'll look at that example, thanks.

Robert-Costello:
Any suggestions on this last stickler issue? I add the <p> tag and it says it's redundant. I take it away and it complains it's not there.

orangejenny:
> Any suggestions on this last stickler issue?

... @shubham1g5 ?

src/test/java/org/commcare/util/GetJsonPropertyTest.java
```diff
@@ -3,9 +3,8 @@
 import org.junit.Assert;
 import org.junit.Test;
 
-/**
+/**Tests for the GetJsonProperty
```
stephherbers, line 6: ha good catch

src/main/java/org/commcare/util/GetJsonProperty.java
Robert-Costello: Right, I think I'll need to handle that in the same try block, since it will also throw an exception. "JSONException - If there is a syntax error in the source string or a duplicated key."

src/test/java/org/commcare/util/GetJsonPropertyTest.java
Robert-Costello: The docs I've found are rather vague. "source - A string beginning with { (left brace) and ending with } (right brace)." I tested it both ways and both work.

src/main/java/org/javarosa/xpath/expr/XPathJsonPropertyFunc.java
```diff
@@ -0,0 +1,48 @@
...
+import org.javarosa.xpath.parser.XPathSyntaxException;
+
+/** Utility for hidden values as geocoder receivers
+ *
+ * @author rcostello
+ * @return A String value for the property name passed in if that property exists else a blank String
+ */
+
+public class XPathJsonPropertyFunc extends XPathFuncExpr {
+    public static final String NAME = "getJsonProperty";
```
snopoke, line 16: We should stick to the 'dash-case' naming convention. I'm also suggesting that we drop the 'get' (similar to 'selected-at')

```suggestion
    public static final String NAME = "json-property";
```

src/test/java/org/javarosa/xpath/expr/test/XPathJsonPropertyFuncTest.java
```diff
@@ -0,0 +1,38 @@
...
+        String testVal3 = XPathJsonPropertyFunc.getJsonProperty(testObj2, "city");
+        String testVal4 = XPathJsonPropertyFunc.getJsonProperty(testObj2, "state");
+        Assert.assertEquals(testVal3, "New York");
+        Assert.assertEquals(testVal4, "");
+
+        String testInvalidObj = "{\"name\"}: \"Sam\"}";
+        try {
+            String testVal5 = XPathJsonPropertyFunc.getJsonProperty(testInvalidObj, "name");  
+        } catch (Exception e) {
+            assertTrue(e instanceof JSONException);
```
snopoke, line 35: Either this should be an `XPathTypeMismatchException` or we should swallow the error and return an empty string.

I'm not completely sure which makes most sense.

src/test/java/org/javarosa/xpath/expr/test/XPathJsonPropertyFuncTest.java
```diff
@@ -0,0 +1,38 @@
...
+        Assert.assertEquals(testVal1, "Sam");
+        Assert.assertEquals(testVal2, "");
+
+        String testObj2 = "{city: New York}";
+        String testVal3 = XPathJsonPropertyFunc.getJsonProperty(testObj2, "city");
+        String testVal4 = XPathJsonPropertyFunc.getJsonProperty(testObj2, "state");
+        Assert.assertEquals(testVal3, "New York");
+        Assert.assertEquals(testVal4, "");
+
+        String testInvalidObj = "{\"name\"}: \"Sam\"}";
```
snopoke, line 31: what happens if you pass in an empty string?

Robert-Costello: added a check/assertion for that scenario.

Robert-Costello:
@snopoke I'm going to ignore the remaining 2 stickler issues that seem to go against the conventions in the code. Lots of other files and classes have two consecutive capital letters, as in "Xpath...".

PR #1124, opened 2022-06-21
Changes virtual instance source reference

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-2083

 Changes source refererences for virtual instance connectors to `jr://instance/<instance-schema>/<instance-id>`

https://github.com/dimagi/formplayer/pull/1159


shubham1g5:
@snopoke It seems to me that these are all the changes needed to make it work ? This seems like a much smaller piece of work to me than originally anticipated so I have a feeling that I might be missing something here.

snopoke (commented):
I think it would be good to update creation of the instances to use the correct ref and also update `addDataInstanceSource` method to store the instances based off the ref and not the instance ID.

That would make the storage and loading more consistent. What do you think?

shubham1g5:
> addDataInstanceSource method to store the instances based off the ref and not the instance ID.

@snopoke It seems like to do that we will also need to change the remote instance reference `jr://instance/remote` to follow the same pattern of jr://instance/remote/<instance-id> as otherwise if there are ever 2 remote instances in the same session they will conflict with each other. I think that's a good change to make give the remote instances are quite similar to virtual instances except that they are persisted differently. Thoughts ?

src/main/java/org/commcare/core/process/CommCareInstanceInitializer.java
```diff
@@ -108,7 +109,7 @@ public InstanceRoot generateRoot(ExternalDataInstance instance) {
      */
     protected InstanceRoot setupExternalDataInstance(ExternalDataInstance instance, String reference,
             String stepType) {
-        String referenceId = getRefId(reference);
+        String referenceId = VirtualInstances.getReferenceId(reference);
         InstanceRoot instanceRoot = getExternalDataInstanceSource(instance, referenceId, stepType);
```
snopoke, line 113: I don't think this will ever return null. I think the last line in `getExternalDataInstanceSource` should be changed to `return null` and this function should change to:

```java
if (instanceRoot != null) {
    return instanceRoot;
}
return instance.getSource() == null ? ConcreteInstanceRoot.NULL : instance.getSource();

File: src/main/java/org/commcare/suite/model/StackFrameStep.java [hunk 2]
```diff
@@ -104,26 +105,27 @@ public String getValue() {
     public void addDataInstanceSource(ExternalDataInstanceSource source) {
         Objects.requireNonNull(source, String.format(
                 "Unable to add null instance data source to stack frame step '%s'", getId()));
-        String instanceId = source.getInstanceId();
-        if (dataInstanceSources.containsKey(instanceId)) {
+        String reference = source.getReference();
+        String referenceId = VirtualInstances.getReferenceId(reference);
+        if (dataInstanceSources.containsKey(referenceId)) {
```
snopoke, line 110: I think this is OK for now but in future we'll want to use the full ref in case there are 2 instances with the same refID:
```
jr://instance/remote/results
jr://instance/search-input/results
```

snopoke (approved):

snopoke (approved):
Nice, thanks for the updates

PR #1127, opened 2022-06-22
stack observer

snopoke:
stack observer to provide visibility into the changes resulting from stack ops

companion to https://github.com/dimagi/formplayer/pull/1162

### DIFF
--- a/src/main/java/org/commcare/session/SessionFrame.java
+++ b/src/main/java/org/commcare/session/SessionFrame.java
@@ -136,15 +136,19 @@ public class SessionFrame implements Externalizable {
         return recentPop;
     }
 
-    protected boolean rewindToMarkAndSet(StackFrameStep step, EvaluationContext evalContext) {
+    protected boolean rewindToMarkAndSet(StackFrameStep step, EvaluationContext evalContext,
+            StackObserver observer) {
         int markIndex = getLatestMarkPosition(steps);
 
         if (markIndex >= 0) {
             String markDatumId = steps.get(markIndex).getId();
+            observer.stepsRewound(steps.subList(markIndex, steps.size()));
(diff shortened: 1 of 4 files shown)


src/main/java/org/commcare/session/SessionFrame.java
```diff
@@ -136,15 +136,19 @@ public StackFrameStep popStep() {
...
 
         if (markIndex >= 0) {
             String markDatumId = steps.get(markIndex).getId();
+            observer.stepsRewound(steps.subList(markIndex, steps.size()));
             steps = new Vector<>(steps.subList(0, markIndex));
             if (step.getValue() != null) {
                 String evaluatedStepValue = step.evaluateValue(evalContext);
-                steps.addElement(new StackFrameStep(SessionFrame.STATE_UNKNOWN, markDatumId, evaluatedStepValue));
+                StackFrameStep rewindStep = new StackFrameStep(SessionFrame.STATE_UNKNOWN, markDatumId, evaluatedStepValue);
+                steps.addElement(rewindStep);
```
orangejenny, line 150: Noticed this `addElement` is the same as calling `pushStep` - should the `observer.stepPushed` move into `pushStep`, and this code call that function instead of directly updating `steps`?

orangejenny: Similar question about `popStep` - should it be using an observer? Or should it not, because we only need to observe particular code paths?

snopoke: This only instruments changes that happen as a result of stack operations. It could be expanded but we don't have a use case for that and generally those events have clearer triggers outside the session. The changes due to stack ops are mostly invisible from outside.

orangejenny (approved):

shubham1g5 (approved):

PR #1130, opened 2022-06-24
just in time virtual instances

snopoke:
Builds on https://github.com/dimagi/commcare-core/pull/1124

This allows instances to be added to an existing evaluation context with the correct ID after the eval context is initialized. The use case here is for dynamic instances like 'search-input'. They are not present in the session when the evaluation context is created so must be added just in time.

In order make sure the ID's are correct we need to clone the instance root with the new ID and then inject it into the matching 'null' instance.

This also updates the 'search-input' ref to remove redundant 'search-input' prefix:
```diff
- jr://instance/search-input/search-input:patients
+ jr://instance/search-input/patients
```


src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -748,17 +755,56 @@ public EvaluationContext spawnWithCleanLifecycle() {
...
+                for (ExternalDataInstance existing : byRef.get(ref)) {
+                    if (existing.getRoot() == null) {
+                        // just in time initializing of the instance
+                        String instanceId = existing.getInstanceId();
+                        TreeElement root = (TreeElement)instance.getRoot();
+                        if (instanceId != name) {
+                            root = TreeUtilities.renameInstance(root, instanceId);
+                        }
+                        root.setParent(existing.getBase());
+                        existing.copyFromSource(new ConcreteInstanceRoot(root));
```
snopoke, line 788: I'm a little unsure if this is the best way to accomplish this but it works.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -748,17 +755,56 @@ public EvaluationContext spawnWithCleanLifecycle() {
...
+                    }
+                }
+            }
+            if (!formInstances.containsKey(name) || formInstances.get(name).getRoot() == null) {
+                // instance name is the same so no need to rename it
+                formInstances.put(name, instance);
```
shubham1g5, line 794: should we skip writing to `formInstances` in case there is already an instance in this table with the matching reference to `instance` ? Since one of the conceptual point we are pushing with this redesign is that a reference should uniquely identify the instance, it seems a bit contradictory to me to have multiple instances with same reference but different ids in eval context.

shubham1g5 (commented):
Thanks for taking this up and driving this to completion! Left one comment that I am a bit unsure about with this design.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -748,17 +755,56 @@ public EvaluationContext spawnWithCleanLifecycle() {
...
+            String ref = instance.getReference();
+            if (!byRef.containsKey(ref)) {
+                if (formInstances.containsKey(name)) {
+                    throw new RuntimeException(String.format(
+                            "EvaluationContext already contains an instance with ID %s with a different ref", name));
+                }
+                formInstances.put(name, instance);
+            } else {
+                for (ExternalDataInstance existing : byRef.get(ref)) {
+                    if (existing.getRoot() == null) {
```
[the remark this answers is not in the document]
snopoke: Interesting, do you have reproduction steps for that?

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
snopoke: This loop only runs after we've already checked against the refs. This is actually a temporary measure to for the existing apps that use the legacy instance names. Once all apps are declaring instances correctly we could remove this entirely.

In the mean time I think this is still safe since we are only replacing instances if they are not already in the context OR they have a null root. I don't think the 2nd condition will actually ever happen so I could look into removing that.

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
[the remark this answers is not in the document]
shubham1g5: Added them in https://github.com/dimagi/commcare-core/pull/1132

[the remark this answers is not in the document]
snopoke: Resolved: https://github.com/dimagi/commcare-core/pull/1132#issuecomment-1168374785

src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
shubham1g5: that makes sense to me and seems good to go without any further changes. Might make sense to add a comment although to remove the handling around multiple instances once the current apps are all migrated to new source references.

shubham1g5 (approved):

shubham1g5 (commented):
Ahh looks like there are other test failures due to the eval context change!

snopoke:
> Ahh looks like there are other test failures due to the eval context change!

:( it looks like there are places that rely on the shallow copy

PR #1131, opened 2022-06-27
Evalulate itemsets for case search fields using search input instance

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-1776

### DIFF
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -162,13 +162,8 @@ public class RemoteQuerySessionManager {
     }
 
     public void populateItemSetChoices(QueryPrompt queryPrompt) {
-        EvaluationContext evalContextWithAnswers = evaluationContext.spawnWithCleanLifecycle();
-        Map<String, String> userQueryValues = getUserQueryValues(true);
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -162,13 +162,8 @@ public static String evalXpathExpression(XPathExpression expr,
     }
 
     public void populateItemSetChoices(QueryPrompt queryPrompt) {
-        EvaluationContext evalContextWithAnswers = evaluationContext.spawnWithCleanLifecycle();
-        Map<String, String> userQueryValues = getUserQueryValues(true);
-        userQueryValues.forEach((promptId, value) -> {
-            evalContextWithAnswers.setVariable(promptId, userAnswers.get(promptId));
-        });
```
snopoke, line 169: I think we'll need to preserve this for existing apps right?

shubham1g5: Jenny actually did [some investigation](https://dimagi-dev.atlassian.net/browse/USH-1776?focusedCommentId=211489) on this and found that no apps are currently using the existing variable reference syntax.

snopoke (approved):

PR #1132, opened 2022-06-28
Tests updating input updates the search input instance

shubham1g5:


snopoke (commented):
In this test the same evaluation context is used for both set's of input. Does that ever happen when running the platform? I don't think we ever reuse an evaluation context across multiple user interactions like this.

snopoke:
> In this test the same evaluation context is used for both set's of input. Does that ever happen when running the platform? I don't think we ever reuse an evaluation context across multiple user interactions like this.

While I still think this is true but there was an issue with changes to a spawned evaluation context propagating back to the original evaluation context.

snopoke (approved):

PR #1135, opened 2022-07-01
fix loading of legacy search-input instances

snopoke:
The search-input instance ref is not `base/[instanceId]` so it requires special handling.


snopoke:
Tested on staging. See https://github.com/dimagi/formplayer/pull/1176#issuecomment-1172246562

shubham1g5 (approved):

PR #1138, opened 2022-07-04
Case search validation Conditions

shubham1g5:
Adds support for validation condition on case search prompts. 

Xml Spec:

````
<prompt key="name">
    <validation test=”xpath_condition"”>       
                <text>
                    <locale id="locale.id"/>
                </text>
    </validation>

    <display>
        <text>
            <locale id="search_property.m0.name"/>
        </text>
    </display>
</prompt>
````

[Design Doc](https://docs.google.com/document/d/1wF9SFfX7-R86cPRgabrqbTU9BaGdp3IlAYlTjInWXTU/edit#)

https://github.com/dimagi/formplayer/pull/1177

### DIFF
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -8,7 +8,6 @@ import org.commcare.cases.util.StringUtils;
 import org.commcare.data.xml.VirtualInstances;
 import org.commcare.suite.model.QueryData;
 import org.commcare.suite.model.QueryPrompt;
-import org.commcare.suite.model.QueryPromptValidation;
 import org.commcare.suite.model.RemoteQueryDatum;
 import org.commcare.suite.model.SessionDatum;
 import org.javarosa.core.model.ItemsetBinding;
@@ -252,8 +251,9 @@ public class RemoteQuerySessionManager {
         for (Enumeration en = userInputDisplays.keys(); en.hasMoreElements(); ) {
             String key = (String)en.nextElement();
             QueryPrompt queryPrompt = userInputDisplays.get(key);
+            String value = userAnswers.get(key);
             TreeReference currentRef = getReferenceToInstanceNode(instanceId, key);
-            if (queryPrompt.isInvalidInput(new EvaluationContext(ec, currentRef))) {
+            if (value != null && queryPrompt.isInvalidInput(new EvaluationContext(ec, currentRef))) {
                 errors.put(key, queryPrompt.getValidationMessage(ec));
             }
         }
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -254,9 +254,8 @@ public class RemoteQuerySessionManager {
             QueryPrompt queryPrompt = userInputDisplays.get(key);
             QueryPromptValidation validation = queryPrompt.getValidation();
             TreeReference currentRef = getReferenceToInstanceNode(instanceId, key);
-            if (validation != null && !((Boolean)validation.getTest().eval(
-                    new EvaluationContext(ec, currentRef)))) {
-                errors.put(key, validation.getMessage().evaluate(ec));
+            if (queryPrompt.isInvalidInput(new EvaluationContext(ec, currentRef))) {
+                errors.put(key, queryPrompt.getValidationMessage(ec));
             }
         }
     }
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -231,8 +231,13 @@ public class RemoteQuerySessionManager {
         }
     }
 
+    // Recalculates screen properties that are dependent on user input
+    public void refresh() {
+        refreshItemSetChoices();
+        validateUserAnswers();
+    }
 
-    public void validateUserAnswers() {
+    private void validateUserAnswers() {
         OrderedHashtable<String, QueryPrompt> userInputDisplays = getNeededUserInputDisplays();
         EvaluationContext ec = getEvaluationContextWithUserInputInstance();
         for (Enumeration en = userInputDisplays.keys(); en.hasMoreElements(); ) {
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -239,7 +239,7 @@ public class RemoteQuerySessionManager {
     }
 
     // Recalculates screen properties that are dependent on user input
-    public void refresh() {
+    public void refreshInputDependentState() {
         refreshItemSetChoices();
         validateUserAnswers();
     }
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -140,7 +140,7 @@ public class RemoteQuerySessionManager {
                 QueryPrompt prompt = queryDatum.getUserQueryPrompts().get(key);
                 XPathExpression excludeExpr = prompt.getExclude();
                 if (!(params.containsKey(key) && params.get(key).contains(value))) {
-                    if (value != null && (excludeExpr == null || !(boolean) excludeExpr.eval(evaluationContext))) {
+                    if (value != null && (excludeExpr == null || !(boolean)excludeExpr.eval(evaluationContext))) {
                         params.put(key, userAnswers.get(key));
                     }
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/main/java/org/commcare/suite/model/QueryPromptValidation.java [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+import org.javarosa.xpath.expr.XPathExpression;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+
+public class QueryPromptValidation implements Externalizable {
+
+    private XPathExpression xpath;
+    private String message;
```
snopoke, line 17: should this not be a `Text` attribute to support translation?

Also commented on the spec: https://docs.google.com/document/d/1wF9SFfX7-R86cPRgabrqbTU9BaGdp3IlAYlTjInWXTU/edit?disco=epp37nri35q

src/main/java/org/commcare/xml/QueryPromptParser.java
```diff
@@ -34,6 +36,8 @@ public class QueryPromptParser extends CommCareElementParser<QueryPrompt> {
     private static final String ATTR_ALLOW_BLANK_VALUE = "allow_blank_value";
     private static final String ATTR_EXCLUDE = "exclude";
     private static final String ATTR_REQUIRED = "required";
+    private static final String ATTR_VALIDATION_XPATH = "xpath";
```
snopoke, line 39: what about calling this 'test' to align with 'assertions'?

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -226,6 +227,20 @@ public void refreshItemSetChoices() {
...
 
+
+    public void validateUserAnswers() {
+        OrderedHashtable<String, QueryPrompt> userInputDisplays = getNeededUserInputDisplays();
+        EvaluationContext ec = getEvaluationContextWithUserInputInstance();
+        for (Enumeration en = userInputDisplays.keys(); en.hasMoreElements(); ) {
+            String key = (String)en.nextElement();
+            QueryPrompt queryPrompt = userInputDisplays.get(key);
+            QueryPromptValidation validation = queryPrompt.getValidation();
+            if (validation != null && !((Boolean)validation.getXpath().eval(ec))) {
```
snopoke, line 247: can the evaluation context be re-scoped for each of these to allow referencing the current prompt value with `.` or `current()`?

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -226,6 +227,20 @@ public void refreshItemSetChoices() {
         }
     }
 
+
+    public void validateUserAnswers() {
```
snopoke: `errors` should be cleared at the start of this method

snopoke: a test for this would be good to ensure that errors are resolved after user input is corrected

src/main/java/org/commcare/xml/QueryPromptParser.java
```diff
@@ -55,15 +61,41 @@ public QueryPrompt parse() throws InvalidStructureException, IOException, XmlPul
...
         return new QueryPrompt(key, appearance, input, receive, hidden, display,
-                               itemsetBinding, defaultValue, allowBlankValue, exclude, required);
+                itemsetBinding, defaultValue, allowBlankValue, exclude, required, validation);
+    }
+
+    private QueryPromptValidation parseValidationBlock(String key)
+            throws InvalidStructureException, XmlPullParserException, IOException {
+        String testStr = parser.getAttributeValue(null, ATTR_VALIDATION_TEST);
+        if (testStr == null) {
+            throw new InvalidStructureException("No test condition defined in validation for key " + key);
```
snopoke, line 82: nit: wording suggestion
```suggestion
            throw new InvalidStructureException("No test condition defined in validation for prompt: " + key);
```

src/main/java/org/commcare/xml/QueryPromptParser.java
```diff
@@ -55,15 +61,41 @@ public QueryPrompt parse() throws InvalidStructureException, IOException, XmlPul
...
+        }
+        XPathExpression test = xpathPropertyValue(testStr);
+        Text message = null;
+        while (nextTagInBlock(NAME_VALIDATION)) {
+            if (parser.getName().equals(NAME_TEXT)) {
+                message = new TextParser(parser).parse();
+            } else {
+                throw new InvalidStructureException(
+                        "Unrecognised node " + parser.getName() + "in validation for key " + key);
```
snopoke, line 91:
```suggestion
                        "Unrecognised node " + parser.getName() + "in validation for prompt: " + key);
```

src/main/java/org/commcare/xml/QueryPromptParser.java
```diff
@@ -55,15 +61,41 @@ public QueryPrompt parse() throws InvalidStructureException, IOException, XmlPul
...
+            }
+        }
+        if (message == null) {
+            throw new InvalidStructureException(
+                    "No validation message defined in the validation block for key " + key);
```
snopoke, line 96:
```suggestion
                    "No validation message defined in the validation block for prompt: " + key);
```

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -226,6 +238,34 @@ public void refreshItemSetChoices() {
         }
     }
 
+    // Recalculates screen properties that are dependent on user input
+    public void refresh() {
```
snopoke, line 242: nit: naming suggestion. Typically `refresh` on it's own implies refreshing the full model but here we are only re-computing some dependent state. 
```suggestion
    public void refreshDependentState() {
```

Not a great name so feel free to adjust.

snopoke (approved):

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -254,9 +254,8 @@ private void validateUserAnswers() {
             QueryPrompt queryPrompt = userInputDisplays.get(key);
             QueryPromptValidation validation = queryPrompt.getValidation();
```
snopoke, line 255: unused var

src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -181,13 +184,18 @@ public boolean isSelect() {
...
+     * @return evaluated validation message or a default text if no validation message is defined
      */
     public String getValidationMessage(EvaluationContext ec) {
         if (validation != null && validation.getMessage() != null) {
             return validation.getMessage().evaluate(ec);
         }
-        return "";
+
+        try {
+            return Localization.get("invalid_answer_error");
```
snopoke, line 195: nit: more consistent naming:
```suggestion
            return Localization.get("case.search.answer.invalid");
```

src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -24,6 +26,7 @@ public class QueryPrompt implements Externalizable {
     public static final String INPUT_TYPE_DATERANGE = "daterange";
     public static final String INPUT_TYPE_DATE = "date";
     public static final String INPUT_TYPE_ADDRESS = "address";
+    public static final String DEFAULT_VALIDATION_ERROR = "This answer is outside the allowed range";
```
snopoke, line 29: nit: more generic message (it may not be a range type, also user's don't think of these as 'answers'): 
```suggestion
    public static final String DEFAULT_VALIDATION_ERROR = "Input validation failed";
```

src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -156,11 +168,43 @@ public XPathExpression getRequired() {
...
+     * @param ec eval context to evaluate the validation message
+     * @return evaluated validation message or a default text if no validation message is defined
+     */
+    public String getValidationMessage(EvaluationContext ec) {
+        if (validation != null && validation.getMessage() != null) {
+            return validation.getMessage().evaluate(ec);
+        }
+
+        try {
+            return Localization.get("case.search.answer.invalid");
```
snopoke, line 195: how come we are providing a default here but not for the required condition message: https://github.com/dimagi/commcare-core/pull/1139/files#diff-53ece50c900e231fe4323b42c57451b8b09b66d48f43cdf213dc08587c0a5bbdR194

shubham1g5: Mostly because HQ already has a default implemented for required messages but agree that FP should supply a default message for required condition as well. I will add it soon.

snopoke (approved):

PR #1139, opened 2022-07-11
Dynamic required Conditions + Custom Required Messaging

shubham1g5:
[xml Spec](https://docs.google.com/document/d/1-POkDP4J2YDzrbwP9R4hWtk99SAwX-qQlEPVDQ1UjPA/edit#heading=h.23qs7gt1bu1q)

https://github.com/dimagi/formplayer/pull/1180

### DIFF
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -257,14 +257,11 @@ public class RemoteQuerySessionManager {
         for (Enumeration en = userInputDisplays.keys(); en.hasMoreElements(); ) {
             String key = (String)en.nextElement();
             QueryPrompt queryPrompt = userInputDisplays.get(key);
-            XPathExpression oldRequired = queryPrompt.getOldRequired();
             XPathExpression requiredCondition =
                     queryPrompt.getRequired() != null ? queryPrompt.getRequired().getTest() : null;
             boolean isRequired = false;
             if (requiredCondition != null) {
                 isRequired = (Boolean)requiredCondition.eval(ec);
-            } else if (oldRequired != null) {
-                isRequired = (Boolean)oldRequired.eval(ec);
             }
             requiredPrompts.put(key, isRequired);
         }
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -55,6 +55,7 @@ public class RemoteQuerySessionManager {
         this.evaluationContext = evaluationContext;
         this.supportedPrompts = supportedPrompts;
         initUserAnswers();
+        refreshInputDependentState();
     }
 
     private void initUserAnswers() throws XPathException {
@@ -67,7 +68,6 @@ public class RemoteQuerySessionManager {
                 userAnswers.put(prompt.getKey(),
                         FunctionUtils.toString(prompt.getDefaultValueExpr().eval(evaluationContext)));
             }
-
         }
     }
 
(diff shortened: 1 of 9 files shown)


orangejenny (approved):

src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java
```diff
@@ -312,4 +312,49 @@ private RemoteQuerySessionManager testErrorsWithUserInput(Map<String, String> us
 
         return remoteQuerySessionManager;
     }
+
+    @Test
+    public void testRequiredWithUserInput_dependentConditions() throws Exception {
+        // when age, dob is not required
+        testRequiredWithUserInput(
+                ImmutableMap.of("age", "15"),
+                ImmutableMap.of("age", true, "dob", false),
```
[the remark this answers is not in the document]
shubham1g5: agree! @snopoke deserve the credit for introducing these in the codebase.

src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -158,12 +164,18 @@ public XPathExpression getExclude() {
         return exclude;
     }
 
-    public XPathExpression getRequired() {
+    @Nullable
```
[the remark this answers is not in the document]
shubham1g5: It's not compulsory but is more of a Java best practice. The annotation is compile time only and helps IDE in giving out warnings for eg. when you don't perform a null check before accessing a nullable field.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -242,6 +247,27 @@ public void refreshItemSetChoices() {
...
+            QueryPrompt queryPrompt = userInputDisplays.get(key);
+            XPathExpression oldRequired = queryPrompt.getOldRequired();
+            XPathExpression requiredCondition =
+                    queryPrompt.getRequired() != null ? queryPrompt.getRequired().getTest() : null;
+            boolean isRequired = false;
+            if (requiredCondition != null) {
+                isRequired = (Boolean)requiredCondition.eval(ec);
+            } else if (oldRequired != null) {
+                isRequired = (Boolean)oldRequired.eval(ec);
+            }
```
snopoke, line 268: what do you think about pushing this down into the model?
```java
boolean isRequired = queryPrompt.isRequired(ec);
```

src/test/resources/case_claim_example/suite.xml
```diff
@@ -102,6 +107,18 @@
             </text>
           </display>
         </prompt>
+        <prompt key="dob" required="false()">
```
snopoke, line 110: Should having both the attribute and the child element be allowed? Seems like it could be confusing.

src/main/java/org/commcare/suite/model/QueryPrompt.java
```diff
@@ -174,4 +176,21 @@ public boolean isSelect() {
         return getItemsetBinding() != null;
     }
 
+    // Evaluates required in the given eval context
+    public boolean isRequired(EvaluationContext ec) {
+        XPathExpression requiredCondition = required != null ? required.getTest() : null;
+        if (requiredCondition != null) {
+            return (Boolean)requiredCondition.eval(ec);
+        }
```
snopoke, line 184: nit: perhaps just preference but I prefer the way you coded it for the `getRequiredMessage` function
```suggestion
        if (required != null && required.getTest() != null) {
            return (Boolean) required.getTest().eval(ec);
        }
```

snopoke (approved):

shubham1g5:
[Fix for QA-4414]( https://github.com/dimagi/commcare-core/pull/1139/commits/59efa3498a40842f875c3f05339689d84fdf72ed )

PR #1140, opened 2022-07-25
Parses Android dependencies

shubham1g5:
Adds support for follwing definiton in profile file - 

````
<features>
        <dependencies active="true">
            <android_package id ="org.commcare.test" name=”Test”/>
            <android_package id ="org.commcare.reminders" name=”Reminders”/>
        </dependencies
</features>
````

Easier to review the PR as a whole. 

cross_request: https://github.com/dimagi/commcare-android/pull/2569

### DIFF
--- a/src/main/java/org/commcare/suite/model/AndroidPackageDependency.java
+++ b/src/main/java/org/commcare/suite/model/AndroidPackageDependency.java
@@ -13,9 +13,6 @@ public class AndroidPackageDependency implements Externalizable {
     private String id;
     private String name;
 
-    // Set later during dependency check
-    private boolean installed = false;
-
     /**
      * Serialization Only!!!
      */
@@ -55,13 +52,5 @@ public class AndroidPackageDependency implements Externalizable {
                 "id='" + id + '\'' +
                 ", name='" + name + '\'' +
                 '}';
-    }
-
-    public boolean isInstalled() {
-        return installed;
-    }
-
-    public void setInstalled(boolean installed) {
-        this.installed = installed;
     }
 }
--- a/src/main/java/org/commcare/suite/model/AndroidPackageDependency.java
+++ b/src/main/java/org/commcare/suite/model/AndroidPackageDependency.java
@@ -9,7 +9,7 @@ import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
 
-public class AppDependency implements Externalizable {
+public class AndroidPackageDependency implements Externalizable {
     private String id;
     private String name;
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by shubham1g5** [comment]
@ctsims Bumping this for review.

**Review by ctsims — COMMENTED** [review]
Hey @shubham1g5 - Overall code looks good to me but I realized something I missed during spec review.

The "name" for the dependency feels a bit odd to me. Forcing the user to type out a name seems like a little bit of a burden for a string that won't actually show up in most apps currently (in the spec we discussed you'd only see it when there are multiple apps), and it's weird that it's only in one language since the string only shows up after the app is "Seated" and translations are available. 

I wonder whether it might make more sense to do something like look for a localization key with a hardcoded prefix on the package name, like

```
android.package.${packageid}
```

and display that when we need to describe the package, falling back to displaying the package name itself if no translation exists. A couple subtle advantages that has:

1) It doesn't force users to enter text by default, so if we come up with a better automated solution for deriving the package name later we won't need to deal with conflicts between that name and a user entered name
2) It'll let HQ more easily share the translations with the existing system
3) If we wanted users to manually enter a package name on the HQ UI they still could, it would just 'compile down' to a different place

shubham1g5:
That's a good suggestion on supporting translations. Added it as part of spec [here](https://docs.google.com/document/d/1F_nKK8dSvpEKB3QOqq2A8Rb1AcMKZwpg9_TLK_IrhJ4/edit#heading=h.h6lt4a2ntiky). Also removed `name` from the `android_package` xml node in [here](https://docs.google.com/document/d/1F_nKK8dSvpEKB3QOqq2A8Rb1AcMKZwpg9_TLK_IrhJ4/edit#heading=h.wnwyziw0uedb). I will followup soon with the changes on the PR as well. 

> in the spec we discussed you'd only see it when there are multiple apps

For clarity I ended up on showing the app name even for a single app as seen in the first screenshot over [here](https://github.com/dimagi/commcare-android/pull/2569). 

> It doesn't force users to enter text by default

Another clarification that with current design HQ will pre-poulate a super set of dependencies that app builders can choose from to add dependency for their respective CC apps.

shubham1g5:
> I will followup soon with the changes on the PR as well.

https://github.com/dimagi/commcare-core/pull/1140/commits/56b95cb0c6c1adff339807d670fc8da734acc57f and https://github.com/dimagi/commcare-android/pull/2569/commits/05ed9d1afdb560c0b261d9cc734f665ac02fa0f0

ctsims (approved):
thanks for the feedback and adjustments!

PR #1142, opened 2022-07-28
Fixes Dependent Itemsets Selection in Case search

shubham1g5:
https://dimagi-dev.atlassian.net/browse/USH-2267

Issue was that we were doing the [selection index to itemset choice conversion](https://github.com/dimagi/commcare-core/blob/formplayer/src/cli/java/org/commcare/util/screen/QueryScreen.java#L202-L216) before updating the itemset choices for the select prompt. This caused the selection to not be valid for the new inputs and ultimately removed in [refreshItemSetChoices](https://github.com/dimagi/commcare-core/blob/formplayer/src/main/java/org/commcare/session/RemoteQuerySessionManager.java#L232)

This PR surfaces(not a regression) another funny behaviour which is ticketed [here](https://dimagi-dev.atlassian.net/browse/USH-2289)

https://github.com/dimagi/formplayer/pull/1195

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -199,6 +199,7 @@ public class QueryScreen extends Screen {
             // If select question, we should have got an index as the answer which should
             // be converted to the corresponding value
             if (queryPrompt.isSelect() && !StringUtils.isEmpty(answer)) {
+                remoteQuerySessionManager.populateItemSetChoices(queryPrompt);
                 Vector<SelectChoice> selectChoices = queryPrompt.getItemsetBinding().getChoices();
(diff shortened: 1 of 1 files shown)


snopoke (approved):
Just wondering if this makes the call to `populateItemSetChoices` in [refreshItemSetChoices](https://github.com/dimagi/commcare-core/blob/ff368eec6088e6fcaeb262ee70dae9264836b80c/src/main/java/org/commcare/session/RemoteQuerySessionManager.java#L217) redundant in some cases? Would it make sense to check if the choices are already populated?

shubham1g5:
@snopoke a select prompt can also depend on another case search prompt that comes later in order on UI. In that case we will need to repopulate itemset choices again after all the prompts have been answered which is why I think we still need to call `populateItemSetChoices` in `refreshItemSetChoices`

PR #1148, opened 2022-08-24
QueryScreen.init title translation change for Master

dominicochoa-dimagi:
## Technical Summary

[Jira](https://dimagi-dev.atlassian.net/browse/USH-2271)

### Safety story

This code change is limited to the query screen code. There is logic to handle errors such that it defaults to the original behavior. 

This is the follow up to the formplayer PR [here](https://github.com/dimagi/commcare-core/pull/1110)

### Rollback instructions

### Review


shubham1g5 (changes requested):
Code looks good, can we replicate the test here as well from the `formplayer` PR if possible ? Otherwise we can just add the new `title` xml in the app for [this test](https://github.com/dimagi/commcare-core/blob/master/src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java) so that we know that parsing apps with `title` don't error.

shubham1g5 (approved):
thanks!

PR #1149, opened 2022-09-12
Added csv support to multi-select functions

orangejenny:
## Product Description
https://dimagi-dev.atlassian.net/browse/USH-2437

## Technical Summary
These are already based on spaces, but they don't support the quote behavior needed for multi-word options.

Interdependent with https://github.com/dimagi/formplayer/pull/1222

### Safety story
Automated test coverage here is good.

### Automated test coverage
In PR

### QA Plan
Not requesting QA

### Rollback instructions
If this is rolled back, https://github.com/dimagi/formplayer/pull/1222 also needs to be rolled back.

### Review

### DIFF
--- a/src/main/java/org/javarosa/core/util/DataUtil.java
+++ b/src/main/java/org/javarosa/core/util/DataUtil.java
@@ -64,7 +64,43 @@ public class DataUtil {
         if ("".equals(s)) {
             return new String[0];
         }
-        return s.split("[ ]+");
+
+        Vector<String> items = new Vector<String>();
+        StringBuilder item = new StringBuilder();
+        char quote = '\u0000';
+
+        for (int i = 0; i < s.length(); i++) {
+            char current = s.charAt(i);
+            if (current == ' ' && quote == '\u0000') {
+                if (i == 0 || s.charAt(i - 1) != ' ') {     // allow to delimit with multiple spaces
+                    items.add(item.toString());
+                    item = new StringBuilder();
+                }
+            } else if (current == quote && s.charAt(i - 1) != '\\') {
+                quote = '\u0000';
+            } else if (quote == '\u0000' && (current == '\'' || current == '"') && (i == 0 || s.charAt(i - 1) != '\\')) {
+                quote = current;
+            } else {
+                item.append(current);
+            }
+        }
+
+        if (item.length() > 0) {
+            items.add(item.toString());
+        }
+
+        return items.toArray(new String[0]);
+    }
+
+    public static String joinWithSpaces(String[] items) {
+        int index = 0;
+        for (String item : items) {
+            if (item.indexOf(" ") != -1) {
+                items[index] = "\"" + item.replace("\"", "\\\"") + "\"";
+            }
+            index++;
+        }
+        return String.join(" ", items);
     }
 
     public static boolean intArrayContains(int[] source, int target) {
--- a/src/main/java/org/javarosa/core/util/DataUtil.java
+++ b/src/main/java/org/javarosa/core/util/DataUtil.java
@@ -62,7 +62,7 @@ public class DataUtil {
 
     public static String[] splitOnSpaces(String s) {
         if ("".equals(s)) {
-            return new String[0];
+            return new String[]{""};
         }
 
         Vector<String> items = new Vector<String>();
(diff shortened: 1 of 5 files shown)


src/test/java/org/javarosa/core/util/test/DataUtilTest.java
```diff
@@ -41,4 +43,23 @@ public void intersectionTest() {
         assertFalse(intersectSet.contains("two"));
         assertFalse(intersectSet.contains("three"));
     }
+
+    @Test
+    public void splitOnSpacesTest() {
```
shubham1g5, line 48: nit: would be good to add a test case demonstrating what happens when multiple spaces are added consecutively in between words.

shubham1g5 (commented):
A bit of late realisation that this code does have the potential to change the existing functionality for multi-select functions and might cause unexpected behaviour for existing apps. 

I think to protect against that we should introduce a flag in the arguments for this function that enables the CSV parse and let the default behaviour be unchanged. Thoughts ?

src/main/java/org/javarosa/core/util/DataUtil.java
```diff
@@ -64,7 +64,43 @@ public static String[] splitOnSpaces(String s) {
         if ("".equals(s)) {
             return new String[0];
         }
-        return s.split("[ ]+");
+
+        Vector<String> items = new Vector<String>();
```
snopoke, line 68: I think this could be replaced with:
```java
reader = CsvReader.parse(s)
reader.readRecord()
String[] values = reader.getValues();
```

snopoke (commented):
> A bit of late realisation that this code does have the potential to change the existing functionality for multi-select functions and might cause unexpected behaviour for existing apps.

@shubham1g5 do you mean if a value has quotes in it?
```
selected('texas "new mexico" utah', 1)
// before
>> '"new'
// after
>> 'new mexico'
```
That is true. It will change behaviour and I don't know how likely it to happen. 
Another option would be to create a new function `selected-csv`?

src/main/java/org/javarosa/core/util/DataUtil.java
```diff
@@ -64,7 +64,43 @@ public static String[] splitOnSpaces(String s) {
...
+        }
+
+        if (item.length() > 0) {
+            items.add(item.toString());
+        }
+
+        return items.toArray(new String[0]);
+    }
+
+    public static String joinWithSpaces(String[] items) {
```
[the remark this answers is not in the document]
orangejenny: Formplayer PR wasn't up yet, it's here now: https://github.com/dimagi/formplayer/pull/1222/

src/main/java/org/javarosa/core/util/DataUtil.java
orangejenny: I definitely wrote the javascript version of this first and just copied it here and tweaked it until it compiled. That's much better.

One thing I'm stuck on - `CsvReader` is in a different package(?), it's under `src/translate` instead of `src/main`. I'm having trouble importing it correctly, running into errors with the package not existing. Do I need to modify gradle to make the CsvReader availabl to this code?

orangejenny:
Regarding the behavior change, I see what you mean but it also seems unfortunate to make this a different function when it's basically an enhancement. Still thinking about this.

You're only concerned about web apps, right? Since on mobile it would be reasonable to change the behavior slightly from one CC version to the next?

shubham1g5:
> You're only concerned about web apps, right? Since on mobile it would be reasonable to change the behavior slightly from one CC version to the next?

Even on mobile, we can't do changes like this between CC updates. Typically for anything that changes behaviour of existing functionality, we introduce a new app setting with options say `new` and `old` with default being the `old`.  We then work towards transitioning the default behaviour to the new behaviour over releases by giving projects heads up that we are going to switch the default behaviour to the new one in a coming release and if they wish to be on old behaviour, they need to explicity change the setting `old` in their apps and make a new release. If we want to keep the method signature same and don't introduce a new method that does the same thing in most cases, I think the setting path way  seems most clean and decoupled.

src/main/java/org/javarosa/core/util/DataUtil.java
snopoke: ah yes, that does make it harder. The `CsvReader` class is also surprisingly large. I think let's stick with this for now.

orangejenny:
@shubham1g5 ah, I see. A setting does sound like the cleanest approach. I added a setting to the HQ PR (it's not related to the rest of the PR, but making more PRs will just mean more to keep track of). I'm working on figuring out how to access it inside the xpath function logic.

orangejenny:
Closing, see discussion on ticket.

PR #1153, opened 2022-10-11
Skip evaluation of entity list fields when input is an Action

shubham1g5:
## Technical Summary

https://dimagi-dev.atlassian.net/browse/SUPPORT-14839

We do not need to initialise entity list fields when we are going to launch an Action that will take us away from the current entity screen. 

### Safety story

Did some testing locally and seem to work fine. It should only affect the case search and registation from case list workflows.

### Automated test coverage

Case claim workflows have a good number of tests on Formplayer side which should be sufficient to catch any regressions

### QA Plan

https://dimagi-dev.atlassian.net/browse/QA-4595

### Rollback instructions

We most probably won't need to roll this back, instead rolling back the cross-requested FP PR should be enough which is responsible for activating code in this PR. 

### Review

cross-request: https://github.com/dimagi/formplayer/pull/1244

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityScreen.java
@@ -63,6 +63,9 @@ public class EntityScreen extends CompoundScreenHost {
     private boolean initialized = false;
     private Action autoLaunchAction;
 
+    // Whether the input to the entity screen is going to be an action
+    private boolean inActionMode = false;
+
     public EntityScreen(boolean handleCaseIndex) {
         this.handleCaseIndex = handleCaseIndex;
     }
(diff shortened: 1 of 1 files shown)


snopoke:
> I'm a bit confused as to why we need to do this. My understanding is that the first EntityScreen should be skipped over due to the autolaunch action.
> 
> I'm going to try and set up an app locally to test with.

Using a simple app to test locally (configure module with 'skip to default search results' and then use a 'results' instance expression in a case list property) I think I see the issue:

* for detail requests we set `isDetailScreen` which results in [`needsFullEntityScreen`](https://github.com/dimagi/formplayer/blob/d72c2b6e1adbe62fd68cd5943e655f7ec4d815e6/src/main/java/org/commcare/formplayer/services/MenuSessionRunnerService.java#L274) being set to `True` even if the entity screen isn't the one showing the details.

I *think* that `isDetailScreen` should not result in full init except for the screen where the details is needed.

I can connect with you more on this tomorrow.

shubham1g5:
Offlined with @snopoke and going to go ahead with this fix for now by sending it to QA today. There is scope for more code cleanups + performance improvements here but this feels like the simplest fix to deploy for now and we should revisit this code again soon.

shubham1g5:
Closing in favour of https://github.com/dimagi/commcare-core/pull/1155

PR #1155, opened 2022-10-14
Auto Launch Changes

shubham1g5:
API change and cleanup due to https://github.com/dimagi/formplayer/pull/1251

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
@@ -215,9 +215,6 @@ public class EntityListSubscreen extends Subscreen<EntityScreen> {
                 host.setPendingAction(actions.elementAt(chosenActionIndex));
                 return true;
             }
-        } else if (host.getAutoLaunchAction() != null && allowAutoLaunch) {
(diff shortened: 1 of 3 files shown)


src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
```diff
@@ -215,9 +215,6 @@ public boolean handleInputAndUpdateHost(String input, EntityScreen host, boolean
                 host.setPendingAction(actions.elementAt(chosenActionIndex));
                 return true;
             }
-        } else if (host.getAutoLaunchAction() != null && allowAutoLaunch) {
-            host.setPendingAction(host.getAutoLaunchAction());
-            return true;
```
snopoke, line 220: Can you say more about why this isn't needed anymore.

shubham1g5: We were earlier relying on this to set the pending auto launch action to the Entity Screen through the `menuSession.handleInput` pathway but since we are directly setting it now [in here](https://github.com/dimagi/formplayer/pull/1251/files) instead of calling `handleInput` we don't need it anymore.

snopoke (approved):

PR #1160, opened 2022-11-07
Add dependabot.yml file

avazirna:
## Technical Summary
Dependabot is a GitHub Action that runs on schedule to check the dependencies against upstream sources, and when it finds newer versions, is able to automatically create PRs for the update. This PR is to enable Dependabot version updates.

### Rollback instructions

### Review

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: gradle
+    directory: "/"
+    schedule:
+      interval: daily
+    labels:
+      - dependencies


snopoke:
Do the dependencies here need to be kept in sync with formplayer / commcare-android at all?

shubham1g5:
@snopoke I would think so, furthermore we will need to verify whether the formplayer and commcare-android builds correctly along with tests before merging an update PR.

avazirna:
@snopoke yes, I'm tracking the overlaps in a separate document. Also, any updates to the master branch will be applied to the branch being used by formplayer, so we can keep those in sync too.

shubham1g5 (approved):

PR #1161, opened 2022-11-08
Updated ConvertToEthiopian Function to also use the getFields method

RynhardPietersen:
## Product Description

format-date-for-calendar() function is returning a date that is behind when the selected date format is nepali or ethiopian. This fixes the ethiopian side of that.

## Technical Summary

[Ticket](https://dimagi-dev.atlassian.net/browse/SAAS-13168) describes format-date-for-calendar function not working. Since this only happens depending on which time zone you are currently in I was able to conclude that the problem is that the date sent to formplayer has a timezone attached which triggers formplayer to do a date conversion by subtracting the time offset of the date sent. The getFields function already does timezoneoffset math to counteract this so I decided to use that instead of a calendar object

### Safety story

local testing. 
Specifically tested that the following time offsets:
06:00 - 18:00 
23:59 - 00:59
23:01 - 00:01
The reason for testing these time offsets was to ensure that the returned date was correct regardless of client side and server side date differentials.

### Automated test coverage

The tests already written are supposed to cover this already since they work with time zones but we know the function works as intended since the date being passed into the function is wrong, not the function itself. I'm not sure it is too plausible to write a test that catches this specific problem.

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/javarosa/xform/util/CalendarUtils.java
+++ b/src/main/java/org/javarosa/xform/util/CalendarUtils.java
@@ -14,6 +14,8 @@ import java.util.Calendar;
 import java.util.Date;
 import java.util.HashMap;
 
+import static org.javarosa.core.model.utils.DateUtils.getFields;
+
 public class CalendarUtils {
 
     private static ArrayDataSource arrayDataSource = new LocaleArrayDataSource(
@@ -49,9 +51,8 @@ public class CalendarUtils {
             format = "%e %B %Y";
         }
 
-        Calendar c = Calendar.getInstance();
-        c.setTime(d);
-        return ConvertToEthiopian(c.get(Calendar.YEAR), c.get(Calendar.MONTH) + 1, c.get(Calendar.DAY_OF_MONTH), format);
+        DateUtils.DateFields fields = getFields(d);
+        return ConvertToEthiopian(fields.year, fields.month, fields.day, format);
     }
 
     private static final HashMap<Integer, int[]> NEPALI_YEAR_MONTHS = new HashMap<>();


ctsims (approved):
Hi Rynhard,

Code looks good to me here, but in the safety story can you articulate the battery of local testing you performed that we discussed? It's really important / helpful context if future issues are discovered.

shubham1g5:
>  was able to conclude that the problem is that the date sent to formplayer has a timezone attached which triggers formplayer to do a date conversion by subtracting the time offset of the date sent. The getFields function already does timezoneoffset math to counteract this

I looked into this more and the exact issue seem to be the mismatch between browser timezones and the FP server timezone. For eg. If we select Nov 10 in the browser which is set to GMT +5 and suppose FP machine is running the timezone GMT +4, the date gets converted to Nov 9 23:00 GMT +4 on Formplayer. The getFields() method fixes this since it sets the [timezone supplied by the browser](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/utils/DateUtils.java#L145) to the calendar first. 

A similar strategy should also fix the Nepali widget.  Currently the [date conversion here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/xform/util/CalendarUtils.java#L248) is utilising the local/default timezone on FP server. If we change this to pass the browser timezone from `DateUtils`, we should have the right date for Neplai widget too. It might make sense to move the whole `fromMillis` method from `CalendarUtils` to `DateUtils` in that case and probably rename this method to `getUniversalDateFromMillis()` to be more clear.

shubham1g5 (approved):

PR #1164, opened 2022-11-14
Case search subtitle/description text

nospame:
## Technical Summary
Adds description text to QueryScreen and RemoteQueryDatum following the example of title label.
Required for https://github.com/dimagi/commcare-hq/pull/32338, https://github.com/dimagi/formplayer/pull/1267

### Safety story
Changes mirror existing code used for query screen title. Ran through workflows to ensure translations work as expected.

### Automated test coverage
Not 100% sure if covered or might need additional - formplayer tests failed until modifying code to more closely mirror title label, implying at least some coverage.

### QA Plan
To be included in QA for https://github.com/dimagi/commcare-hq/pull/32338 

### Rollback instructions

### Review

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -56,6 +56,7 @@ public class QueryScreen extends Screen {
     private SessionWrapper sessionWrapper;
     private String[] fields;
     private String mTitle;
+    private String description;
     private String currentMessage;
 
     private String domainedUsername;
@@ -98,6 +99,7 @@ public class QueryScreen extends Screen {
         }
 
         mTitle = getTitleLocaleString();
+        description = getDescriptionLocaleString();
         
     }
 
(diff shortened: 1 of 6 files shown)


shubham1g5 (changes requested):
@nospame The code looks great to me. Reg. test coverage,  I think the simplest way to test this out is to add the xml model for `description` (and `title` since that seems to be missing as well) in the [suite.xml](https://github.com/dimagi/commcare-core/blob/master/src/test/resources/case_claim_example/suite.xml#L57) for `case_claim_example` test app and then modify [this test ](https://github.com/dimagi/commcare-core/blob/formplayer/src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java#L49) to have additional asserts for `title` and `description`. Feel free to ping me if you need any help around it.

nospame:
> @nospame The code looks great to me. Reg. test coverage, I think the simplest way to test this out is to add the xml model for `description` (and `title` since that seems to be missing as well) in the [suite.xml](https://github.com/dimagi/commcare-core/blob/master/src/test/resources/case_claim_example/suite.xml#L57) for `case_claim_example` test app and then modify [this test ](https://github.com/dimagi/commcare-core/blob/formplayer/src/test/java/org/commcare/backend/suite/model/test/CaseClaimModelTests.java#L49) to have additional asserts for `title` and `description`. Feel free to ping me if you need any help around it.

Thanks for your help on this. I added those asserts with `suite.xml` pulling from `app_strings.txt`. I don't know if it's more correct to do it that way, or to put the text directly in the `<text>` blocks in `suite.xml`, but both seem to work the same.

shubham1g5 (approved):
Thanks, this looks great!

PR #1152, opened 2022-09-23
Adds support for auto selecting multiple cases at once 

shubham1g5:
## Technical Summary

https://dimagi-dev.atlassian.net/browse/USH-2435

[Spec](https://docs.google.com/document/d/1C7siuOd6tZbgENuSU3pd2oFQmNEUprnuCAy7EeikeDM/edit)

Utilises the existing `autoselect` datum attribute to select all cases on a multi select entity screen and forwards user to next screen on a valid entities selection.

### Safety story

- The change should only affect Formplayer multi-select workflow where we have good navigation tests to catch issues in this area
- Will go through Formplayer regression QA

**Have not been able to test end to end yet, but would do so before merge.**

### Automated test coverage

Added tests in cross-requested formplayer PR

### QA Plan

https://dimagi-dev.atlassian.net/browse/QA-4555

### Special deploy instructions

- [x] This PR can be deployed after merge with no further considerations.

### Rollback instructions

- [x] This PR can be reverted after deploy with no further considerations.

### Review

- [x] The set of people pinged as reviewers is appropriate for the level of risk of the change.

cross-request: https://github.com/dimagi/formplayer/pull/1226

PR #1167, opened 2022-11-18
Revert "Revert "Adds support for auto selecting multiple cases at once ""

shubham1g5:
Reimplements dimagi/commcare-core#1152

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityScreen.java
@@ -131,7 +131,7 @@ public class EntityScreen extends CompoundScreenHost {
 
         evalContext.setQueryContext(newContext);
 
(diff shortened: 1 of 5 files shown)


src/cli/java/org/commcare/util/screen/EntityScreen.java
```diff
@@ -131,7 +131,7 @@ public void init(SessionWrapper session) throws CommCareSessionException {
 
         evalContext.setQueryContext(newContext);
 
-        if (needsFullInit || references.size() == 1) {
+        if (needsFullInit || references.size() == 1 || shouldAutoSelect()) {
```
snopoke, line 134: Having the extra condition here doesn't makes any difference: `A or B or (C and B) = A or B`
(re-posting from https://github.com/dimagi/commcare-core/pull/1152#discussion_r1024935312)

snopoke (approved):

PR #1169, opened 2022-11-18
Fixed Ethiopian and Nepali date conversion functions

RynhardPietersen:
## Product Description

format-date-for-calendar() function is returning a date that is behind when the selected date format is nepali or ethiopian. This fixes that.

## Technical Summary

[Ticket](https://dimagi-dev.atlassian.net/browse/SAAS-13168) describes format-date-for-calendar function not working. Since this only happens depending on which time zone you are currently in I was able to conclude that the problem is that the date sent to formplayer has a timezone attached which triggers formplayer to do a date conversion by subtracting the time offset of the date sent. The getFields function already does timezoneoffset math to counteract this so I decided to use that instead of a calendar object for the Ethiopian function. However, the Nepali function had significant enough differences that it made sense to create its own set of functions that would handle this problem. I simply altered the fromMillis function to do some timeoffset math to fix the problem.

### Safety story

local testing.
Specifically tested that the following time offsets:
14 nov 2022 06:00 - 14 nov 2022 18:00
14 nov 2022 23:59 - 15 nov 2022 00:59
14 nov 2022 23:01 - 15 nov 2022 00:01
The reason for testing these time offsets was to ensure that the returned date was correct regardless of client side and server side date differentials.

### Automated test coverage

I believe it is super hard to write a test for this specific problem because the function is handling the conversion as expected. The problem is the date being passed into formplayer getting converted to a default timezone which is by design. The test would have to mock this happening and then we test to see if the time is being added to the date. Which I believe is something that we can reliably assume would work as expected. It is because of this that I don't think writing a test like that is valuable.

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/javarosa/core/model/utils/DateUtils.java
+++ b/src/main/java/org/javarosa/core/model/utils/DateUtils.java
@@ -120,7 +120,7 @@ public class DateUtils {
         return tzProvider.getTimezoneOffsetMillis();
     }
 
-    private static TimeZone timezone() {
+    public static TimeZone timezone() {
         return tzProvider.getTimezone();
     }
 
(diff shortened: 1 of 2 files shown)


shubham1g5 (approved):
Looks good to me! Thanks.

PR #1176, opened 2022-11-22
Master to Formplayer Backmerge

shubham1g5:
## Technical Summary

Backmerge of https://github.com/dimagi/commcare-core/pull/1120 for `formplayer`. There are only two additional commits on top of https://github.com/dimagi/commcare-core/pull/1120  which is related to Formplayer being on Java 8 - 

- https://github.com/dimagi/commcare-core/commit/e8a8990b89e19ec6782e1a065d4e94810f010772
- https://github.com/dimagi/commcare-core/commit/86d455e496aac03212462f5025a36b97986a2b29

This results into some serialisation changes for `profile.xml` that will cause App DBs to rebuild when this gets deployed. 

### Safety story

Most of the changes are related to Mobile features and should not impact FP. I am planning to test this on staging for a few days before deploying. 

### Automated test coverage

Relying on existing navigation tests and QA automation scripts

### QA Plan

Not planning as changes looks very limited. 

### Rollback instructions

### Review

The PR as a whole is much easier to review in addition to the manual commits listed above. 

cross-request: https://github.com/dimagi/formplayer/pull/1273


snopoke (approved):

shubham1g5:
> Will this happen automatically? 

Yes

> Is there any risk for deploy?

I don't think so. It's going to result in all users asking for the app ccz after download but in my memory I have not seen it causing any issues before. @ctsims would you agree with this ?

shubham1g5:
Confirmed with Clayton that we have rolled out App DB serialisation changes in past without taking any special considerations and it has not caused issues so far. So we think it's fine to do these kind of deploys occassionally.

PR #1179, opened 2022-11-28
Formplayer dependabot/gradle/junit junit 4.13.2

shubham1g5:
https://github.com/dimagi/commcare-core/pull/1171 for FP


shubham1g5:
@avazirna Just FYI that we will need to duplicate `commcare-core` PRs from `dependabot` like this.

PR #1189, opened 2022-11-30
Trigger dependabot version updates for formplayer

shubham1g5:
[This](https://github.com/dependabot/feedback/issues/889) led me believe we can have dependabot create dependency version updates PR for both `formplayer` and `master`. So want to see if it actually works as expected.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,12 @@ updates:
       interval: daily
     labels:
       - dependencies
+
+  # Copy of above config for formplayer
+  - package-ecosystem: gradle
+    directory: "/"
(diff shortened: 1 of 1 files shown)


snopoke (approved):
Seems to be supported by the docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#target-branch

avazirna (approved):
Looks good to me!

shubham1g5:
So this seems to work but it doesn't guarantee that same dependency upgrade is created for both the branches in one go. So we will still need to open manual PRs to keep the two branches in sync.

PR #1198, opened 2022-12-07
Sets Java source and target to version 17 (Formplayer)

avazirna:
Formplayer PR: [#1292](https://github.com/dimagi/formplayer/pull/1292)

### DIFF
--- a/src/translate/java/org/javarosa/engine/xml/XmlUtil.java
+++ b/src/translate/java/org/javarosa/engine/xml/XmlUtil.java
@@ -41,7 +41,13 @@ public class XmlUtil {
             DOMConfiguration domConfig = lsSerializer.getDomConfig();
             domConfig.setParameter("format-pretty-print", true);
             domConfig.setParameter("cdata-sections", true);
-            lsSerializer.setNewLine("\n");
+
+            // LSSerializer does not have an explicit control for ending the XML Declaration with a
+            // newline, hence the need for the property isStandalone
+            // More in https://bugs.openjdk.org/browse/JDK-8259502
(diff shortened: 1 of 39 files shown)


snopoke (commented):
LGTM but there are some merge conflicts.

src/translate/java/org/javarosa/engine/xml/XmlUtil.java
```diff
@@ -41,7 +41,13 @@ public static String getPrettyXml(byte[] xml) {
             DOMConfiguration domConfig = lsSerializer.getDomConfig();
             domConfig.setParameter("format-pretty-print", true);
             domConfig.setParameter("cdata-sections", true);
-            lsSerializer.setNewLine("\n");
+
+            // LSSerializer does not have an explicit control for ending the XML Declaration with a
+            // newline, hence the need for the property isStandalone
+            // More in https://bugs.openjdk.org/browse/JDK-8259502
+            domConfig.setParameter("http://www.oracle.com/xml/jaxp/properties/isStandalone", true);
```
shubham1g5, line 48: Does this property add or remove new lines ? From the description it feels like it adds a new line, but I see that we have removed new lines from `pretty_printed_xml`.

shubham1g5 (approved):

src/translate/java/org/javarosa/engine/xml/XmlUtil.java
avazirna: It just adds a new line after the xml declaration. The removed lines are from deleting newlines between namespaces, it seems that that format is not applied by the `format-pretty-print` property.

PR #1201, opened 2022-12-07
Auto multi select no results displays empty case list

stephherbers:
## Product Description

Ticket: [USH-2684](https://dimagi-dev.atlassian.net/browse/USH-2684)
formplayer PR:

This change will navigate the user to an empty case list rather that throwing an error on a case list query screen if there are no results to select.

## Technical Summary

Noting this is the only exception to the idea that if auto-select is enabled, a case list screen will never be shown

### Safety story

Tested locally and (shortly) on staging
Plus implements a workflow that already existed outside of auto-select

### QA Plan

I was not planning on QA and rather testing locally and on staging

### Rollback instructions

### Review


shubham1g5:
@stephherbers Would you be able to duplicate this PR for `master` as well.

PR #1202, opened 2022-12-13
Sets JDK source and target to version 17

avazirna:
This PR is to update Java to version 17.

cross-request: https://github.com/dimagi/commcare-android/pull/2600

### DIFF
--- a/src/main/java/org/commcare/xml/MarkupParser.java
+++ b/src/main/java/org/commcare/xml/MarkupParser.java
@@ -29,6 +29,6 @@ public class MarkupParser extends ElementParser<Integer> {
         //exit grid block
         parser.nextTag();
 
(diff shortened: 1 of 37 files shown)


shubham1g5 (approved):

src/main/java/org/commcare/xml/MarkupParser.java
```diff
@@ -29,6 +29,6 @@ public Integer parse() throws InvalidStructureException, IOException, XmlPullPar
         //exit grid block
         parser.nextTag();
 
-        return new Integer(1);
+        return Integer.valueOf(1);
```
shubham1g5, line 32: we should cherry-pick this commit in the FP side PR as well.

PR #1215, opened 2023-02-03
[No Op] Refactoring for instance loading methods

shubham1g5:
## Technical Summary

- [Major change](https://github.com/dimagi/commcare-core/commit/2e645e85123505092a2a7ed248bd6fd701616b8f) in PR is to generalise instance root as `AbstractTreeElement` so that we can support other storage backed roots in instance
- Some function renaming and moving around

## Safety Assurance

Abundant test coverage around changed methods, errors should surface at build time. 

### QA Plan

Not required

### Rollback instructions

### Review

FP: https://github.com/dimagi/formplayer/pull/1324

### DIFF
--- a/src/main/java/org/javarosa/core/model/instance/utils/InstanceUtils.java
+++ b/src/main/java/org/javarosa/core/model/instance/utils/InstanceUtils.java
@@ -13,16 +13,15 @@ import java.io.IOException;
 import java.io.InputStream;
 
 /**
- * Collection of static form loading methods
+ * Collection of static instance loading methods
  *
  * @author Phillip Mates
  */
-public class FormLoadingUtils {
+public class InstanceUtils {
 
     public static FormInstance loadFormInstance(String formFilepath)
             throws InvalidStructureException, IOException {
         TreeElement root = xmlToTreeElement(formFilepath);
-
         return new FormInstance(root, null);
     }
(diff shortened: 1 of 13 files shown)


src/main/java/org/javarosa/core/model/instance/utils/InstanceUtils.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /**
+     * Sets instance properties to the given instance root
+     *
+     * @param instanceRoot instance root
+     * @param instanceId   instance id to set
+     * @param instanceBase instance base to set
+     */
+    public static void setUpInstanceRoot(AbstractTreeElement instanceRoot, String instanceId,
```
snopoke, line 33: should this complain if it's not a `TreeElement`?

snopoke (approved):

src/main/java/org/javarosa/core/model/instance/utils/InstanceUtils.java
shubham1g5: https://github.com/dimagi/commcare-core/pull/1215/commits/a95afe09f18477fa1b911dfbf2d9781602681ed0

snopoke (approved):

PR #1219, opened 2023-02-10
Adds method to Join Multiple Case search field values

shubham1g5:
To be used only by Android as it happens on Browser side for Formplayer

### Rollback instructions

### Review

cross-request: https://github.com/dimagi/commcare-android/pull/2637

### DIFF
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -153,7 +153,10 @@ public class RemoteQuerySessionManager {
                 XPathExpression excludeExpr = prompt.getExclude();
                 if (!(params.containsKey(key) && params.get(key).contains(value))) {
                     if (value != null && (excludeExpr == null || !(boolean)excludeExpr.eval(evaluationContext))) {
-                        params.put(key, value);
+                        String[] choices = RemoteQuerySessionManager.extractMultipleChoices(value);
+                        for (String choice : choices) {
+                            params.put(key, choice);
+                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


avazirna (approved):

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -153,7 +153,10 @@ public Multimap<String, String> getRawQueryParams(boolean skipDefaultPromptValue
                 XPathExpression excludeExpr = prompt.getExclude();
                 if (!(params.containsKey(key) && params.get(key).contains(value))) {
                     if (value != null && (excludeExpr == null || !(boolean)excludeExpr.eval(evaluationContext))) {
-                        params.put(key, userAnswers.get(key));
+                        String[] choices = RemoteQuerySessionManager.extractMultipleChoices(value);
+                        for (String choice : choices) {
+                            params.put(key, choice);
+                        }
```
snopoke, line 159: This seems like a functional change but I don't see any test changes. How is it that it doesn't impact existing functionality?

shubham1g5: this was earlier happening in `getRequestData` method in `QueryScreen` which was used in FP to get the request params. We are only adding support for multi-valued params on mobile now which prompted the change to move this code to the `getRawQueryParams` method and remove `getRequestData` completely. So FP will [now directly utilise](https://github.com/dimagi/formplayer/pull/1340/files) `getRawQueryParams` as well which is a no-op from FP perspective. For mobile, this doesn't result into a change as `extractMultipleChoices` just return the `value` itself for single-valued parameters. For multi-valued parameteres it results into a change of splitting the multiple values into `choices` and adding them to the `params` which is the change we want to make. There is a [test change ](https://github.com/dimagi/commcare-android/pull/2637/commits/3ab113b8bf27696ac0908329be1157bc69c3650f)on mobile PR to test the same.

snopoke (approved):

PR #1254, opened 2023-03-28
Script to automate duplication of PRs for Formplayer and Mobile

Jtang-1:
## Product Description

Script for internal use.

## Technical Summary

https://dimagi-dev.atlassian.net/browse/USH-2975
The arguments will be pulled from github actions - feature to be added as part of follow up task https://dimagi-dev.atlassian.net/browse/USH-2669

### Safety story

Local testing, has no impact on data

### QA Plan

No QA

### Rollback instructions

### Review

### DIFF
--- a/scripts/duplicate_pr.py
+++ b/scripts/duplicate_pr.py
@@ -1,0 +1,19 @@
+from enum import Enum
+
+import argparse
+
+class BranchName(Enum):
+    MASTER = "master"
+    FORMPLAYER = "formplayer"
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Duplicate and push PR between formplayer/master')
+    parser.add_argument('orig_pr_id', type=str, help="ID of PR to be cloned")
+    parser.add_argument('orig_source_branch', type=str, help="Branch name of PR to be duplicated")
+    parser.add_argument('orig_target_branch', type=str, help="Name of branch the original PR merged into",
+                            choices = [key.value for key in BranchName])
+    args = parser.parse_args()
+
+if __name__ == "__main__":
+    main()
--- a/scripts/duplicate_pr.py
+++ b/scripts/duplicate_pr.py
@@ -64,6 +64,11 @@ def cherry_pick_new_commits(commits:list[str], branch:str, git=None):
         git("cherry-pick", commits)
 
 
+def git_push_pr(branch:str, git=None):
+    git = git or get_git()
+    git.push("origin", branch)
+
+
 def merge_base_commit(branch1: str, branch2:str, git=None):
     git = git or get_git()
     return str(git("merge-base", branch1, branch2).replace("\n", ""))
@@ -110,6 +115,9 @@ def main():
 
     print("Cherry-picking commits from {} to {}".format(args.orig_source_branch, new_source_branch))
     cherry_pick_new_commits (new_commits, new_source_branch)
+
+    print("Pushing {}".format(new_source_branch))
+    git_push_pr(new_source_branch)
 
 
 if __name__ == "__main__":
--- a/scripts/duplicate_pr.py
+++ b/scripts/duplicate_pr.py
@@ -47,6 +47,26 @@ def git_pull_pr(pr_id:str, new_branch_name:str, git=None):
         exit(1)
 
 
+def get_new_commits(base_branch: str, curr_branch:str, git=None):
+    git = git or get_git()
+    base_commit = merge_base_commit(base_branch, curr_branch)
+    recent_commit = latest_commit(curr_branch)
+
+    commits_range = "{}..{}".format(base_commit, recent_commit)
+    interested_commits = git("rev-list", "--no-merges", commits_range).split()
+    return interested_commits
+
+
+def merge_base_commit(branch1: str, branch2:str, git=None):
+    git = git or get_git()
+    return str(git("merge-base", branch1, branch2).replace("\n", ""))
+
+
+def latest_commit(branch:str, git=None):
+    git = git or get_git()
+    return str(git("rev-parse", branch).replace("\n", ""))
+
+
 def _wrap_with(code):
 
     def inner(text, bold=False):
@@ -77,6 +97,9 @@ def main():
 
     print("Pulling {}".format(args.orig_source_branch))
     git_pull_pr(args.orig_pr_id, args.orig_source_branch)
+
+    print("Getting new commits from {}".format(args.orig_source_branch))
+    new_commits = get_new_commits(args.orig_target_branch, args.orig_source_branch)
 
 
 if __name__ == "__main__":
--- a/scripts/duplicate_pr.py
+++ b/scripts/duplicate_pr.py
@@ -23,6 +23,20 @@ def get_target_branch(orig_target_branch:str):
         exit(1)
 
 
+def git_create_branch(orig_branch_name:str, new_branch_name: str, git=None):
+    git = git or get_git()
+    try:
+        git.checkout(orig_branch_name)
+    except sh.ErrorReturnCode_1 as e:
+        print(e.stderr.decode())
+        exit(1)
+    try:
+        git.checkout('-b', new_branch_name)
+    except sh.ErrorReturnCode_128 as e:
+        print(e.stderr.decode())
+        exit(1)
+
+
 def main():
     parser = argparse.ArgumentParser(description='Duplicate and push PR between formplayer/master')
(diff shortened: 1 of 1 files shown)


scripts/duplicate_pr.py
```diff
@@ -0,0 +1,124 @@
...
+    args = parser.parse_args()
+
+    new_source_branch = "copy_of_" + args.orig_source_branch
+    new_target_branch = get_target_branch(args.orig_target_branch)
+
+    print("Creating branch {} from {}".format(new_source_branch, new_target_branch))
+    git_create_branch(orig_branch_name=new_target_branch, new_branch_name=new_source_branch)
+
+    print("Pulling {}".format(args.orig_source_branch))
+    git_pull_pr(args.orig_pr_id, args.orig_source_branch)
```
snopoke, line 111: what is this step doing?

Jtang-1: It pulls the PR branch in locally and names it the same branch name. This is branch is used to compare with the master or formplayer branch to get the new commits.

scripts/duplicate_pr.py
```diff
@@ -0,0 +1,124 @@
...
+        return "\033[%sm%s\033[0m" % (c, text)
+    return inner
+
+
+red = _wrap_with('31')
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Duplicate and push PR between formplayer/master')
+    parser.add_argument('orig_pr_id', type=str, help="ID of PR to be cloned")
```
shubham1g5, line 98: do we need to have PR ID if we have original branch name ? Can't we just pull the commits using branch name ?

Jtang-1: I'm using the PR ID to pull the latest PR branch from remote in cases where that branch doesn't exist locally. Are you saying there's a way to do this with just the branch name?

shubham1g5: yeah, we can pull the branch directly from remote - 

````
git fetch origin/branch_name
git checkout branch_name
````

Comment (Jtang-1) [reply scripts/duplicate_pr.py]: Thanks for the tip - [5e34e42](https://github.com/dimagi/commcare-core/pull/1254/commits/5e34e42c75e16722f05e19cc8b544e490e37a748)

And I think it's `git fetch origin branch_name`

File: scripts/duplicate_pr.py [hunk 3]
```diff
@@ -0,0 +1,124 @@
...
+        git.checkout('-b', new_branch_name)
+    except sh.ErrorReturnCode_128 as e:
+        print(red(e.stderr.decode()))
+        exit(1)
+
+
+def git_fetch_pr_branch(branch_name:str, git=None):
+    git = git or get_git()
+    # fetch remote branch without switching branches
+    input = "{0}:{0}".format(branch_name)
```
shubham1g5, line 43: would this make a string like `branch_name:branch_name` ? If so, what's the purpose of doing so ?

scripts/duplicate_pr.py
```diff
@@ -0,0 +1,124 @@
...
+    git_fetch_pr_branch(args.orig_source_branch)
+
+    print("Getting new commits from {}".format(args.orig_source_branch))
+    new_commits = get_new_commits(args.orig_target_branch, args.orig_source_branch)
+
+    print("Cherry-picking commits from {} to {}".format(args.orig_source_branch, new_source_branch))
+    cherry_pick_new_commits (new_commits, new_source_branch)
+
+    print("Pushing {}".format(new_source_branch))
+    git_push_pr(new_source_branch)
```
shubham1g5, line 120: Is it also possible to create a new PR after this from the new source branch ? Maybe not but wanted to check.

scripts/duplicate_pr.py
Jtang-1: `git fetch origin branch_name:branch_name` has the same result as 
`git fetch origin branch_name` + `git checkout branch_name` but without having to switch to the new branch. I think switching to the new branch would be a side affect of this function, so I wanted to avoid that.

scripts/duplicate_pr.py
Jtang-1: I don't think it's possible with just git - there's an extension that can do it but don't think we should do that.
https://stackoverflow.com/questions/4037928/can-you-issue-pull-requests-from-the-command-line-on-github

I was thinking to create the pull request using github action

scripts/duplicate_pr.py
shubham1g5: ahh interesting. We probably don't need to do `git checkout branch_name` here though and just fetching should be enough ?

scripts/duplicate_pr.py
shubham1g5: makes sense, thanks!

scripts/duplicate_pr.py
Jtang-1: Just fetching doesn't create the branch locally if the branch doesn't already exist. So `get_new_commits` step would throw something like "object does not exist"

shubham1g5: sounds right, thanks!

scripts/duplicate_pr.py
```diff
@@ -0,0 +1,124 @@
...
+        return BranchName.FORMPLAYER.value
+    else:
+        print("Only following branches are allowed: {}".format(", ".join([key.value for key in BranchName])))
+        exit(1)
+
+
+def git_create_branch(orig_branch_name:str, new_branch_name: str, git=None):
+    git = git or get_git()
+    try:
+        git.checkout(orig_branch_name)
```
snopoke, line 29: Won't this error if the branch doesn't exist locally? I think you need to call `git_fetch_pr_branch` before this is called

scripts/duplicate_pr.py
```diff
@@ -0,0 +1,124 @@
...
+    git_create_branch(orig_branch_name=new_target_branch, new_branch_name=new_source_branch)
+
+    print("Fetching {}".format(args.orig_source_branch))
+    git_fetch_pr_branch(args.orig_source_branch)
+
+    print("Getting new commits from {}".format(args.orig_source_branch))
+    new_commits = get_new_commits(args.orig_target_branch, args.orig_source_branch)
+
+    print("Cherry-picking commits from {} to {}".format(args.orig_source_branch, new_source_branch))
+    cherry_pick_new_commits (new_commits, new_source_branch)
```
snopoke, line 117: nit:
```suggestion
    cherry_pick_new_commits(new_commits, new_source_branch)
```

scripts/duplicate_pr.py
```diff
@@ -0,0 +1,124 @@
...
+def cherry_pick_new_commits(commits:list[str], branch:str, git=None):
+    git = git or get_git()
+    git.checkout(branch)
+    for commits in reversed(commits):
+        git("cherry-pick", commits)
+
+
+def git_push_pr(branch:str, git=None):
+    git = git or get_git()
+    git.push("origin", branch)
```
snopoke, line 70: it would be nice to show the output of this command since it will contain the link to create the PR on Github

scripts/duplicate_pr.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def merge_base_commit(branch1: str, branch2:str, git=None):
+    git = git or get_git()
+    return str(git("merge-base", branch1, branch2).replace("\n", ""))
+
+
+def latest_commit(branch:str, git=None):
```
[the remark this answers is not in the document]
Jtang-1: [44a2ff1](https://github.com/dimagi/commcare-core/pull/1254/commits/44a2ff12d803664213f5656417f4d6422f86d789)

scripts/duplicate_pr.py
Jtang-1: [44a2ff1](https://github.com/dimagi/commcare-core/pull/1254/commits/44a2ff12d803664213f5656417f4d6422f86d789)

scripts/duplicate_pr.py
Jtang-1: yes fair point, I was assuming part of the user set up would be making sure latest master/formplayer is pulled but this change improves that.
[0e71243](https://github.com/dimagi/commcare-core/pull/1254/commits/0e71243aa392edb24de1c1960b4f8c5057bb390f)

scripts/duplicate_pr.py
Jtang-1: [00dc3c6](https://github.com/dimagi/commcare-core/pull/1254/commits/00dc3c68e10c144bdab0b480989001243565c440)

PR #1266, opened 2023-04-19
Rc/custom assertions core

Robert-Costello:
## Product Description

This handles custom assertions in menus. Menus will be displayed or not depending on the evaluation of the assertions.

## Technical Summary

https://docs.google.com/document/d/112St-i7h7jZHpqxcAU-d6VASftXKLfiFnqkS3ovsEVU/edit#heading=h.37a7vgf78nog
Jira ticket:  https://dimagi-dev.atlassian.net/browse/USH-2983
Jira epic with this and HQ tickets: https://dimagi-dev.atlassian.net/browse/USH-2905

### Safety story

Feature Flagged (CUSTOM_ASSERTIONS: Inject custom assertions into the suite), will go through QA

### Automated test coverage

tests add here

### QA Plan

will be tested in QA along with required HQ changes

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/suite/model/MenuLoader.java
+++ b/src/main/java/org/commcare/suite/model/MenuLoader.java
@@ -73,7 +73,7 @@ public class MenuLoader {
             for (Menu m : s.getMenus()) {
                 try {
                     if (m.getId().equals(menuID)) {
-                        if (menuIsRelevant(sessionWrapper, m)) {
+                        if (menuIsRelevant(sessionWrapper, m) && menuAssertionsPass(sessionWrapper, m)) {
                             addRelevantCommandEntries(sessionWrapper, m, items, badges, map, includeBadges);
                         }
                     } else {
@@ -143,6 +143,11 @@ public class MenuLoader {
             return result;
         }
         return true;
+    }
+
+    public boolean menuAssertionsPass(SessionWrapperInterface sessionWrapper, Menu m) {
+        Text text =  m.getAssertions().getAssertionFailure(sessionWrapper.getEvaluationContext());
+        return text == null;
     }
 
     private void addRelevantCommandEntries(SessionWrapperInterface sessionWrapper,
--- a/src/main/java/org/commcare/suite/model/MenuLoader.java
+++ b/src/main/java/org/commcare/suite/model/MenuLoader.java
@@ -15,7 +15,9 @@ import org.javarosa.xpath.expr.FunctionUtils;
 import org.javarosa.xpath.expr.XPathExpression;
 import org.javarosa.xpath.parser.XPathSyntaxException;
 
+import java.util.HashMap;
 import java.util.Hashtable;
+import java.util.Map;
 import java.util.Set;
 import java.util.Vector;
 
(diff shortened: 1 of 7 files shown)


shubham1g5 (approved):
Looks good to me, although we should also have a corresponding FP PR that tests for error behaviour with failing assertions, for eg - if we get the right error in response from FP when an assertion fail for a `navigate_menu` request to FP. 

It would also be nice to add more real world tests for assertions either in FP PR or here.

Robert-Costello:
> if we get the right error in response from FP when an assertion fail for a `navigate_menu` request to FP.
> 

working on this now

src/main/java/org/commcare/suite/model/MenuLoader.java
```diff
@@ -73,7 +73,7 @@ private void getMenuDisplayables(CommCarePlatform platform,
             for (Menu m : s.getMenus()) {
                 try {
                     if (m.getId().equals(menuID)) {
-                        if (menuIsRelevant(sessionWrapper, m)) {
+                        if (menuIsRelevant(sessionWrapper, m) && menuAssertionsPass(sessionWrapper, m)) {
```
snopoke, line 76: Having this here seems like it will behave exactly like menu relevancy conditions. I thought this would only trigger when you select a menu and that the assertion text would get shown to the user.

esoergel:
FYI for reviewers, this is all on staging now, and I set up a test app if you want to try it out.  Instructions [here](https://dimagi-dev.atlassian.net/browse/USH-2983?focusedCommentId=261665)

src/main/java/org/commcare/suite/model/MenuLoader.java
Robert-Costello: after adding in the error handling here 7797be3e93df0517bdbb0e5a8936c4128ade02ce this seems to work as expected. I tested locally with @esoergel's with custom assertions set up.

src/main/java/org/commcare/suite/model/MenuLoader.java
```diff
@@ -145,6 +145,14 @@ private boolean menuIsRelevant(SessionWrapperInterface sessionWrapper, Menu m) t
         return true;
     }
 
+    public boolean menuAssertionsPass(SessionWrapperInterface sessionWrapper, Menu m) {
+        Text text =  m.getAssertions().getAssertionFailure(sessionWrapper.getEvaluationContext());
+        if (text != null) {
+            loadException = new Exception(text.evaluate());
+        }
+        return true;
```
snopoke, line 153: shouldn't this return false if the assertion failed?

Robert-Costello: fixed here https://github.com/dimagi/commcare-core/pull/1266/commits/61fb618032bba12b414049e2b0bbefdf7ba44296

src/main/java/org/commcare/suite/model/MenuLoader.java
```diff
@@ -145,11 +146,28 @@ private boolean menuIsRelevant(SessionWrapperInterface sessionWrapper, Menu m) t
...
+            EvaluationContext ec = sessionWrapper.getRestrictedEvaluationContext(m.getId(),
+                    instancesNeededByAssertionCondition);
+
+            EvaluationContext traceableContext = new EvaluationContext(ec, ec.getOriginalContext());
+            if (traceReporter != null) {
+                traceableContext.setDebugModeOn(traceReporter);
+            }
+
+            InstrumentationUtils.printAndClearTraces(traceReporter, "menu load expand");
+            Text text =  m.getAssertions().getAssertionFailure(ec);
```
shubham1g5, line 165: is it possible to not repeat this code for assertions and relevancy xpaths by moving it to different method.

src/main/java/org/commcare/suite/model/MenuLoader.java
shubham1g5: > Having this here seems like it will behave exactly like menu relevancy conditions. I thought this would only trigger when you select a menu and that the assertion text would get shown to the user.

@snopoke This method only gets invoked on clicking a menu. The `menuIsRelevant(sessionWrapper, m)` is more of a safety check here and in reality user would not be able to see this menu in the first place if it's not relevant.

src/main/java/org/commcare/suite/model/MenuLoader.java
Robert-Costello: I was working on breaking out the debug code into a separate function, but then I realized you might be referring to that and the instance accumulation code as well. Do you mean both?

src/main/java/org/commcare/suite/model/MenuLoader.java
```diff
@@ -129,39 +131,22 @@ private boolean menuIsRelevant(SessionWrapperInterface sessionWrapper, Menu m) t
...
             return result;
         }
         return true;
     }
 
-    public boolean menuAssertionsPass(SessionWrapperInterface sessionWrapper, Menu m) throws XPathSyntaxException{
-        Vector<String> assertionXpathStrings = m.getAssertions().getAssertionsXPaths();
-        if (!assertionXpathStrings.isEmpty()) {
-            XPathExpression assertionXpath = XPathParseTool.parseXPath(assertionXpathStrings.get(0));
+    public boolean menuAssertionsPass(SessionWrapperInterface sessionWrapper, Menu m) throws XPathSyntaxException {
```
shubham1g5, line 143: The code organisation in this method doesn't make much sense as of now. We are looping over individual Xpaths and then calling `m.getAssertions().getAssertionFailure(ec);` inside the loop which goes over all assertions again. Also the `ec` is specific to the individual assertion Xpath that you are iterating against and can't be used to evaluate all assertions. 

I think we need to eval individual assertions instead here, so instead of doing `m.getAssertions().getAssertionFailure(ec);` , we wanna do something like - 
````
Text text  = m.getAssertions().evalAssertionAtIndex(i, ec)
 ````

shubham1g5 (changes requested):

shubham1g5:
@Robert-Costello Not sure how this branch was setup but it seems to not have the latest code from `master` or `formplayer` branches which is causing staging to break. I did a formplayer merge on this branch but it didn't resolve it so it seems like it's in some kind of detached state. I am removing the branch from staging again for now and we will need to resolve this issue before we put it back on staging.

You can compare the [InstanceUtils](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/instance/utils/InstanceUtils.java#L41) between this branch and `formplayer` or `master` to see this issue.

Robert-Costello:
> we will need to resolve this issue before we put it back on staging.

staging issue is fixed here https://github.com/dimagi/commcare-core/pull/1267/commits/0ad33c5383ebf6783a2d245b5467704b89f11f58

> also noting that it's been quite confusing to follow the review with 2 parallel branches, I think it's best to duplicate the branch once the changes have been finalised just before the merge and use the duplicate PR script for it)

agreed, it's gotten a bit messy. I'll run the duplicate PR script after QA is complete and we're ready to merge.

shubham1g5:
Closing out in favour of https://github.com/dimagi/commcare-core/pull/1304

PR #1283, opened 2023-05-17
[USH-3239] reverse index query optimization

snopoke:
Jira: https://dimagi-dev.atlassian.net/browse/USH-3239
## Technical Summary
This PR contains the platform changes necessary to optimize the reverse index predicate: `[index/parent = current()/@case_id]`.

This predicate is already partially optimized to do a bulk lookup for each value of `current()/@case_id`. What this PR does is fully optimize the lookups so that a single query is performed to find all the index cases for all the cases in the current scope.

This data is then cached and used for later evaluation.

### Safety story
The safety of this should be validated by unit tests.

### Automated test coverage
This XPath evaluation code is well covered by tests and I have added tests explicitly for this optimization in the Formplayer PR. 

### QA Plan
I will put this on staging for final testing just to be safe but do not intend to open a QA ticket for it.

### Rollback instructions

### Review

### DIFF
--- a/src/main/java/org/commcare/modern/engine/cases/ReverseCaseIndexQuerySetTransform.java
+++ b/src/main/java/org/commcare/modern/engine/cases/ReverseCaseIndexQuerySetTransform.java
@@ -1,0 +1,82 @@
+package org.commcare.modern.engine.cases;
+
+import org.commcare.cases.instance.CaseInstanceTreeElement;
+import org.commcare.cases.query.QueryContext;
+import org.commcare.cases.query.queryset.*;
+import org.commcare.modern.util.PerformanceTuningUtil;
+import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.core.model.trace.EvaluationTrace;
+
+import java.util.Set;
+
+/**
+ * Created by skelly on 2023/05/17.
+ */
+public class ReverseCaseIndexQuerySetTransform implements QuerySetTransform {
+
+    private final CaseIndexTable table;
+
+    public ReverseCaseIndexQuerySetTransform(CaseIndexTable table) {
+        this.table = table;
+    }
+
+    @Override
+    public QuerySetLookup getTransformedLookup(QuerySetLookup incoming, TreeReference relativeLookup) {
+        if(incoming.getQueryModelId().equals(CaseQuerySetLookup.CASE_MODEL_ID)) {
+            if(relativeLookup.size() == 2 && "index".equals(relativeLookup.getName(0))) {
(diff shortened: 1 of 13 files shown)


shubham1g5 (approved):
looks great!

ctsims:
@snopoke Awesome change, at this is at just the right place in the engine I think.

I poked around in the engine a bit to refresh my memory, and there are a few edge cases I think would also be helpful to add to the test set.

The first would be a scenario where the basis for the `index/parent` transform **isn't** the the same as the `current()` base that's populated from the QueryContext. That's the most likely place for the engine to get confused. It would be a pattern like:

`instance('casedb')/casedb/case[other_optimizable_filter][index/parent = current()/...]`

the other scenario is making sure that the structure chains, ie: Where `[index/parent = current()/@case_id]` is nested in another predicate. That's the main goal of Model Set transforms, but it's also just a place where things are the most likely to just crash outright if anything's off.

src/main/java/org/commcare/modern/engine/cases/ReverseCaseIndexQuerySetTransform.java
```diff
@@ -0,0 +1,91 @@
...
+import org.commcare.cases.query.queryset.QuerySetLookup;
+import org.commcare.cases.query.queryset.QuerySetTransform;
+import org.commcare.modern.util.PerformanceTuningUtil;
+import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.core.model.trace.EvaluationTrace;
+
+import java.util.Set;
+
+/**
+ * Created by skelly on 2023/05/17.
```
ctsims, line 18: Javadoc outline for this would be really helpful.

ctsims (approved):
This looks great, thanks for taking the time to find the moving pieces to wire it in.

It would be good to update the docs at 
https://github.com/dimagi/commcare-core/wiki/Performance:-Query-Planner#structure

to reflect the new availability of this model set pattern.

ctsims:
Another quick note that I checked the trace instrumentation on [nested contexts](https://github.com/dimagi/commcare-android/pull/2674/files#diff-f326941b39367eb33a0124fd7fd9cb4778454c345d0cf77035d6b393ca115f3aR136) and it looks like it's working correctly as long as the queryset lookup volumes pass the bulk threshold, which is awesome to see. 

Also, when reviewing that test case it looks like even though the Single Loads might have a performance regression, bulk loads don't appear to be affected, the original query plan has the ~same bulk load time (10ms longer in the pre-PR query path).

PR #1289, opened 2023-05-23
Eval entity list in Entity Screens

shubham1g5:
## Technical Summary

Couple Major Changes - 
1. `EntityListSubScreen` now inits, filters and sorts the entity list 
2. Adds a class `EntityListContext` to hold the necessary context for entity screens

FP - https://github.com/dimagi/formplayer/pull/1384

[Jira](https://dimagi-dev.atlassian.net/browse/USH-3165)

### Safety story

Changes are limited to CLI/FP with good test coverage for entity list screens.

### QA Plan

[Detailed here](https://dimagi-dev.atlassian.net/browse/QA-5130?focusedCommentId=266271)

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
@@ -33,7 +33,7 @@ public class EntityListSubscreen extends Subscreen<EntityScreen> {
     private static final int SCREEN_WIDTH = 100;
 
     private final TreeReference[] entitiesRefs;
-    private final String[] rows;
+    private String[] rows;
     private final String mHeader;
 
     private final Vector<Action> actions;
@@ -54,7 +54,6 @@ public class EntityListSubscreen extends Subscreen<EntityScreen> {
         references.copyInto(entitiesRefs);
         actions = shortDetail.getCustomActions(context);
         initEntities(context, shortDetail);
-        rows = getRows(shortDetail);
     }
 
     private void initEntities(EvaluationContext context, Detail shortDetail) {
@@ -172,6 +171,9 @@ public class EntityListSubscreen extends Subscreen<EntityScreen> {
 
     @Override
     public String[] getOptions() {
+        if (rows == null) {
+            rows = getRows(shortDetail);
+        }
         return rows;
     }
 
--- a/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
@@ -2,6 +2,8 @@ package org.commcare.util.screen;
 
 import static org.commcare.util.screen.MultiSelectEntityScreen.USE_SELECTED_VALUES;
 
+import org.commcare.cases.entity.Entity;
+import org.commcare.cases.entity.NodeEntityFactory;
 import org.commcare.modern.util.Pair;
 import org.commcare.suite.model.Action;
 import org.commcare.suite.model.Detail;
@@ -16,6 +18,8 @@ import org.javarosa.core.util.DataUtil;
 import org.javarosa.xpath.XPathException;
 
 import java.io.PrintStream;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Vector;
 
 /**
@@ -38,6 +42,7 @@ public class EntityListSubscreen extends Subscreen<EntityScreen> {
     private final EvaluationContext rootContext;
 
     private boolean handleCaseIndex;
+    private List<Entity<TreeReference>> entities;
 
     public EntityListSubscreen(Detail shortDetail, Vector<TreeReference> references, EvaluationContext context,
             boolean handleCaseIndex) throws CommCareSessionException {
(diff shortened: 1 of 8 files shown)


src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
```diff
@@ -48,71 +53,41 @@ public EntityListSubscreen(Detail shortDetail, Vector<TreeReference> references,
         this.handleCaseIndex = handleCaseIndex;
         references.copyInto(entitiesRefs);
         actions = shortDetail.getCustomActions(context);
-        rows = getRows(entitiesRefs, context, shortDetail);
+        initEntities(context, shortDetail);
```
snopoke: nit: It would make this easier to read if it didn't modify the class but instead returned the list:
```java
entities = initEntities(context, shortDetail)
```

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
```diff
@@ -48,71 +53,41 @@ public EntityListSubscreen(Detail shortDetail, Vector<TreeReference> references,
...
-    private static String[] getRows(TreeReference[] references,
-            EvaluationContext evaluationContext,
-            Detail detail) {
-        String[] rows = new String[references.length];
-        int i = 0;
-        for (TreeReference entity : references) {
-            rows[i] = createRow(entity, evaluationContext, detail);
-            ++i;
+    private void initEntities(EvaluationContext context, Detail shortDetail) {
+        NodeEntityFactory nodeEntityFactory = new NodeEntityFactory(shortDetail, context);
```
snopoke: 👍🏻

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
```diff
@@ -54,7 +54,6 @@ public EntityListSubscreen(Detail shortDetail, Vector<TreeReference> references,
         references.copyInto(entitiesRefs);
         actions = shortDetail.getCustomActions(context);
         initEntities(context, shortDetail);
-        rows = getRows(shortDetail);
```
snopoke: isn't most of the work done in `initEntities`?

src/main/java/org/commcare/cases/entity/EntityStringFilterer.java
```diff
@@ -0,0 +1,57 @@
...
+        if (searchTerms == null || searchTerms.length == 0) {
+            matchList.addAll(fullEntityList);
+        }
+    }
+
+    public List<Entity<TreeReference>> buildMatchList() {
+
+        while (!nodeFactory.isEntitySetReady()) {
+            try {
+                Thread.sleep(100);
```
snopoke, line 41: I know you just copied this but I checked the logic here and there's no multithreading so I don't see why this is necessary.

I do see that mobile implements an async version of the factory. Perhaps a task for future refactor to clean this up.

src/cli/java/org/commcare/util/screen/EntityScreenContext.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Holds essential meta data associated with the entity screen
+ */
+public class EntityScreenContext {
+    private final int mOffSet;
+    private final String mSearchText;
+    private final int mSortIndex;
+    private final int mCasesPerPage;
+    private final String[] mSelectedValues;
```
snopoke, line 11: nit: maybe personal preference but I don't see the point of the `m` prefix, espeically since we're very inconsistent with it (even here `fuzzySearch` doesn't use it).

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
```diff
@@ -45,24 +49,64 @@ public class EntityListSubscreen extends Subscreen<EntityScreen> {
...
+        sortEntities(entityScreenContext);
+    }
+
+    private void filterEntities(EntityScreenContext entityScreenContext) {
+        String searchText = entityScreenContext.getSearchText();
+        boolean isFuzzySearchEnabled = entityScreenContext.isFuzzySearch();
+        if (searchText != null && !"".equals(searchText)) {
+            NodeEntityFactory nodeEntityFactory = new NodeEntityFactory(shortDetail, rootContext);
+            EntityStringFilterer filterer = new EntityStringFilterer(searchText.split(" "),
+                    nodeEntityFactory, entities, isFuzzySearchEnabled);
```
snopoke: this class doesn't appear to actually use the `nodeEntityFactory` for anything except checking `nodeFactory.isEntitySetReady()` which could be moved out. Also, since this is a new instance of `NodeEntityFactory` I think this will raise a runtime exception because `prepareEntities` hasn't been called (but that's moot if we remove the need for it in the first place).

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
```diff
@@ -45,24 +49,64 @@ public class EntityListSubscreen extends Subscreen<EntityScreen> {
...
+            int sortFieldIndex = sortIndex - 1;
+            order = new int[]{sortFieldIndex};
+        } else {
+            order = shortDetail.getOrderedFieldIndicesForSorting();
+            for (int i = 0; i < shortDetail.getFields().length; ++i) {
+                String header = shortDetail.getFields()[i].getHeader().evaluate();
+                if (order.length == 0 && !"".equals(header)) {
+                    order = new int[]{i};
+                }
+            }
```
snopoke: nit: I think this could be improved by checking order length before the loop and exiting early once a non-blank field is found.

(could leave it for now)

snopoke (approved):
minor comments

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
snopoke: resolved in later commit

src/main/java/org/commcare/cases/entity/EntityStringFilterer.java
shubham1g5: agree this is not necessary in FP context and  is there to only ensure the mobile compatibility with this class although mobile implements it's own filter class today. I will be happy to remove this code from here but think we will need to explicitly mention that this class should only be used for CLI/FP. Maybe by moving it to `src/cli` ?

src/cli/java/org/commcare/util/screen/EntityScreenContext.java
shubham1g5: Yeah I think it's a style preference. I just don't like having to use `this.var` in the constructor as it's unnecessary verbose.

src/cli/java/org/commcare/util/screen/EntityListSubscreen.java
shubham1g5: Good catch, https://github.com/dimagi/commcare-core/pull/1289/commits/df48f8c50ad6ec67936151745919d1953163e329

PR #1291, opened 2023-05-24
Retrieve HTTP error message from server

OrangeAndGreen:
## Technical Summary
Added some code that attempts to retrieve an error message from an HTTP server response for debugging purposes. The code is wrapped in a try-catch where the catch is ignored, ensuring that nothing else will break if the new code is unable to retrieve an error message.

### Safety story
Wrapped the new code in a try-catch with an empty catch body, such that any problem while attempting to retrieve the error message will be ignored.

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/core/network/ModernHttpRequester.java
+++ b/src/main/java/org/commcare/core/network/ModernHttpRequester.java
@@ -87,7 +87,19 @@ public class ModernHttpRequester implements ResponseStreamAccessor {
         }
         try {
             response = makeRequest();
-            processResponse(responseProcessor, response.code(), this);
+
+            String errorMessage = null;
+            try {
+                ResponseBody body = response.errorBody();
+                if(body != null) {
+                    errorMessage = body.string();
+                }
+            }
+            catch(Exception e) {
+                //Ignore exception when we can't retrieve an error message
+            }
+
+            processResponse(responseProcessor, response.code(), errorMessage, this);
         } catch (IOException e) {
             e.printStackTrace();
             responseProcessor.handleIOException(e);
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -143,6 +155,7 @@ private Response executeAndCheckCaptivePortals(Call currentCall) throws IOExcept
 
     public static void processResponse(HttpResponseProcessor responseProcessor,
                                        int responseCode,
+                                       String errorMessage,
```
shubham1g5, line 158: this is unused at this point of time, would be nice to see how you are actually using it. Though extracting it and sending it to the response processor in case of errors sounds right to me as a change.

OrangeAndGreen: Yeah, I wasn't sure how far to pass the message up the chain so stopped here. In practice I'm only using it during debugging, as some errors from the ConnectID server are coming back as strings. So when I need to debug a failed server call I just set a breakpoint in the error case and read the errorMessage that way.

But in the long run I don't think the plan is for the server to return error messages as text. Instead we'll have the server return error codes that can be converted to translatable strings (for localization). So I'm open to suggestions as far as how to proceed with this change. Maybe it's not worth keeping around in the long run? Or we could log the errorMessage somehow when it's not empty as a way to relevantly use the code?

shubham1g5: Yeah I don't think we want to add any code just for debugging, you should be able to navigate in debugger to the caller of this method to look at `response.errorBody()` directly through value inspection in debugger.

OrangeAndGreen: No problem, I'll revert this code.

PR #1305, opened 2023-07-06
Make sure step id, value and type are non-null

shubham1g5:
## Technical Summary

[Jira](https://dimagi-dev.atlassian.net/browse/QA-5355)

With Case List Grouping Feature, we add a new computed datum for a form as - 

`<datum id="case_id_parent_ids" function="join(' ', distinct-values(instance('casedb')/casedb/case[@case_id = instance('commcaresession')/session/data/case_id]/index/parent))"/>`

If a case have no parent, this evaluates to a empty string which on deserialization was getting coverted to Null value and ultimately resulting into [the exception here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/data/UncastData.java#L33)

Since the null values in the session frame already results into an error like above, I feel confident for platform to enforce their non-nullability upfront and more explicitly. Doing this allows us to support for empty string values explicitly by treating them different than null values while serialisation. 

### Safety story

This can result into some faulty apps breakign upfront while menu navigation instead of a form submission but I think that should be ok.

Going to test this on staging with a normal app. 

### Automated test coverage

None
### QA Plan

None

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/suite/model/StackFrameStep.java
+++ b/src/main/java/org/commcare/suite/model/StackFrameStep.java
@@ -1,5 +1,6 @@
 package org.commcare.suite.model;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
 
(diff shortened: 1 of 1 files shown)


shubham1g5:
Ahh all command steps have null values :( Closing it out.

PR #1306, opened 2023-07-06
Skip null values in session instance

shubham1g5:
## Technical Summary

[Jira](https://dimagi-dev.atlassian.net/browse/QA-5355)

With Case List Grouping Feature, we add a new computed datum for a form as -

`<datum id="case_id_parent_ids" function="join(' ', distinct-values(instance('casedb')/casedb/case[@case_id = instance('commcaresession')/session/data/case_id]/index/parent))"/>`

If a case have no parent, this evaluates to a empty string which on deserialization was getting coverted to Null value and ultimately resulting into [the exception here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/data/UncastData.java#L33)

I think it should be safe to skip the null values from the session instance instead of crashing out. This will ultimately result into having no corrsponding node in the session xml tree which sounds like the right thing to do with null values. 

### Safety story

Tested on staging. 
Change itself looks very safe since it just stops a null crash. 

### Automated test coverage

No
### QA Plan

No

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/session/SessionInstanceBuilder.java
+++ b/src/main/java/org/commcare/session/SessionInstanceBuilder.java
@@ -116,8 +116,10 @@ public class SessionInstanceBuilder {
 
     private static void addData(TreeElement root, String name, String data) {
         TreeElement datum = new TreeElement(name, root.getChildMultiplicity(name));
-        datum.setValue(new UncastData(data));
-        root.addChild(datum);
+        if (data != null) {
+            datum.setValue(new UncastData(data));
+            root.addChild(datum);
+        }
     }
 
 }


snopoke:
If an app tries to access the session variable what will happen? I think as long as it doesn't error and evaluates to an empty string I think this is fine.

shubham1g5:
> I think as long as it doesn't error and evaluates to an empty string I think this is fine.

Confirming this is indeed the case.

snopoke (approved):

shubham1g5:
duplicate this PR

PR #1312, opened 2023-07-19
Support Endpoints for Instance Datums

shubham1g5:
See https://github.com/dimagi/formplayer/pull/1431

- Corresponding FP PR will need to be reverted if we revert this. 

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
@@ -9,6 +9,7 @@ import org.commcare.core.interfaces.UserSandbox;
 import org.commcare.core.interfaces.VirtualDataInstanceStorage;
 import org.commcare.data.xml.VirtualInstances;
 import org.commcare.modern.session.SessionWrapper;
+import org.commcare.modern.util.Pair;
 import org.commcare.session.CommCareSession;
 import org.commcare.suite.model.MultiSelectEntityDatum;
 import org.commcare.util.FormDataUtil;
@@ -80,7 +81,7 @@ public class MultiSelectEntityScreen extends EntityScreen {
             for (int i = 0; i < selectionSize; i++) {
                 evaluatedValues[i] = getReturnValueFromSelection(references.elementAt(i));
             }
-            processSelectionIntoInstance(evaluatedValues);
+            processSelectionIntoInstance(evaluatedValues, getNeededDatumId());
             updateSession(session);
             return true;
         }
@@ -156,7 +157,7 @@ public class MultiSelectEntityScreen extends EntityScreen {
             for (int i = 0; i < selectedRefs.length; i++) {
                 evaluatedValues[i] = getReturnValueFromSelection(selectedRefs[i]);
             }
-            processSelectionIntoInstance(evaluatedValues);
+            processSelectionIntoInstance(evaluatedValues, getNeededDatumId());
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/commcare/xml/EndpointParser.java
```diff
@@ -31,16 +35,23 @@ public Endpoint parse() throws InvalidStructureException, IOException, XmlPullPa
...
                 }
-                arguments.add(argumentID);
+                String argInstanceId = parser.getAttributeValue(null, ATTR_ARGUMENT_INSTANCE_ID);
+                String argInstanceSrc = parser.getAttributeValue(null, ATTR_ARGUMENT_INSTANCE_SRC);
+
+                if (argInstanceId != null && argInstanceSrc == null) {
+                    throw new InvalidStructureException(
+                            "argument containing a non-null instance-id must define an instance-src", parser);
+                }
+                arguments.add(new EndpointArgument(argumentID, argInstanceId, argInstanceSrc));
```
Robert-Costello, line 64: Just to make sure I understand correctly, is the idea here that `EndpointArgument` also handles single case selection where `<argument id="case_id"/>`? So if there is no `argInstanceId ` and no `argInstanceSrc ` the functionality from this point is essentially the same as before.

shubham1g5: yes, spot on.

src/main/java/org/commcare/suite/model/Endpoint.java
```diff
@@ -59,40 +60,60 @@ public Vector<StackOperation> getStackOperations() {
...
     // Utility Functions
     public static void populateEndpointArgumentsToEvaluationContext(Endpoint endpoint, ArrayList<String> args, EvaluationContext evaluationContext) {
-        Vector<String> endpointArguments = endpoint.getArguments();
+        Vector<EndpointArgument> endpointArguments = endpoint.getArguments();
 
         if (endpointArguments.size() > args.size()) {
-            Vector<String> missingArguments = new Vector<String>(endpointArguments.subList(args.size(), endpointArguments.size()));
+            Vector<String> missingArguments = new Vector<>();
+            for (int i = args.size(); i < endpointArguments.size(); i++) {
+                missingArguments.add(endpointArguments.get(i).getId());
```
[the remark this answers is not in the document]
shubham1g5: Yeah we expect the arguments to be passed in order when it's passed as a list without using the argument keys (`arg1=val1&arg2=val2` vs `val1,val2`). Web Apps uses the fuirst format with names but Android can use both those formats. The ordering matters only in the later format.

Robert-Costello (approved):

src/cli/java/org/commcare/util/screen/MultiSelectEntityScreen.java
```diff
@@ -134,9 +135,18 @@ private void prcessSelectionAsGuid(String guid) throws CommCareSessionException
                     "Could not make selection with reference id " + guid + " on this screen. " +
                             " If this error persists please report a bug to CommCareHQ.");
         }
+        validateEntitiesInInstance(cachedInstance);
         storageReferenceId = guid;
     }
 
+    private void validateEntitiesInInstance(ExternalDataInstance instance) throws CommCareSessionException {
```
shubham1g5, line 142: There is currently[ an issue](https://dimagi-dev.atlassian.net/browse/QA-5457) with this validation that I am verifying with delivery

shubham1g5:
duplicate this PR

PR #1319, opened 2023-08-15
Rc/validate required case search input

Robert-Costello:
## Product Description

This adds additional validated that covers required case search inputs. Users will now see a 
Before:
Screenshot 2023-08-15 at 10 49 49 AM

After:
Screenshot 2023-08-15 at 3 35 41 PM

## Technical Summary

jira ticket: https://dimagi-dev.atlassian.net/browse/USH-3131

### Automated test coverage

test were updated here to handle required validation errors

### QA Plan

no QA planned

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -287,7 +287,11 @@ public class RemoteQuerySessionManager {
                 errors.put(key, queryPrompt.getValidationMessage(ec));
             }
             if (StringUtils.isEmpty(value) && queryPrompt.isRequired(ec)) {
-                errors.put(key, queryPrompt.DEFAULT_REQUIRED_ERROR);
+                String message = queryPrompt.getRequiredMessage(ec);
+                if (message.replace('\u00A0',' ').isBlank()) {
+                    message = queryPrompt.DEFAULT_REQUIRED_ERROR;
+                }
+                errors.put(key,  message);
             }
         }
     }
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -271,8 +271,8 @@ public class RemoteQuerySessionManager {
             String key = (String)en.nextElement();
             QueryPrompt queryPrompt = userInputDisplays.get(key);
             boolean isRequired = queryPrompt.isRequired(ec);
-            String value = userAnswers.get(key);
             requiredPrompts.put(key, isRequired);
+            String value = userAnswers.get(key);
             TreeReference currentRef = getReferenceToInstanceNode(instanceId, key);
             if (!StringUtils.isEmpty(value) && queryPrompt.isInvalidInput(new EvaluationContext(ec, currentRef))) {
                 errors.put(key, queryPrompt.getValidationMessage(ec));
--- a/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
+++ b/src/main/java/org/commcare/session/RemoteQuerySessionManager.java
@@ -258,22 +258,11 @@ public class RemoteQuerySessionManager {
     public void refreshInputDependentState() {
         refreshItemSetChoices();
         validateUserAnswers();
-        recalculateRequired();
     }
 
-    private void recalculateRequired() {
-        requiredPrompts = new Hashtable<>();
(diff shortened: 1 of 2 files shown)


src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -286,6 +286,13 @@ private void validateUserAnswers() {
             if (!StringUtils.isEmpty(value) && queryPrompt.isInvalidInput(new EvaluationContext(ec, currentRef))) {
                 errors.put(key, queryPrompt.getValidationMessage(ec));
             }
+            if (StringUtils.isEmpty(value) && queryPrompt.isRequired(ec)) {
+                String message = queryPrompt.getRequiredMessage(ec);
+                if (message.replace('\u00A0',' ').isBlank()) {
```
Robert-Costello, line 291: This replaces the [Non-breaking space](https://www.ascii-code.com/CP1252/160) that is returned from `getRequiredMessage` for fields that are required but don't have a `Require Message` set in app manager. I was stuck on this for a while since `" "`(Space) and `" "`(Non-breaking space) appear the same, but calling `isBlank` or `isEmpty` on a Non-breaking Space returns false. I'm curious if there is a reason a Non-breaking space is returned instead of `null` or an empty string?

shubham1g5: Are you sure that the nbsp is not getting set in HQ for your app ?

shubham1g5 (changes requested):

snopoke (commented):
Generally looks OK pending resolution of other comments.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
Robert-Costello: > Are you sure that the nbsp is not getting set in HQ for your app ?

Looking at this again, I do think the nbsp is coming from HQ. I haven't tracked down exactly where, but it seems to be coming from Backbone js.

shubham1g5: Think this needs to be fixed on HQ in that case instead of us trying to do any string manipulation here.

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
```diff
@@ -258,35 +258,26 @@ public void refreshItemSetChoices() {
...
         errors = new Hashtable<>();
         OrderedHashtable<String, QueryPrompt> userInputDisplays = getNeededUserInputDisplays();
         String instanceId = VirtualInstances.makeSearchInputInstanceID(getSearchInstanceReferenceId());
         EvaluationContext ec = getEvaluationContextWithUserInputInstance();
         for (Enumeration en = userInputDisplays.keys(); en.hasMoreElements(); ) {
             String key = (String)en.nextElement();
             QueryPrompt queryPrompt = userInputDisplays.get(key);
+            boolean isRequired = queryPrompt.isRequired(ec);
             String value = userAnswers.get(key);
+            requiredPrompts.put(key, isRequired);
```
shubham1g5, line 275: lint: probably move the line one up to be in continuation with `isRequired` logic

shubham1g5 (changes requested):

src/main/java/org/commcare/session/RemoteQuerySessionManager.java
Robert-Costello: It turns out that nbsp is actually intentional. It's coming from this translation method: [dumps](https://github.com/dimagi/commcare-translations/blob/0de322558e2bd74b70f051c8790257b0f5929cbe/commcare_translations.py#L107) that is part of generating app strings. Does it make sense in that case to keep `replace` call?

shubham1g5: I see, I am not sure why we are doing that on HQ but I would avoid correcting for HQ bheaviour on FP. I think in context of this PR, we can just assign `message` as `queryPrompt.getRequiredMessage(ec)` and remove this handling. If this becomes a problem down the line, we can push for correcting the behaviour on HQ. I think the right behaviour here will be for HQ to assing the default message instead of nbsp to the string.

shubham1g5 (approved):

Robert-Costello:
duplicate this PR

PR #1323, opened 2023-08-21
Mixed repeat group support for Web Apps

shubham1g5:
## Product Description
There are 2 kinds of repeat groups we support - 1. with no count (user clicks add new repeat to add new group) 2. With count pre-defined. Today Web App supports both of these if you only add one of these kind in the form. But if you add both of these kinds of repeat in a single form, the non counted repeat groups don’t show up at all on web apps. This change is to address support for mixed repeat group in a form on Web Apps.

## Technical Summary

The PR makes a simple change to not rely on `repeatStructure` passed to the [FormEntryModel](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/form/api/FormEntryModel.java#L63) to determine whether the code should follow counted or non counted behaviour.  FP always init the formEntryModel `repeatStructure` as `REPEAT_STRUCTURE_NON_LINEAR` today but [changes it to](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/form/api/FormEntryModel.java#L73) `REPEAT_STRUCTURE_LINEAR` when a counted repeat group is present. This in turn causes the behaviour to be different when a counted repeat group is present in the form along with the non counted repeat group. 

Since a form can have both kinds of repeat groups for different questions, initialising the form model with a particular repeat structure and then navigating the question nodes based on that for all kinds of repeat groups seems to be the root of the problem to me. This PR tries to disregard the repeat structure at certain places and instead directly check for `count` reference defined on the repeat group to fix the issue for mixed repeat groups. 

### Safety story

- Will be [QA'ed](https://dimagi-dev.atlassian.net/browse/QA-5542) both on mobile and Web Apps to test for any regressions

### Automated test coverage

TBD

### QA Plan
https://dimagi-dev.atlassian.net/browse/QA-5542

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/javarosa/core/model/GroupDef.java
+++ b/src/main/java/org/javarosa/core/model/GroupDef.java
@@ -116,6 +116,10 @@ public class GroupDef implements IFormElement {
         return isRepeat;
     }
 
+    public boolean isNonCountedRepeat() {
+        return isRepeat && getCountReference() == null;
+    }
+
     public void setIsRepeat(boolean repeat) {
         this.isRepeat = repeat;
     }
--- a/src/main/java/org/javarosa/form/api/FormEntryModel.java
+++ b/src/main/java/org/javarosa/form/api/FormEntryModel.java
@@ -91,6 +91,10 @@ public class FormEntryModel {
         IFormElement element = form.getChild(index);
         if (element instanceof GroupDef) {
             if (((GroupDef)element).isRepeat()) {
+                if (((GroupDef)element).getCountReference() == null) {
+                    // non counted repeat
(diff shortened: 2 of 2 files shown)


shubham1g5:
Results in breaking existing behaviour for linear structure as for counted repeat groups, we are now touching additional code that was not touched at all. Since mobile is supporting linear structure for both counted and non counted repeat groups, it should be possible to make it work without messing with repeat structure code. I am closing this out and going to approach this in a similar manner as mobile, though that will probably means more changes on Web Apps side to add this support.

PR #1329, opened 2023-09-12
Android 13 update

avazirna:
## Product Description
This PR implements changes required to support Android 13. The relevancy here are ICU updates that made handling data symbols more locale-stricter.

### DIFF
--- a/src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
+++ b/src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java
@@ -12,9 +12,11 @@ import org.commcare.util.LogTypes;
 import org.javarosa.core.services.Logger;
 
 import java.io.IOException;
+import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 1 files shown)


avazirna:
@shubham1g5 I forgot to request your review on this PR.

shubham1g5 (approved):

avazirna:
> Is this not a mandatory API change ? curious why is not it causing android build to fail on `master` without this merged ?

It is. It can cause errors depending on the device configuration, for instance I have my device language set to `en_ZA` (English - South Africa) and now the abbreviation of `September` appears as `Sept` and not `Sep` (like in the English language), which causes the [parser here](https://github.com/dimagi/commcare-core/blob/ac64d2be3211b49e34fbf85907d8b1430d9a9bbb/src/main/java/org/commcare/core/network/CommCareNetworkServiceGenerator.java#L77) to fail. But in this case, this is only relevant for date months in the `MMM` form.

shubham1g5:
I understand the motivation, though how is it related to Android 13 ?

avazirna:
> I understand the motivation, though how is it related to Android 13 ?

My take is that this is caused by the [ICU changes](https://developer.android.com/about/versions/13/features#unicode) brought by Android 13.

avazirna:
duplicate this PR https://github.com/dimagi/commcare-core/commit/5db37b9158f5c70a3e0514208a96a0403632d563

shubham1g5:
duplicate this PR 5db37b9158f5c70a3e0514208a96a0403632d563

PR #1346, opened 2023-10-09
Define end hash of interest for duplicating PR

Jtang-1:
## Technical Summary

Resulting from this workflow failure: https://github.com/dimagi/commcare-core/actions/runs/6459548228/job/17535541395

No commits were returning from `git rev-list --no-merges --first-parent 8850102` This is because commit `af06064` has multiple parents with the `8850102` being the first. The commits of interest are not between the first parent (`8850102`) so it returns an empty list. The commits of interest in the PR to be duplicated are contained between the second parent and `af06064`. However, there is no way to know without user input. With this change, the user will also define the last non-merge commit in the PR. 

So in the linked PR, the comment that would lead to correct PR duplication would be `duplicate this PR 628e7a8 2eacb6b`
### Safety story

Locally tested, only affects internal tool
### Automated test coverage

no automated test
### QA Plan

no QA

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -8,7 +8,7 @@ We try to keep both these branches in sync with each other the best we can. To d
 
 1. If you are working on a Formplayer change, you will want to start by checking out `your_feature_branch` from `formplayer` as the base branch. Make changes on `your_feature_branch` and create your original PR against `formplayer` branch.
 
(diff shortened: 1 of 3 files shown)


shubham1g5 (approved):
Thanks a lot!

Jtang-1:
duplicate this PR bc709b4 29ecf66

PR #1381, opened 2023-11-21
Add flag if to show borders on case list fields

MartinRiese:
## Product Description

Add flag to case list fields to show a border

## Technical Summary

https://dimagi-dev.atlassian.net/browse/USH-3823

### Safety story

Only additive

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/suite/model/DetailField.java
+++ b/src/main/java/org/commcare/suite/model/DetailField.java
@@ -49,6 +49,8 @@ public class DetailField implements Externalizable {
     @Nullable
     private EndpointAction endpointAction;
 
+    private boolean showBorder;
+
     /**
      * Optional hint which provides a hint for whether rich media should be
      * displayed based on <text> returning a URI.  May be either 'image' or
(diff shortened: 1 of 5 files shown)


src/main/java/org/commcare/suite/model/DetailField.java
```diff
@@ -49,6 +49,8 @@ public class DetailField implements Externalizable {
     @Nullable
     private EndpointAction endpointAction;
 
+    private boolean showBorder;
```
shubham1g5, line 52: Feels like this should be part of [Style](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/commcare/suite/model/Style.java) instead

shubham1g5 (changes requested):

shubham1g5:
Can we update the [doc](https://github.com/dimagi/commcare-core/wiki/Suite20) with the attribute as well ? Thanks!

MartinRiese:
Updated the doc

MartinRiese:
duplicate this PR

PR #1390, opened 2023-12-08
treat string of just blank spaces as empty string

Jtang-1:
## Technical Summary

As part of [this change](https://github.com/dimagi/commcare-hq/pull/33738), multiple `data` elements will be added to the `post`.

With this example:
```
<post url="http://localhost:8000/a/jonathanlocal/phone/claim-case/" relevant="$case_id!=''">
  <data key="case_id" ref="instance('commcaresession')/session/data/case_id" exclude="count(instance('casedb')/casedb/case[@case_id=instance('commcaresession')/session/data/case_id]) != 0"/>
  <data key="case_id" ref="instance('commcaresession')/session/data/parent_id" exclude="count(instance('casedb')/casedb/case[@case_id=instance('commcaresession')/session/data/parent_id]) != 0"/>
</post>
```
If both data are blank, the string generated [here](https://github.com/dimagi/commcare-core/blob/fe5114b44ebe4128205a903be9d69d3d066baf23/src/main/java/org/commcare/suite/model/PostRequest.java#L75) results in a string containing only whitespace. This cause  `$case_id!=''` to incorrectly evaluate as `True`. This change saves a value that is a string with only whitespace as an empty string instead.

### Safety story

locally tested. Limited to values that are strings of whitespace which should not exist.
### QA Plan

no QA

Must be deployed with or before `https://github.com/dimagi/commcare-hq/pull/33738`

### Rollback instructions

If this PR is reverted, it will break any existing apps that use inline case search with parent select (related to this [HQ change](https://github.com/dimagi/commcare-hq/pull/33738))

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
+++ b/src/main/java/org/javarosa/core/model/condition/EvaluationContext.java
@@ -256,7 +256,8 @@ public class EvaluationContext {
     public void setVariable(String name, Object value) {
         //No such thing as a null xpath variable. Empty
         //values in XPath just get converted to ""
-        if (value == null) {
+        if (value == null||
+                (value instanceof String && value.toString().trim().isEmpty())) {
             variables.put(name, "");
             return;
         }


shubham1g5 (approved):

Jtang-1:
duplicate this PR 54ef148 54ef148

PR #1406, opened 2024-05-02
Small fix for Stack frame extras

shubham1g5:
## Technical Summary

Changes frame extras from a multi-map of  `String,ListOf(Object)` to `String, Object` . I think it was an oversight that we were handling map values as a listOf(Object) which with multi-map was causing value to be a listOf(ListOf(Object)) for a given key. 

This causes a big confusion today when handling these extras as there are [other places](https://github.com/dimagi/formplayer/blob/master/src/main/java/org/commcare/formplayer/services/MenuSessionFactory.java#L168) where we handle this datastructure correctly (i.e use `putAll` as in this PR on multi-map)

## Safety Assurance

Change radius is limited to smart links with query steps (inline case search). This have [good test coverage](https://github.com/dimagi/formplayer/blob/master/src/test/java/org/commcare/formplayer/tests/MultiSelectCaseClaimTest.java#L339-L394) on FP side. 

### QA Plan

None

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/suite/model/StackFrameStep.java
+++ b/src/main/java/org/commcare/suite/model/StackFrameStep.java
@@ -220,7 +220,7 @@ public class StackFrameStep implements Externalizable {
                 StackFrameStep defined = new StackFrameStep(elementType, id, evaluateValue(ec));
                 extras.forEach((key, value) -> {
                     if (value instanceof QueryData) {
-                        defined.addExtra(key, ((QueryData)value).getValues(ec));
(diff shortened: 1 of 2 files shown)


shubham1g5:
Confirming, I don't see frame step extras being used for any steps other than query or smart link.

snopoke (approved):

shubham1g5:
duplicate this PR ad23db87373ae0d65795751cb8610174fd757621

shubham1g5:
duplicate this PR

PR #1409, opened 2024-05-28
Providing API version header from server to caller when available.

OrangeAndGreen:
Attempts to safely retrieve the x-api-current-version header from API response payloads so calling code can do version checking.
https://dimagi.atlassian.net/browse/CCCT-239

cross-request: https://github.com/dimagi/commcare-android/pull/2757

### DIFF
--- a/src/main/java/org/commcare/core/interfaces/HttpResponseProcessor.java
+++ b/src/main/java/org/commcare/core/interfaces/HttpResponseProcessor.java
@@ -12,7 +12,7 @@ public interface HttpResponseProcessor {
     /**
      * Http response was in the 200s
      */
-    void processSuccess(int responseCode, InputStream responseData);
+    void processSuccess(int responseCode, InputStream responseData, String apiVersion);
 
     /**
      * Http response was in the 400s.
--- a/src/main/java/org/commcare/core/network/ModernHttpRequester.java
+++ b/src/main/java/org/commcare/core/network/ModernHttpRequester.java
@@ -33,7 +33,7 @@ import retrofit2.Response;
  *
  * @author Phillip Mates (pmates@dimagi.com)
  */
-public class ModernHttpRequester implements ResponseStreamAccessor {
+public class ModernHttpRequester {
     /**
      * How long to wait when opening network connection in milliseconds
      */
@@ -87,7 +87,23 @@ public class ModernHttpRequester implements ResponseStreamAccessor {
         }
         try {
             response = makeRequest();
-            processResponse(responseProcessor, response.code(), this);
+            final ModernHttpRequester requester = this;
+            processResponse(responseProcessor, response.code(), new ResponseStreamAccessor() {
+                /**
+                 * Only gets called if response processor is supplied
+                 * @return Input Stream from cache
+                 * @throws IOException if an io error happens while reading or writing to cache
+                 */
+                @Override
+                public InputStream getResponseStream() throws IOException {
+                    return requester.getResponseStream(response);
+                }
+
+                @Override
+                public String getApiVersion() {
+                    return requester.getApiVersion();
+                }
+            });
         } catch (IOException e) {
             e.printStackTrace();
             responseProcessor.handleIOException(e);
@@ -182,17 +198,6 @@ public class ModernHttpRequester implements ResponseStreamAccessor {
         return cache.retrieveCache();
     }
 
-    /**
-     * Only gets called if response processor is supplied
-     * @return Input Stream from cache
-     * @throws IOException if an io error happens while reading or writing to cache
-     */
-    @Override
-    public InputStream getResponseStream() throws IOException {
-        return getResponseStream(response);
-    }
-
-    @Override
     public String getApiVersion() {
         String apiVerison = null;
         try {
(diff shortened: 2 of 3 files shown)


src/main/java/org/commcare/core/interfaces/HttpResponseProcessor.java
```diff
@@ -12,7 +12,7 @@ public interface HttpResponseProcessor {
     /**
      * Http response was in the 200s
      */
-    void processSuccess(int responseCode, InputStream responseData);
+    void processSuccess(int responseCode, InputStream responseData, String apiVersion);
```
shubham1g5, line 15: is `apiVersion` still required here ?

OrangeAndGreen: This is how I planned to expose the value to the calling code. Specifically, after gets/posts in ConnectNetworkHelper I plan to pass the expected version into the getResponseProcessor call, and check it against the received value included in the processSuccess call [here](https://github.com/dimagi/commcare-android/blob/e91160f5f15f4009ebfa3e89d967e630371fd691/app/src/org/commcare/connect/network/ConnectNetworkHelper.java#L341). Does that sound like the right approach?

shubham1g5: I see, I don't think we need to expose it specifically in `processSuccess`, instead implementations of `processSucess` can directly use the `responseData` to get `apiVersion` ?

Also I am confused if `processSuccess` is where callers will be handling `apiVersion` why are we adding `public String getApiVersion() {` in here and ModernHTTPTask ?

OrangeAndGreen: Maybe a better way to phrase my concern with the second approach: If ModernHttpTask is a ResponseStreamAccessor that allows us access to the InputStream and API string, shouldn't we be accessing both the same way?

So if we're going to pass the InputStream via processSuccess instead of calling myModernHttpTask.getResponseStream, I think the API string should be provided the same way.

shubham1g5: > That way access to the response stream and API version would only be available in the callback (I'm going with the proposal to have processSuccess pass the ResponseStreamAccessor). In other words, it wouldn't be possible for calling code to attempt to access the response stream or API string from the task/requester before actually executing the request. 
 
This sounds like a good design choice to me although 

> perhaps instead of implementing ResponseStreamAccessor, both ModernHttpRequester and ModernHttpTask should create and return generic implementations of that interface when calling processResponse.

I am not super clear what you mean by this and having some pseudo code to look at might be helpful.

OrangeAndGreen: No problem. The generic ResponseStreamAccessor would be created when calling processResponse in [ModernHttpTask ](https://github.com/dimagi/commcare-android/blob/776435132ba9f0ad9f3bcca41943fa2ee896c45e/app/src/org/commcare/tasks/ModernHttpTask.java#L90)and [ModernHttpRequester](https://github.com/dimagi/commcare-core/blob/7c8dde6ac6e602468d1802328d75ef0d2faf095a/src/main/java/org/commcare/core/network/ModernHttpRequester.java#L90), rather than returning `this` as the last function input.

For ModernHttpTask, the new code would look like:
image

The class would no longer implement `ResponseStreamAccessor`, and the `getResponseStream` and `getApiVersion` methods at the end of the file would be deleted.

And in places like [this](https://github.com/dimagi/commcare-android/blob/776435132ba9f0ad9f3bcca41943fa2ee896c45e/app/src/org/commcare/activities/PostRequestActivity.java#L120C17-L120C25), it would no longer be possible to call `postTask.getResponseStream` before calling `connect` and `executeParallel`. Note that I currently don't see this happening anywhere in the code, but the potential is there.

shubham1g5: yes, that seems promising to me 👍

OrangeAndGreen: Implemented these changes

src/main/java/org/commcare/core/network/ModernHttpRequester.java
```diff
@@ -181,14 +198,14 @@ public InputStream getResponseStream(Response<ResponseBody> response) throws IOE
...
-     */
-    @Override
-    public InputStream getResponseStream() throws IOException {
-        return getResponseStream(response);
+    public String getApiVersion() {
+        String apiVerison = null;
+        try {
+            apiVerison = response.headers().get("x-api-current-version");
+        } catch(Exception e) {
+            //Ignore exception if API version header isn't present
```
shubham1g5, line 206: the call `response.headers().get("x-api-current-version")` doesn't really throw an exception even if `"x-api-current-version` is not present, as such do we need the catch here ?

shubham1g5 (changes requested):

shubham1g5 (approved):
Approving as don't want to block this further but left one comment to address in previous review.

src/main/java/org/commcare/core/network/ModernHttpRequester.java
OrangeAndGreen: Makes sense, I removed it. I was being a little overly safe!

shubham1g5 (approved):

PR #1414, opened 2024-06-28
Rc/core make widow width available in session

Robert-Costello:
## Technical Summary

[jira ticket](https://dimagi.atlassian.net/browse/USH-4671)
[Related Formplayer PR](https://github.com/dimagi/formplayer/pull/1596)
[Related Web Apps PR  ](https://github.com/dimagi/commcare-hq/pull/34832)

### QA Plan

[QA Ticket](https://dimagi.atlassian.net/browse/QA-6751)

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/modern/session/SessionWrapper.java
+++ b/src/main/java/org/commcare/modern/session/SessionWrapper.java
@@ -24,6 +24,10 @@ public class SessionWrapper extends CommCareSession implements SessionWrapperInt
     final protected CommCarePlatform mPlatform;
     protected CommCareInstanceInitializer initializer;
     protected RemoteInstanceFetcher remoteInstanceFetcher;
+    /**
+     * A string representing the width of the user's screen in pixels.
+     * To be used in a display condition determining what content to show to the user.
+     */
     private String windowWidth;
 
     public SessionWrapper(CommCareSession session, CommCarePlatform platform, UserSandbox sandbox,
--- a/src/main/java/org/commcare/modern/session/SessionWrapper.java
+++ b/src/main/java/org/commcare/modern/session/SessionWrapper.java
@@ -24,17 +24,19 @@ public class SessionWrapper extends CommCareSession implements SessionWrapperInt
     final protected CommCarePlatform mPlatform;
     protected CommCareInstanceInitializer initializer;
     protected RemoteInstanceFetcher remoteInstanceFetcher;
+    private String windowWidth;
 
     public SessionWrapper(CommCareSession session, CommCarePlatform platform, UserSandbox sandbox,
-            RemoteInstanceFetcher remoteInstanceFetcher) {
-        this(platform, sandbox, remoteInstanceFetcher);
+            RemoteInstanceFetcher remoteInstanceFetcher, String windowWidth) {
+        this(platform, sandbox, remoteInstanceFetcher, windowWidth);
         this.frame = session.getFrame();
         this.setFrameStack(session.getFrameStack());
+        this.windowWidth = windowWidth;
     }
 
 
-    public SessionWrapper(CommCareSession session, CommCarePlatform platform, UserSandbox sandbox) {
-        this(session, platform, sandbox, null);
+    public SessionWrapper(CommCareSession session, CommCarePlatform platform, UserSandbox sandbox, String windowWidth) {
+        this(session, platform, sandbox, null, windowWidth);
     }
     
     public SessionWrapper(CommCarePlatform platform, UserSandbox sandbox) {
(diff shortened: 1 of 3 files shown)


shubham1g5 (changes requested):
looks good though we should be adding tests here or on FP to demonstrate the sample usage of this parameter.

Robert-Costello:
>we should be adding tests here or on FP to demonstrate the sample usage of this parameter.

Yes, definitely. I wanted to make sure this approach was making sense before adding a test/tests. 
What are we looking for in a test for a single datum like this, in high level terms? Do we want to add `windowWidth` to a mock query and verify that it gets added into the menu/form sessions as expected? Or a more narrow unit test where it's in a test session already and we're just retrieving and verifying the value?
One test that seems somewhat related is this [testSessionDatumParser](https://github.com/dimagi/commcare-core/blob/a84395916af663e6dfda3732d4b70c5792391676/src/test/java/org/commcare/xml/SessionDatumParserTest.java#L22) that looks at `device_id`.

shubham1g5:
> Do we want to add windowWidth to a mock query and verify that it gets added into the menu/form sessions as expected? 
Yes, I think this is what we want.

Robert-Costello:
duplicate this PR

Robert-Costello:
@shubham1g5 I'd like to go ahead and merge this, since the test in the FP PR depends on the changes here. How does that sound to you?

shubham1g5:
It will break the `master` builds on FP if we merge this without merging FP changes as well, it's fine for tests on FP PR to fail until we can finalise things on  that PR. When we are ready to merge we should merge this first and then before merging FP side get the FP build passes with these changes.

src/main/java/org/commcare/session/SessionInstanceBuilder.java
```diff
@@ -88,14 +88,15 @@ private static String getCalloutSearchResultCount(StackFrameStep step) {
...
-                                    String userId, long drift) {
+                                    String userId, long drift, String windowWidth) {
         TreeElement sessionMeta = new TreeElement("context", 0);
 
         addData(sessionMeta, "deviceid", deviceId);
         addData(sessionMeta, "appversion", appversion);
         addData(sessionMeta, "username", username);
         addData(sessionMeta, "userid", userId);
         addData(sessionMeta, "drift", String.valueOf(drift));
+        addData(sessionMeta, "windowwidth", windowWidth);
```
shubham1g5, line 99: `windowwidth` troubles me a bit, can we underscore it as `window_width` although inconsistent with other params.

Robert-Costello: sure, I'll change it to window_width

src/main/java/org/commcare/modern/session/SessionWrapper.java
```diff
@@ -24,17 +24,23 @@ public class SessionWrapper extends CommCareSession implements SessionWrapperInt
...
+    private String windowWidth;
 
     public SessionWrapper(CommCareSession session, CommCarePlatform platform, UserSandbox sandbox,
-            RemoteInstanceFetcher remoteInstanceFetcher) {
-        this(platform, sandbox, remoteInstanceFetcher);
+            RemoteInstanceFetcher remoteInstanceFetcher, String windowWidth) {
+        this(platform, sandbox, remoteInstanceFetcher, windowWidth);
         this.frame = session.getFrame();
         this.setFrameStack(session.getFrameStack());
+        this.windowWidth = windowWidth;
```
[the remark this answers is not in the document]
Robert-Costello: yes, I think you're right

shubham1g5 (approved):
Looks good, though we should address https://github.com/dimagi/commcare-core/pull/1414/files#r1705888450 before merging.

MartinRiese (approved):

ctsims:
Sorry for the late jump in on this, but was the documentation updated in the core repos for how this new field works and what the value sets are? There's lot of discussion in the specs but I didn't see the outcome become clear anywhere.

This feature has a very, very high probability of creating confusing behaviors where two people receive different experiences in ways that are extremely opaque, and being overloaded for different purposes, so I want to make sure that we have all of the ways it works be extremely unambiguous.

Robert-Costello:
The docs need to be updated here https://github.com/dimagi/commcare-core/wiki/commcaresession. I'll update it today.
Differing behaviors will come from app building that specifically utilizes this property in display conditions and I'll make that unambiguous in the docs. 

@shubham1g5 @MartinRiese Is there any other place you're aware of that documentation that should be updated?

PR #1417, opened 2024-07-03
Fix repeat deletion Issue by implementing Clear volatiles for tree element

shubham1g5:
## Technical Summary

https://dimagi.atlassian.net/browse/USH-4708 (Deleting a repeat was causing errors in resolving references for remaining repeat nodes)

After deleting a repeat, a node's reference can change due to change in repeat index of the node. Therefore we need to expire the existing cached references for TreeElements. 

### Safety story

- Should only affect repeat deletion which is covered by test

### Automated test coverage
 
 Added in the PR plus in https://github.com/dimagi/formplayer/pull/1597

### QA Plan
None

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
+++ b/src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
@@ -200,8 +200,11 @@ public abstract class StorageBackedChildElement<Model extends Externalizable>
 
     @Override
     public void clearVolatiles() {
-        // we should also clear the other cache implementation in this class in future
         ref = null;
+        TreeElement cache = cache();
+        if (cache != null) {
+            cache.clearVolatiles();
+        }
     }
 
     //Context Sensitive Methods
--- a/src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
+++ b/src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
@@ -200,6 +200,7 @@ public abstract class StorageBackedChildElement<Model extends Externalizable>
 
     @Override
     public void clearVolatiles() {
+        // we should also clear the other cache implementation in this class in future
         ref = null;
     }
 
--- a/src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
+++ b/src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
@@ -198,6 +198,10 @@ public abstract class StorageBackedChildElement<Model extends Externalizable>
         return ref;
     }
 
+    @Override
+    public void clearVolatiles() {
+        ref = null;
+    }
 
     //Context Sensitive Methods
     @Override
--- a/src/main/java/org/javarosa/core/model/instance/DataInstance.java
+++ b/src/main/java/org/javarosa/core/model/instance/DataInstance.java
@@ -291,5 +291,16 @@ public abstract class DataInstance<T extends AbstractTreeElement<T>> implements 
 
     public void cleanCache() {
         referenceCache.clear();
+        cleanTreeElementCache(getBase());
+    }
+
+    private void cleanTreeElementCache(AbstractTreeElement<T> node) {
+        if (node == null) {
+            return;
+        }
+        node.clearVolatiles();
+        for (int i = 0; i < node.getNumChildren(); i++) {
+            cleanTreeElementCache(node.getChildAt(i));
+        }
     }
 }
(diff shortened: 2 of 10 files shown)


ctsims:
I think we should be moving the logic for dispatching the volatile clearing from the [DataInstance](https://github.com/dimagi/commcare-core/pull/1417/files#diff-69efdff5cefbee28593cfa0cf1b36fb2ad9531f0e5f0da534bbc2596e46da232R309) into each [implementation of the method](https://github.com/dimagi/commcare-core/pull/1417/files#diff-da522826a7f034350d7896d83c36793c9bd8d93d205c342d623164c786569a3cR341), where each implementation is required to both clear its own volatile state and dispatch the clear call to its children and other methods. 

There are two big reasons why.
The biggest reason is that not all implementations of AbstractTreeElement actually have real child nodes. getChild() sometimes returns real elements, but also sometimes returns [templates](https://github.com/dimagi/commcare-core/pull/1417/files#diff-47715811598c0e35e374e7f7a0e4c6c527945f83834beb87970135767ea5b2f6R82)  or virtual elements (say, for Repeat Indexing), so the "Get child count" and "dispatch to each child by multiplicity" approach could either end up triggering unintended consequences (walking the repeat tree and creating repeats, for instance), could return confusing elements, could miss data (say, nodes with a multiplicity of -2), or could just end up being slower than necessary
Additionally, right now there's just something a little awkward about requiring a DataInstance root for dispatching the tree walk in the first place.

I checked, and I don't think this will actually create meaningfully more repetitive implementations. The only thing this change makes a little confusing is that I think the default implementation of [clearCache](https://github.com/dimagi/commcare-core/pull/1417/files#diff-69efdff5cefbee28593cfa0cf1b36fb2ad9531f0e5f0da534bbc2596e46da232R293)  should definitely call getBase().clearVolatiles(), so we either need a second implementation of cleanCache that gives it a different target, or we should just take the hit and clear all of the caches

src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java
```diff
@@ -198,6 +198,11 @@ public TreeReference getRef() {
         return ref;
     }
 
+    @Override
+    public void clearVolatiles() {
+        // we should also clear the other cache implementation in this class in future
```
shubham1g5, line 203: @ctsims Think it's for me to clear this cache now but I can't find a way to do it without using something like - 

````
for (int i = 0; i < getNumChildren(); i++) {
            getChildAt(i).clearVolatiles();
}
````

Comment (shubham1g5) [reply src/main/java/org/commcare/cases/instance/StorageBackedChildElement.java]: nvm, It was simple.

**Review by ctsims — APPROVED** [review]
Looks good to me after the revisions! Thanks for getting this implemented across the models.

**Comment by shubham1g5** [comment]
duplicate this branch 00c3d8a5ce94c4bc6e518036de180306d8f0401b 9302ea1481eddfdc64ea199c9e8875f28954af56

File: src/main/java/org/javarosa/core/model/instance/DataInstance.java [hunk 2]
```diff
@@ -289,7 +289,11 @@ public void setCacheHost(CacheHost cacheHost) {
         this.mCacheHost = cacheHost;
     }
 
+    /**
+     * Cleans reference caches maintained by the instance and TreeElements contained in the contextNode
+     */
     public void cleanCache() {
         referenceCache.clear();
+        getBase().clearVolatiles();
```
snopoke, line 297: Just to clarify, this `cleanCache` function is already being called in the correct way and we're just adding more cleanup code. Is that correct?

shubham1g5: yes, for this particular use case of deleting repeat it gets called [here](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/FormDef.java#L376)

shubham1g5:
duplicate this PR https://github.com/dimagi/commcare-core/commit/00c3d8a5ce94c4bc6e518036de180306d8f0401b https://github.com/dimagi/commcare-core/commit/9302ea1481eddfdc64ea199c9e8875f28954af56

shubham1g5:
duplicate this PR https://github.com/dimagi/commcare-core/commit/00c3d8a5ce94c4bc6e518036de180306d8f0401b https://github.com/dimagi/commcare-core/commit/9302ea1481eddfdc64ea199c9e8875f28954af56

PR #1424, opened 2024-08-07
Make FormIndex serializable

avazirna:
## Product Description
This PR makes the `FormIndex` class serializable in order to save its objects in SharedPreferences and allow an automatic restoration of the user session when the session expires or CommCare is force closed.

## Technical Summary
Context: [Graceful Session Pause](https://docs.google.com/document/d/1j5ume0J0UwwraijxGw-Pq-8ZYFC7L07cKb7CXL30poo/edit)

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormIndex.java
+++ b/src/main/java/org/javarosa/core/model/FormIndex.java
@@ -472,7 +472,7 @@ public class FormIndex implements Externalizable {
         endOfForm = ExtUtil.readBool(in);
         localIndex = ExtUtil.readInt(in);
         instanceIndex = ExtUtil.readInt(in);
-        reference = (TreeReference)ExtUtil.read(in, TreeReference.class, pf);
+        reference = (TreeReference)ExtUtil.read(in, new ExtWrapNullable(TreeReference.class), pf);
         nextLevel = (FormIndex)ExtUtil.read(in, new ExtWrapNullable(FormIndex.class), pf);
     }
 
@@ -482,7 +482,7 @@ public class FormIndex implements Externalizable {
         ExtUtil.writeBool(out, endOfForm);
         ExtUtil.writeNumeric(out, localIndex);
         ExtUtil.writeNumeric(out, instanceIndex);
-        ExtUtil.write(out, reference);
+        ExtUtil.write(out, new ExtWrapNullable(reference));
         ExtUtil.write(out, new ExtWrapNullable(nextLevel));
     }
 }
--- a/src/main/java/org/javarosa/core/model/FormIndex.java
+++ b/src/main/java/org/javarosa/core/model/FormIndex.java
@@ -1,7 +1,15 @@
 package org.javarosa.core.model;
 
 import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.core.util.externalizable.DeserializationException;
+import org.javarosa.core.util.externalizable.ExtUtil;
+import org.javarosa.core.util.externalizable.ExtWrapNullable;
+import org.javarosa.core.util.externalizable.Externalizable;
+import org.javarosa.core.util.externalizable.PrototypeFactory;
 
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
 import java.util.Vector;
 
 /**
@@ -22,7 +30,7 @@ import java.util.Vector;
  *
  * @author Clayton Sims
  */
-public class FormIndex {
+public class FormIndex implements Externalizable {
 
     private boolean beginningOfForm = false;
     private boolean endOfForm = false;
@@ -30,7 +38,7 @@ public class FormIndex {
     /**
      * The index of the questiondef in the current context
      */
-    private final int localIndex;
+    private int localIndex;
 
     /**
      * The multiplicity of the current instance of a repeated question or group
@@ -43,6 +51,10 @@ public class FormIndex {
     private FormIndex nextLevel;
 
     private TreeReference reference;
+
+    // needed for serialization
+    public FormIndex(){
+    }
 
     public static FormIndex createBeginningOfFormIndex() {
         FormIndex begin = new FormIndex(-1, null);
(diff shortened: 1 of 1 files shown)


src/main/java/org/javarosa/core/model/FormIndex.java
```diff
@@ -453,4 +465,24 @@ public void assignRefs(FormDef f) {
...
         }
     }
+
+    @Override
+    public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
+        beginningOfForm = ExtUtil.readBool(in);
+        endOfForm = ExtUtil.readBool(in);
+        localIndex = ExtUtil.readInt(in);
+        instanceIndex = ExtUtil.readInt(in);
+        reference = (TreeReference)ExtUtil.read(in, TreeReference.class, pf);
```
shubham1g5, line 475: should this be wrapped with `ExtWrapNullable` as well ? I can see it's `null` for atleast beginning and ending indexes.

shubham1g5 (changes requested):

src/main/java/org/javarosa/core/model/FormIndex.java
avazirna: I had it wrapped before but during tests I noticed that when `reference` is `null` the logic to return to the form index throws a NPE and I didn't want to change that behaviour. I'm open to change this

shubham1g5: > when reference is null the logic to return to the form index throws a NPE
Do we need to solve for that before merging this work ? Or this NPE crash would not happen for users ?

And think serialization logic here should not depent on what happens this class and if a field is nullable it must be wrapped correctly with `ExtWrapNullable` ?

avazirna: > Do we need to solve for that before merging this work ? Or this NPE crash would not happen for users ?

From what I could see, it's not expected for a `FormIndex` object to not have a reference. I can run more tests around this to see if that's 100%. 

> And think serialization logic here should not depent on what happens this class and if a field is nullable it must be wrapped correctly with ExtWrapNullable ?
True

shubham1g5: >  it's not expected for a FormIndex object to not have a reference

What about [beginning of Form Index ](https://github.com/dimagi/commcare-core/blob/master/src/main/java/org/javarosa/core/model/FormIndex.java#L47) ? I think we should be safely handling NPE atleast to not crash the app on moving to that index.

shubham1g5 (approved):

src/main/java/org/javarosa/core/model/FormIndex.java
avazirna: @shubham the beginning and end of Form Indexes seem to be temporary states, part of loading cycles during form loading, question transitioning and form completion. During those cycles, [buildIndex(...)](https://github.com/dimagi/commcare-core/blob/db6ecb4fad6c0c15513d6a3880d87ea04edb7f43/src/main/java/org/javarosa/core/model/FormDef.java#L1539) is called which instantiates the `reference` so, it doesn't seem likely for a question to be displayed with `reference` `null`. As per the NPE, here's the stack trace:
```
java.lang.RuntimeException: Unable to stop activity {org.commcare.dalvik.debug/org.commcare.activities.FormEntryActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.javarosa.core.model.instance.TreeReference.isAbsolute()' on a null object reference
      at android.app.ActivityThread.callActivityOnStop(ActivityThread.java:5713)
      at android.app.ActivityThread.performStopActivityInner(ActivityThread.java:5649)
      at android.app.ActivityThread.handleStopActivity(ActivityThread.java:5750)
      at android.app.servertransaction.StopActivityItem.execute(StopActivityItem.java:43)
      at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:45)
      at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
      at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
      at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2438)
      at android.os.Handler.dispatchMessage(Handler.java:106)
      at android.os.Looper.loopOnce(Looper.java:226)
      at android.os.Looper.loop(Looper.java:313)
      at android.app.ActivityThread.main(ActivityThread.java:8663)
      at java.lang.reflect.Method.invoke(Native Method)
      at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1135)
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.javarosa.core.model.instance.TreeReference.isAbsolute()' on a null object reference
      at org.javarosa.core.model.instance.DataInstance.resolveReference(DataInstance.java:90)
      at org.javarosa.core.model.instance.DataInstance.resolveReference(DataInstance.java:86)
      at org.javarosa.form.api.FormEntryModel.getTreeElement(FormEntryModel.java:110)
      at org.javarosa.form.api.FormEntryController.saveAnswer(FormEntryController.java:172)
      at org.javarosa.form.api.FormController.saveAnswer(FormController.java:182)
      at org.commcare.activities.FormEntryActivity.saveAnswer(FormEntryActivity.java:1356)
      at org.commcare.activities.FormEntryActivity.saveAnswersForCurrentScreen(FormEntryActivity.java:656)
      at org.commcare.activities.FormEntryActivity.saveDataToDisk(FormEntryActivity.java:828)
      at org.commcare.activities.FormEntryActivity.saveIncompleteFormToDisk(FormEntryActivity.java:782)
      at org.commcare.activities.FormEntryActivity.interruptAndSaveForm(FormEntryActivity.java:267)
      at org.commcare.activities.FormEntryActivity.onStop(FormEntryActivity.java:931)
      at android.app.Instrumentation.callActivityOnStop(Instrumentation.java:1496)
```
it's thrown when calling [ref.isAbsolute()](https://github.com/dimagi/commcare-core/blob/db6ecb4fad6c0c15513d6a3880d87ea04edb7f43/src/main/java/org/javarosa/core/model/instance/DataInstance.java#L90) during form loading however, this only happens when I purposely save the `reference` as null, I couldn't replicate that issue otherwise.

shubham1g5: Ahh that makes sense, I thought you are able to repro during normal app workflow. Thanks for looking into this.

PR #1431, opened 2024-09-03
Fix Levenshtein distance algorithm

avazirna:
## Product Description
This PR fixes a small issue with the Levenshtein distance calculation, which was causing the fuzzy search to return false positives. For instance, the difference between `aply` into `cape` was previously 2 while it should be 3, and because our distance threshold is set at 2, the cape was a valid value for `aply`.

Another look to the [source of the algorithm](https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Levenshtein_distance), I could also confirm that this [was initially incorrect](https://en.wikibooks.org/w/index.php?title=Algorithm_Implementation/Strings/Levenshtein_distance&oldid=2612649#Java):
```
		// initial cost of skipping prefix in String s1
		newcost[0]=j-1;
```
 and later on updated to the current form:
```
		// initial cost of skipping prefix in String s1
		newcost[0]=j;
```
</td>
</tr>
</table>

Ticket: https://dimagi.atlassian.net/browse/SAAS-15738

### Automated test coverage
There are unit tests around this feature and they seem to have not been affected by this change.

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/cases/util/StringUtils.java
+++ b/src/main/java/org/commcare/cases/util/StringUtils.java
@@ -119,7 +119,7 @@ public class StringUtils {
         for (int j = 1; j < len1; j++) {
 
             // initial cost of skipping prefix in String s1
-            newcost[0] = j - 1;
+            newcost[0] = j;
 
             // transformation cost for each letter in s0
             for (int i = 1; i < len0; i++) {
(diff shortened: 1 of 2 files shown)


shubham1g5 (changes requested):
Looks good to me though can we write a unit test for the particular edge case we are trying to fix here.

shubham1g5 (approved):

avazirna:
Duplicate PR

avazirna:
duplicate this PR

PR #1433, opened 2024-09-05
Sync screen session status

Jtang-1:
## Technical Summary

Helper function for https://github.com/dimagi/formplayer/pull/1622

### QA Plan

no QA

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/cli/java/org/commcare/util/screen/SyncScreen.java
+++ b/src/cli/java/org/commcare/util/screen/SyncScreen.java
@@ -125,6 +125,10 @@ public class SyncScreen extends Screen {
         sessionWrapper.addExtraToCurrentFrameStep(EXTRA_POST_SUCCESS, true);
     }
 
+    public Object getSessionSuccessStatus() {
+        return sessionWrapper.getCurrentFrameStepExtra(EXTRA_POST_SUCCESS);
+    }
+
     @Override
     public String[] getOptions() {
         return new String[0];
--- a/src/cli/java/org/commcare/util/screen/SyncScreen.java
+++ b/src/cli/java/org/commcare/util/screen/SyncScreen.java
@@ -125,8 +125,8 @@ public class SyncScreen extends Screen {
         sessionWrapper.addExtraToCurrentFrameStep(EXTRA_POST_SUCCESS, true);
     }
 
-    public Boolean isPostAlreadyDone() {
-        return sessionWrapper.getCurrentFrameStepExtra(EXTRA_POST_SUCCESS) == null ;
+    public Boolean isPostAlreadyAttempted() {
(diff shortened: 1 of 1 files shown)


src/cli/java/org/commcare/util/screen/SyncScreen.java
```diff
@@ -125,6 +125,10 @@ public void updateSessionOnSuccess() {
         sessionWrapper.addExtraToCurrentFrameStep(EXTRA_POST_SUCCESS, true);
     }
 
+    public Object getSessionSuccessStatus() {
```
shubham1g5, line 128: think we should return a boolean here ?

shubham1g5 (changes requested):

src/cli/java/org/commcare/util/screen/SyncScreen.java
Jtang-1: I was thinking about that, but that means we would need to also check if `sessionWrapper.getCurrentFrameStepExtra(EXTRA_POST_SUCCESS);` is null and return `false` in that case. Would `false` and `null` always mean the same thing? Or would there be cases where we would want to differentiate the two?

shubham1g5: yeah think `false` and `null` are same things here as they both signify that no post request has been successful so far.

Jtang-1: mmm i'm thinking they don't necessarily mean the same thing. Since `null` means no post request has been attempted yet while `false` means a post request has been attempted but not successful. 

I think this distinction matters because for this use case, if it is already set to `false`, then calling `updateSessionOnSuccess` will result in this [error](https://github.com/dimagi/commcare/blob/ad23db87373ae0d65795751cb8610174fd757621/src/main/java/org/commcare/suite/model/StackFrameStep.java#L176-L179). So that will need to be handled differently than `null`

shubham1g5: I see, that makes sense to me. Maybe we can have this method as below instead to represent more literally what we are using this to check - 

````
public Boolean isPostAlreadyDone() {
return sessionWrapper.getCurrentFrameStepExtra(EXTRA_POST_SUCCESS) == null ;
}

````

Jtang-1: ah ya i like that suggestion. thank you https://github.com/dimagi/commcare-core/pull/1433/commits/5fa972268075657ba740298b3415c3c67c41c39d

shubham1g5 (approved):

shubham1g5 (approved):

Jtang-1:
duplicate this PR 030a6d5 dc30183

Jtang-1:
duplicate this PR 030a6d5 dc30183

PR #1446, opened 2024-10-17
Restored 3-input constructor for SessionWrapper (to preserve mobile b…

OrangeAndGreen:
Restored 3-input constructor for SessionWrapper (to preserve mobile build).

This change is to fix the mobile build, after a recent commit changed a constructor for SessionWrapper to take a new field called windowWidth. This field is only stored by the wrapper for later external access, and is never used by the mobile code. So I added a new 3-input constructor that then passes null for the windowWidth.


avazirna:
@OrangeAndGreen it might be better to lean on the consistent side, why not pass the window width on mobile too instead of setting it to `null`? It sounds that there is no use for it, but considering that the session will now carry this property we might want to set an actual value.

shubham1g5 (approved):
Added an issue to pass window width - https://github.com/dimagi/commcare-android/issues/2878  as a separate item, and going to merge this to get the `master` builds passing. In general, we should be treating the `master` build failures as a critical event and should be fixing those promptly as it affects other external and internal devs working on the project.

PR #1267, opened 2023-04-19
add handling assertions in menus

Robert-Costello:
This is a duplicate of https://github.com/dimagi/commcare-core/pull/1266 for formplayer branch. I missed the boat a little here with regards to the new duplication procedure that Jonathan implemented. It looks like it shouldn't be a problem to run the android tests. I've just done things a slightly out of order. 
## Product Description

This handles custom assertions in menus. Menus will be displayed or not depending on the evaluation of the assertions.

## Technical Summary

https://docs.google.com/document/d/112St-i7h7jZHpqxcAU-d6VASftXKLfiFnqkS3ovsEVU/edit#heading=h.37a7vgf78nog
Jira ticket:  https://dimagi-dev.atlassian.net/browse/USH-2983
Jira epic with this and HQ tickets: https://dimagi-dev.atlassian.net/browse/USH-2905

### Safety story

Feature Flagged (CUSTOM_ASSERTIONS: Inject custom assertions into the suite), will go through QA

### Automated test coverage

tests add here

### QA Plan

will be tested in QA along with required HQ changes

### Rollback instructions

- [x] This PR can be reverted after deploy with no further considerations.

### Review

- [x] The set of people pinged as reviewers is appropriate for the level of risk of the change.

PR #1448, opened 2024-10-24
Fix menu instances Accumulation for menus with same ids

shubham1g5:
## Product Description

https://dimagi.atlassian.net/browse/SC-3993

Fixes a regression with CommCare 2.54 that caused incorrect menu instances evaluation when there are multiple menus with same id in the suite.xml

## Technical Summary

This is a regression that was introduced with custom assertions work ([FP](https://github.com/dimagi/commcare-core/pull/1267) and [HQ](https://github.com/dimagi/commcare-hq/pull/32842)) which re-organised how we define instances for menus in suite.xml. With these changes, we were only looking for menu instances in the first menu with a specific id. So if there is a suite config like below - 

````
<menu id="root" relevant="instance('casedb')/casedb/case[@case_type ='commcare-user']/date_of_last_submission != today()" style="grid">
    <text>
      <locale id="modules.m0"/>
    </text>
    <command id="m0-f0"/>
</menu>

<menu id="root" relevant="instance('item-list:polio_app_settings')/polio_app_settings_list/polio_app_settings[ID = 'settings']/show_registration = 1" style="grid">    
    <command id="m1-f0"/>
    <instance id="item-list:polio_app_settings" src="jr://fixture/item-list:polio_app_settings"/>
</menu>
````

while evaluating the relevant condition for second menu with `instance('item-list:polio_app_settings')`, we were only adding instances from the first menu (casedb) and skipping the instances from relevant menu resulting into an instance not found error. 

This PR makes a change to look for instances in all menus with a given command id and not only the first one.

### Safety story

- This is an additive change which adds more instances to the eval context which should not have any side effects while calculating xpath.  I am not worried about performance either with this change as the code to populate instance should be a quick O(n) loop without any exponentiality. 

- Tested Locally that the change indeed fix the issue described in the JIra linked above. 

### Automated test coverage

PR adds a test to verify the issue

### QA Plan

No QA

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/commcare/session/CommCareSession.java
+++ b/src/main/java/org/commcare/session/CommCareSession.java
@@ -633,7 +633,6 @@ public class CommCareSession {
         Vector<Entry> entries = getEntriesForCommand(command);
         Vector<Menu> menus = getMenusForCommand(command);
 
-        Menu menu = null;
         Entry entry = null;
         Hashtable<String, DataInstance> instancesInScope = new Hashtable<>();
         Hashtable<String, DataInstance> menuInstances = null;
@@ -649,8 +648,7 @@ public class CommCareSession {
             }
         }
 
-        if (!menus.isEmpty()) {
-            menu = menus.elementAt(0);
+        for (Menu menu : menus) {
             if (menu != null) {
                 menuInstances = menu.getInstances(instancesToInclude);
             }
--- a/src/test/java/org/commcare/backend/session/test/MenuTests.java
+++ b/src/test/java/org/commcare/backend/session/test/MenuTests.java
@@ -2,7 +2,10 @@ package org.commcare.backend.session.test;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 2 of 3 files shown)


Robert-Costello (approved):
Did this issue not pop up because no apps at the time were configured in a way that this bug would pop up? It also seems like something that would have come up in QA, but maybe the testing apps were also configured so that this bug didn't cause issues.

avazirna (approved):
It seems that this is mainly applicable to the _Display only forms_ feature which makes the menu id to be set to _root_. Are there any other features affected by this?

shubham1g5:
> Did this issue not pop up because no apps at the time were configured in a way that this bug would pop up?

@Robert-Costello Yeah i think so, also the fact that CC 2.54 got released recently and this issue only surface when an app targets CC 2.54 in app settings along with a particular config. 

> It seems that this is mainly applicable to the Display only forms feature which makes the menu id to be set to root. Are there any other features affected by this?

I am not sure if there are other features which would cause menus with same ids but yeah at-least display only forms.

PR #1454, opened 2025-01-03
Cache and Lazy Loading Related Modifications

shubham1g5:
## Product Description

Supporting changes for https://github.com/dimagi/commcare-android/pull/2928

- Adds entity loading cancellation and progress hooks 
- Supporting methods to turn on cache and lazy loading for an entity list with the help of the [new models detailed here](https://docs.google.com/document/d/1VE6yjo-yOfhnwpHMRZWavgJmbmbRhsoVvWS3Kq3yh2g/edit?tab=t.0#heading=h.7ysxtar2iceb)
- Changes to `AsyncEntity` to support seletive caching and lazy-loading for only optimizable fields

### Safety story

Doesn't introduce any changes in existing behaviour

### QA Plan

Will go through UAT but possibly after merge. 

### Rollback instructions

### Review

### Duplicate PR

## Release Notes

- **New Features**
	- Introduced advanced caching and performance optimization mechanisms for entity loading
	- Added progress tracking for entity loading processes
	- Enhanced detail view configuration with new performance attributes

- **Improvements**
	- Implemented cancellation support for asynchronous entity loading
	- Refined field evaluation and caching strategies
	- Added optimization options for detail fields

- **Technical Updates**
	- Updated parsing logic for detail and field configurations
	- Introduced new interfaces for tracking entity loading progress
	- Modified serialization methods to support new caching features

### DIFF
--- a/src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
+++ b/src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
@@ -124,7 +124,7 @@ public class AsyncNodeEntityFactory extends NodeEntityFactory {
     }
 
     @Override
-    public void cacheEntities(List<Entity<TreeReference>> entities, Boolean inBackground) {
+    public void cacheEntities(List<Entity<TreeReference>> entities, boolean inBackground) {
         if (detail.isCacheEnabled()) {
             primeCache();
             setUnCachedData(entities, inBackground);
@@ -134,7 +134,7 @@ public class AsyncNodeEntityFactory extends NodeEntityFactory {
         }
     }
 
-    protected void setUnCachedData(List<Entity<TreeReference>> entities, Boolean inBackground) {
+    protected void setUnCachedData(List<Entity<TreeReference>> entities, boolean inBackground) {
         for (int i = 0; i < entities.size(); i++) {
             if (isCancelled) return;
             AsyncEntity e = (AsyncEntity)entities.get(i);
--- a/src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
+++ b/src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
@@ -57,7 +57,7 @@ public class AsyncNodeEntityFactory extends NodeEntityFactory {
         String entityKey = loadCalloutDataMapKey(nodeContext);
         AsyncEntity entity =
                 new AsyncEntity(detail.getFields(), nodeContext, data, mVariableDeclarations,
-                        mEntityCache, mCacheIndex, detail.getId(), entityKey, detail.getGroup());
+                        mEntityCache, mCacheIndex, detail.getId(), entityKey, detail.getGroup(), detail.isCacheEnabled());
 
         if (mCacheIndex != null) {
             mEntitySet.put(mCacheIndex, entity);
@@ -106,7 +106,7 @@ public class AsyncNodeEntityFactory extends NodeEntityFactory {
         // if blocking mode load cache on the same thread and set any data thats not cached
         if (isBlockingAsyncMode) {
             primeCache();
-            setUnCachedData(entities);
+            setUnCachedDataOld(entities);
         } else {
             // otherwise we want to show the entity list asap and hence want to offload the loading cache part to a separate
             // thread while caching any uncached data later on UI thread during Adapter's getView
@@ -121,8 +121,13 @@ public class AsyncNodeEntityFactory extends NodeEntityFactory {
 
     @Override
     public void cacheEntities(List<Entity<TreeReference>> entities) {
-        primeCache();
-        setUnCachedData(entities);
+        if (detail.isCacheEnabled()) {
+            primeCache();
+            setUnCachedData(entities);
+        } else {
+            primeCache();
+            setUnCachedDataOld(entities);
+        }
     }
 
     protected void setUnCachedData(List<Entity<TreeReference>> entities) {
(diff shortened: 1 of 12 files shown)


shubham1g5:
@coderabbitai full review

shubham1g5:

src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
```diff
@@ -78,7 +81,12 @@ protected void setEvaluationContextDefaultQuerySet(EvaluationContext ec,
      * Bulk loads search field cache from db.
      * Note that the cache is lazily built upon first case list search.
      */
-    private void primeCache() {
+    protected void primeCache() {
+        if (isCancelled) return;
+        if (progressListener != null) {
```
ctsims, line 86: Will it cause issues for this progress update to occur before all of the null checks below that dump out of the method and/or without a `finally {}` clause? It seems it it would keep the listener spinning forever.

src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
```diff
@@ -192,4 +194,34 @@ public final boolean isEntitySetReady() {
...
+        throw new RuntimeException("Method not supported for normal Node Entity Factory");
+    }
+
+    /**
+     * Sets the progress listener for entity loading operations.
+     *
+     * @param progressListener The progress listener to use
+     */
+    public void setEntityProgressListener(EntityLoadingProgressListener progressListener) {
+        this.progressListener = progressListener;
```
ctsims, line 215: If there is an existing listener set, I think this method should close it out before changing the current one. Alternatively this method should potentially error out, or otherwise treat the listener as immutable once set.

ctsims (dismissed):
These changes all seem clear / appropriate for the stated spec. Left a couple of points of feedback

Knowing that part of the goal here is long term to be managing entity loading over more flexible lifecycles, I'm a little worried that issues with managing the listener lifecycle here could be an issue.

src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
shubham1g5: https://github.com/dimagi/commcare-core/pull/1454/commits/5a1422e017dec6897375e3c0f074982593f0d7dc

src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
shubham1g5: https://github.com/dimagi/commcare-core/pull/1454/commits/5b8b77bc057d7afae0e4ae15559daeb1b82898f6

src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
```diff
@@ -123,22 +124,29 @@ protected void prepareEntitiesInternal(List<Entity<TreeReference>> entities) {
     }
 
     @Override
-    public void cacheEntities(List<Entity<TreeReference>> entities) {
+    public void cacheEntities(List<Entity<TreeReference>> entities, Boolean skipLazyLoad) {
```
ctsims: Curiosity: is there a reason you went with Object booleans (Kotlin compatibility maybe) rather than raw values?

src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
```diff
@@ -199,10 +199,11 @@ public final boolean isEntitySetReady() {
      * Caches the provided entities. Default implementation throws RuntimeException.
      * Subclasses should override this method if they support caching.
      *
-     * @param entities List of entities to cache
+     * @param entities     List of entities to cache
+     * @param skipLazyLoad true if we want to calculate even lazy-loading properties as part of the process
      * @throws RuntimeException if caching is not supported
      */
-    public void cacheEntities(List<Entity<TreeReference>> entities) {
+    public void cacheEntities(List<Entity<TreeReference>> entities, Boolean skipLazyLoad) {
```
ctsims, line 206: Probably a nitpick, but it feels a bit to me like the "force this to be synchronous" call should normally either be a separate named function, or potentially part of the constructor (or a 'set' function) / nature of the factory. I think it's a little funny for individual method signatures to be either synchronous and blocking or asynchronous based on an arg. 

Reading the implementations it does look like we might already know whether we want/need it to be blocking when it's created / before this method is called.

ctsims (approved):
Re-reviewed commit by commit and I think my core feedback was all addressed. Only remaining input is that it'd be really great to get how this works written down somewhere, and make sure that the underlying states are communicated effectively for debugging. There are now a lot of factors involved in how things get processed.

src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
shubham1g5: Ahh It wasn't a conscious choice, also Kotlin `Boolean` does compile to primitive `boolean` as well. Therefore seems better to use primitives here - 073205487a678f944fda74223523f148990d42cc

src/main/java/org/commcare/cases/entity/NodeEntityFactory.java
shubham1g5: https://github.com/dimagi/commcare-core/pull/1454/commits/61d0a096f4679bc09d16739bf4f55fc03a4fb4ab

shubham1g5:
duplicate this PR fd33ba772f72e243f9ede5c8477ee44df667d16e 61d0a096f4679bc09d16739bf4f55fc03a4fb4ab

shubham1g5:

PR #1459, opened 2025-02-18
Create micro image widget

avazirna:

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/javarosa/core/model/data/Base64ImageData.java
+++ b/src/main/java/org/javarosa/core/model/data/Base64ImageData.java
@@ -1,0 +1,83 @@
+package org.javarosa.core.model.data;
+
+import org.commcare.modern.util.Pair;
+import org.javarosa.core.util.DataUtil;
+import org.javarosa.core.util.externalizable.DeserializationException;
+import org.javarosa.core.util.externalizable.ExtUtil;
+import org.javarosa.core.util.externalizable.PrototypeFactory;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+
+/**
+ * A response to a micro widget question, it returns a Pair of Strings, with the first storing the name of the file
+ * and the second, a Base64 encoded image
+ *
+ * @author avazirna
+ */
+public class Base64ImageData implements IAnswerData {
+    private Pair<String, String> base64EncodedImage;
+
+    /**
+     * Empty Constructor, necessary for dynamic construction during deserialization.
+     * Shouldn't be used otherwise.
+     */
+    public Base64ImageData() {}
+    public Base64ImageData(Pair<String, String> s) {
+        setValue(s);
+    }
+
+    @Override
+    public IAnswerData clone() {
+        return new Base64ImageData(base64EncodedImage);
+    }
+
+    @Override
+    public void setValue(Object o) {
+        if (o == null) {
+            throw new NullPointerException("Attempt to set an IAnswerData class to null.");
+        }
+        base64EncodedImage = (Pair<String, String>)o;
+    }
+
+    @Override
+    public Pair<String, String> getValue() {
+        return base64EncodedImage;
+    }
+
+    @Override
+    public String getDisplayText() {
+        return base64EncodedImage.first;
+    }
+
+    @Override
+    public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
+        base64EncodedImage = new Pair<>(ExtUtil.readString(in), ExtUtil.readString(in));
+    }
+
+    @Override
+    public void writeExternal(DataOutputStream out) throws IOException {
+        ExtUtil.writeString(out, base64EncodedImage.first);
+        ExtUtil.writeString(out, base64EncodedImage.second);
+    }
+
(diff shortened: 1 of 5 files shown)


src/main/java/org/javarosa/core/model/data/Base64ImageData.java
```diff
@@ -0,0 +1,87 @@
...
+        return new Base64ImageData(base64EncodedImage);
+    }
+
+    @Override
+    public void setValue(Object o) {
+        if (o == null) {
+            throw new NullPointerException("Attempt to set an IAnswerData class to null.");
+        }
+        base64EncodedImage = (Pair<String, String>)o;
+    }
```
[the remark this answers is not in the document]
shubham1g5: +1 on more argument validation here along with making sure none of the two objects in the pair are `null`

src/main/java/org/javarosa/core/model/data/Base64ImageData.java
```diff
@@ -0,0 +1,91 @@
...
+    public UncastData uncast() {
+        return new UncastData(toString());
+    }
+
+    @Override
+    public Base64ImageData cast(UncastData data) throws IllegalArgumentException {
+        String[] elements = DataUtil.splitOnSpaces(data.value);
+        if (elements.length != 2) {
+            throw new IllegalArgumentException("Two elements are expected, found "+ elements.length);
+        }
```
shubham1g5, line 79: this validation needs to happen when we init/set `Base64ImageData` instead

src/main/java/org/javarosa/core/util/DataUtil.java
```diff
@@ -87,4 +87,11 @@ public static String[] splitOnColon(String s) {
     public static String[] splitOnPlus(String s) {
         return s.split("[+]");
     }
+
+    public static boolean isImageFile(String filename){
+        String lowerFilename = filename.toLowerCase();
+        return lowerFilename.endsWith(".jpg") || lowerFilename.endsWith(".jpeg") ||
+                lowerFilename.endsWith(".png") || lowerFilename.endsWith(".gif") ||
+                lowerFilename.endsWith(".bmp") || lowerFilename.endsWith(".webp");
+    }
```
shubham1g5, line 96: where is this list coming from ? . Also better to store this list in some sort of collection  for easy maintaineability.

src/main/java/org/javarosa/xform/util/XFormAnswerDataSerializer.java
```diff
@@ -167,6 +170,11 @@ public Object serializeAnswerData(GeoPointData data) {
         return data.getDisplayText();
     }
 
+    public Object serializeAnswerData(Base64ImageData data) {
+        Pair<String, String> base64ImageData = (Pair<String, String>) data.getValue();
+        return base64ImageData.first + " " + base64ImageData.second;
```
shubham1g5, line 175: should just call `toString()` here instead ?

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -885,12 +887,9 @@ private void saveInstanceNode(Element instance) {
     }
 
     protected QuestionDef parseUpload(IFormElement parent, Element e, int controlUpload) {
-        Vector<String> usedAtts = new Vector<>();
-        usedAtts.addElement("mediatype");
-
```
shubham1g5, line 890: why was it removed ?

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -904,7 +903,16 @@ protected QuestionDef parseUpload(IFormElement parent, Element e, int controlUpl
     }
 
     protected QuestionDef parseControl(IFormElement parent, Element e, int controlType) {
```
shubham1g5, line 905: this is an overloaded method for another `parseControl` method and should not contain any logic except calling the other `parseControl` method

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -904,7 +903,16 @@ protected QuestionDef parseUpload(IFormElement parent, Element e, int controlUpl
...
-        return parseControl(parent, e, controlType, new Vector<String>());
+        QuestionDef question = parseControl(parent, e, controlType, new Vector<String>());
+
+        if (controlType == Constants.CONTROL_INPUT) {
+            String mediaType = e.getAttributeValue(null, MEDIA_TYPE_ATTR);
+            String appearance = e.getAttributeValue(null, APPEARANCE_ATTR);
+            if ("image/*".equals(mediaType) && MICRO_IMAGE_APPEARANCE_ATTR.equals(appearance)) {
+                question.setControlType(Constants.CONTROL_MICRO_IMAGE);
+            }
+        }
```
shubham1g5, line 914: It feel un-necessary to do any manipulation here with `controlType` but can just decide the widget based on the appearance attribute itself. But if we do want to set a distinct `controlType` for micro image, it's best to do in a new top level method say `parseInput` (similar to `parseUpload` above)

shubham1g5 (changes requested):

PR #1460, opened 2025-02-19
Adds a method to get all related records from case graph

shubham1g5:
## Product Description

Required for https://github.com/dimagi/commcare-android/pull/2955

### Automated test coverage

PR adds a few tests to verify basic functionality

### Rollback instructions

### Review

### Duplicate PR

## Summary by CodeRabbit

- **New Features**
	- Introduced bulk record retrieval to streamline the fetching of multiple record identifiers.
	- Enhanced connectivity tracking to efficiently discover and link related records.

- **Tests**
	- Updated testing scenarios to incorporate relationship outcomes, improving the validation of case purging workflows and ensuring data integrity.

### DIFF
--- a/src/main/java/org/javarosa/core/util/DAG.java
+++ b/src/main/java/org/javarosa/core/util/DAG.java
@@ -1,8 +1,12 @@
 package org.javarosa.core.util;
 
+import org.commcare.modern.util.Pair;
+
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Hashtable;
+import java.util.LinkedList;
+import java.util.Set;
 import java.util.Stack;
 import java.util.Vector;
 
@@ -136,6 +140,33 @@ public class DAG<I, N, E> {
             }
         }
         return roots;
+    }
+
+    public Set<I> getRelatedRecords(Set<I> recordIds) {
+        Set<I> visited = new HashSet<>();
+        LinkedList<I> queue = new LinkedList<>(recordIds);
+        while (!queue.isEmpty()) {
+            I current = queue.poll();
+            if(visited.contains(current)){
+                continue;
+            }
+            visited.add(current);
+            addNeighbors(edges, current, queue, visited);
+            addNeighbors(inverseEdges, current, queue, visited);
+        }
+        return visited;
+    }
+
+    private void addNeighbors(Hashtable<I, Vector<Edge<I, E>>> edges, I current, LinkedList<I> queue,
+            Set<I> visited) {
+        if (edges.containsKey(current)) {
+            Vector<Edge<I, E>> neighbors = edges.get(current);
+            for (Edge<I, E> neighbor : neighbors) {
+                if(!visited.contains(neighbor.i)){
+                    queue.add(neighbor.i);
+                }
+            }
+        }
     }
 
     public static class Edge<I, E> {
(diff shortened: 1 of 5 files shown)


src/main/java/org/javarosa/core/util/DAG.java
```diff
@@ -138,6 +142,33 @@ public Stack<I> getSinks() {
...
+                continue;
+            }
+            visited.add(current);
+            addNeighbors(edges, current, queue, visited);
+            addNeighbors(inverseEdges, current, queue, visited);
+        }
+        return visited;
+    }
+
+    private void addNeighbors(Hashtable<I, Vector<Edge<I, E>>> edges, I current, LinkedList<I> queue,
```
ctsims, line 160: "Add" is likely to be a misleading name for this method. As stated it sounds like it would be adding neighbors to the DAG itself. 

I'd probably just rename this to `enqueueNeighbors` or `accumulateNeighbors`, but it should be clear that the method itself is stateless.

ctsims (changes requested):
Very minor feedback on code structure / naming, once addressed everything else is good.

src/main/java/org/javarosa/core/util/DAG.java
```diff
@@ -138,6 +142,33 @@ public Stack<I> getSinks() {
         return roots;
     }
 
+    public Set<I> getRelatedRecords(Set<I> recordIds) {
```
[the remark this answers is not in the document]
shubham1g5: https://github.com/dimagi/commcare-core/pull/1460/commits/24da04f8e80bb90348f69698c9826ba0a2063f48

src/main/java/org/javarosa/core/util/DAG.java
shubham1g5: https://github.com/dimagi/commcare-core/pull/1460/commits/24da04f8e80bb90348f69698c9826ba0a2063f48

src/main/java/org/javarosa/core/util/DAG.java
```diff
@@ -138,6 +140,42 @@ public Stack<I> getSinks() {
         return roots;
     }
 
+    /**
+     * Performs a breadth-first search (BFS) to find all related records
```
ctsims, line 144: More nitpicking:

1 - the BFS part of this is (from my perspective) an implementation detail that I don't think you want. We should be able to change the implementation of the search in the future and stating in the docstring that it's BFS locks us in (or 'should' lock us in) to that behavior in the future.
2 - "Record" is a downstream implementation detail here (IE: The use of the DAG holds records in your case, but that's not fundamentally part of the data type), and shouldn't be used in the descriptions or the variable names. All of the other methods use the shared vocabulary of the DAG (Node, Edge, etc), and this method should do the same.
3 - The big thing this docstring needs to communicate isn't how it's doing it, just a more precise description of what the outcomes and concepts are. Words like "Related" aren't really appropriate for the root datatype, so it's ambiguous what the method returns. I think you just want a specific sentence here like "Return all nodes which are reachable from any of the set of input nodes." I'm not sure if "reachable" in this context is actually quite right since you're traversing edges both directions, there might be another word that formalizes edge traversal in either direction.

ctsims (approved):
Note: I'd like an update to the vocab still, but can come after merge

src/main/java/org/javarosa/core/util/DAG.java
shubham1g5: that makes a lot of sense, corrected here - https://github.com/dimagi/commcare-core/pull/1460/commits/6d9ea681d2ec5f58edaaa3de11b0c7d1e7df99a0

avazirna (approved):

PR #1488, opened 2025-06-19
Inline Case Search Breadcrumb Display Bug Fix

Jtang-1:
## Product Description

This fixes a bug where when a module uses "inline case search", the breadcrumb would not display the selected case.

## Technical Summary

[Jira Ticket](https://dimagi.atlassian.net/browse/USH-5823)
[Formplayer PR](https://github.com/dimagi/formplayer/pull/1720)

The theory of why this bug happens is because in “Inline Case Search”, the <post> element is defined within form entry instead of the case list. As a result, case claiming is deferred until form entry, meaning the case is not yet available when trying to resolve the breadcrumb — leading to the exception and incorrect UI behavior.

This solution uses the `caseSearchStorage` to fetch the case name. `caseSearchStorage` is only populated if the app also uses `cc-index-case-search-results` ([docs](https://dimagi.atlassian.net/wiki/spaces/GS/pages/2146607335/Indexing+on+Case+Search+Results)). This means that the bug will continue to be present if the module uses "Inline Case Search" but the app does not use `cc-index-case-search-results`. However, this app property will likely be used with case search.

### Safety story

locally tested. It introduces logic to retrieve caseSearchStorage, but includes safeguards for scenarios where it does not exist. In those cases, it will fall back to the pre-existing behavior from before this PR. No changes are made to existing data, and the defect blast radius is limited by maintaining prior logic when new conditions aren't met.

### Automated test coverage

no automated test exists for checking displayed breadcrumb value.

### QA Plan

[QA-7869](https://dimagi.atlassian.net/browse/QA-7869)

### Rollback instructions
The FP PR depends on this so if this is reverted, the FP PR will also need to be reverted.

### Review

### Duplicate PR

### DIFF
--- a/src/cli/java/org/commcare/util/screen/EntityScreen.java
+++ b/src/cli/java/org/commcare/util/screen/EntityScreen.java
@@ -1,6 +1,7 @@
 package org.commcare.util.screen;
 
 import org.commcare.cases.entity.EntityUtil;
+import org.commcare.cases.model.Case;
 import org.commcare.cases.query.QueryContext;
 import org.commcare.cases.query.queryset.CurrentModelQuerySet;
 import org.commcare.core.interfaces.UserSandbox;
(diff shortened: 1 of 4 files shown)


Jtang-1:
@coderabbitai review

PR #1492, opened 2025-06-27
Fix Nepali date widget bug

avazirna:

### Rollback instructions

### Review

### Duplicate PR

### DIFF
--- a/src/main/java/org/javarosa/xform/util/CalendarUtils.java
+++ b/src/main/java/org/javarosa/xform/util/CalendarUtils.java
@@ -354,7 +354,7 @@ public class CalendarUtils {
             cd.setTimeZone(DateUtils.timezone());
         }
         long dateInMillis = cd.getTime().getTime();
-        DateTimeZone timezoneObject = DateTimeZone.forOffsetMillis(cd.getTimeZone().getRawOffset());
+        DateTimeZone timezoneObject = DateTimeZone.forTimeZone(cd.getTimeZone());
         return fromMillis(dateInMillis, timezoneObject);
     }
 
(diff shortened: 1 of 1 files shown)


Jignesh-dimagi:
@avazirna Is that right understanding that there was error when converting Gregorian/English date to Nepali date and this was recursive due to moving back and forth? If this is the case, how it was calculating correctly from Nepali to Gregorian/English date?

avazirna:
> @avazirna Looks like tests are failing on this PR.

@shubham1g5 fixed - [e831bea](https://github.com/dimagi/commcare-core/pull/1492/commits/e831bea0834bb162a8b0ac264857003736c51104)

src/test/java/org/javarosa/xform/util/test/CalendarTests.java
```diff
@@ -35,6 +55,33 @@ public void testTimesFallOnSameDate() {
         assertSameDate(endOfDay, beginningOfDay);
     }
 
+    // millis <=> date is different in every timezone
+    @Test
+    public void testConvertToNepaliString() {
```
shubham1g5, line 60: can you confirm these tests are failing without the changes made in `CalendarUtils` (I tried on my end but it seems to pass) ? I think we want to confirm that this test indeed reproduces the error we are trying to fix.

avazirna: @shubham1g5 the tests were considering the raw offset, which doesn't account for DST. I have updated them [334a900](https://github.com/dimagi/commcare-core/pull/1492/commits/334a9006d6b1cbb691b005c2d9ee38c1c5dd759f). Can you test this again?

shubham1g5 (approved):

PR #1469, opened 2025-04-14
Bump JSON dependency

avazirna:
### Special deploy instructions

- [ ] This PR can be deployed after merge with no further considerations.

### Rollback instructions

- [ ] This PR can be reverted after deploy with no further considerations.

### Review

- [ ] The set of people pinged as reviewers is appropriate for the level of risk of the change.

### Duplicate PR
Automatically duplicate this PR as defined in [contributing.md](https://github.com/dimagi/commcare-core/blob/master/.github/contributing.md).

PR #1494, opened 2025-07-28
Duplicate of Bump JSON dependency to version 20250517

avazirna:
## Product Description
Copy of PR #1469


avazirna:
> No concerns, but we should make sure that formplayer builds on top of these changes.

I tested it and it was successful.

PR #1500, opened 2025-09-10
Appends info to crash messaging

shubham1g5:
## Product Description

We are loosing the extra info we log into `Logger.exception(info, e)` into Crashlytics messaging, which makes it quite hard to recognise where a particular non-fatal is being logged from in Crashlytics. 

### Rollback instructions

### Review

### Duplicate PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures exception messages in logs are consistently formatted, even when the original message is missing.
  * Preserves stack traces while improving how related causes are recorded for clearer diagnostics.

* **Refactor**
  * Streamlined error logging to reduce ambiguity in logged messages without changing runtime behavior.

### DIFF
--- a/src/main/java/org/javarosa/core/services/Logger.java
+++ b/src/main/java/org/javarosa/core/services/Logger.java
@@ -53,10 +53,15 @@ public class Logger {
 
     public static void exception(String info, Throwable e) {
         e.printStackTrace();
-        log(LogTypes.TYPE_EXCEPTION, (info != null ? info + ": " : "") + WrappedException.printException(e));
+        info = info != null ? info + ": " : "";
+        log(LogTypes.TYPE_EXCEPTION, info + WrappedException.printException(e));
         if (logger != null) {
             try {
-                logger.logException(e);
+                String message = e.getMessage();
+                if (message == null) {
+                    message = "";
+                }
+                logger.logException(new Exception(info + message, e));
             } catch (RuntimeException ex) {
                 logger.panic();
             }


src/main/java/org/javarosa/core/services/Logger.java
```diff
@@ -53,10 +53,15 @@ public static void log(String type, String message) {
 
     public static void exception(String info, Throwable e) {
         e.printStackTrace();
```
Jignesh-dimagi, line 55: Please remove this

src/main/java/org/javarosa/core/services/Logger.java
```diff
@@ -53,10 +53,15 @@ public static void log(String type, String message) {
...
-        log(LogTypes.TYPE_EXCEPTION, (info != null ? info + ": " : "") + WrappedException.printException(e));
+        info = info != null ? info + ": " : "";
+        log(LogTypes.TYPE_EXCEPTION, info + WrappedException.printException(e));
```
Jignesh-dimagi, line 57: It is required to log? Generally log is for data point and events while exception should not be here?

src/main/java/org/javarosa/core/services/Logger.java
```diff
@@ -53,10 +53,15 @@ public static void log(String type, String message) {
...
         if (logger != null) {
             try {
-                logger.logException(e);
+                String message = e.getMessage();
+                if (message == null) {
+                    message = "";
+                }
+                logger.logException(new Exception(info + message, e));
```
Jignesh-dimagi, line 64: Should we only put info and message will anyways get logged by exception?

Jignesh-dimagi (changes requested):

src/main/java/org/javarosa/core/services/Logger.java
shubham1g5: if we only log `info`, the original `message` will be logged as `Caused By exception $message` which won't be directly searcheable in Crashlytics (think crashlytics only allow search by exception title and not on nested exceptions)

Jignesh-dimagi (approved):

PR #1515, opened 2026-01-21
Increase connection timeout 

avazirna:
## Product Description
A project is experiencing issues submitting forms that contain hundreds of multimedia files. The multipart POST request seems to be timing out during execution. This PR increases the connection timeout as an experimental measure to determine whether it enables successful form submission. This change is exploratory and is not intended to be merged at this point. 

Ticket: https://dimagi.atlassian.net/browse/SAAS-19090


avazirna:
Closing this out. We shared an APK with the project team and haven't heard back since, so it appears the issue has been resolved.

PR #1537, opened 2026-05-06
Remove support for dynamic search

MartinRiese:
## Product Description

Commcare HQ does not set the dynamic search flag anymore. It has been removed in this PR: https://github.com/dimagi/commcare-hq/pull/37634. Formplayer should behave the same way as
if the flag is set to false.

## Technical Summary

https://dimagi.atlassian.net/browse/USH-6541

### Rollback instructions

### Review

## Summary by CodeRabbit

* **Refactor**
  * Removed the dynamic search capability from query management.
  * Replaced with default search functionality in remote query processing.
  * Updated related data models and parsing logic to align with the new search approach.

### DIFF
--- a/src/cli/java/org/commcare/util/screen/QueryScreen.java
+++ b/src/cli/java/org/commcare/util/screen/QueryScreen.java
@@ -63,7 +63,6 @@ public class QueryScreen extends Screen {
 
     private boolean defaultSearch;
 
-    private boolean dynamicSearch;
     private boolean searchOnClear;
 
     private IStorageUtilityIndexed<Case> caseSearchStorage;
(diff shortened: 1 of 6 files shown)


shubham1g5 (approved):

MartinRiese:
@shubham1g5 how is the formplayer test supposed to work? Is that just for when you do not have a formplayer change? And I can ignore it here?

shubham1g5:
> how is the formplayer test supposed to work? Is that just for when you do not have a formplayer change? And I can ignore it here?

Think it checks this change against `master` for FP, but given you have added FP changes in a different PR and tests are passing there, good to ignore this. 

FYI @dannyroberts in case you have thoughts on how to improve this for such changes.

PR #1538, opened 2026-05-15
Only prime cache, if it is available

MartinRiese:
## Product Description

Add a check in the cache priming code. Only do the priming, if there is a cache set.

## Technical Summary

* https://dimagi.atlassian.net/browse/USH-6551 (Original Issue)
* https://dimagi.atlassian.net/browse/USH-6555 (fix implementation story)
* https://docs.google.com/document/d/1gwfS8snRbByB69vc-unPk30mOhB9vX6uNpkq_8yxaAs/edit?tab=t.0 (Write up)

### Safety story

Deploy to staging and test variations:
* Web apps:
  * clickable icon ✅ 
  * sort defined, not defined ✅ 
* mobile app:
  * sort works for enum field
  * sort works for other fields.

### Automated test coverage

Added a test to check for future regressions which doubles as check on the fix.

### Rollback instructions

### Review

## Summary by CodeRabbit

* **Bug Fixes**
  * Optimized entity data computation to conditionally process legacy data only when cache storage is available, reducing unnecessary processing and improving performance during data loading operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dimagi/commcare-core/pull/1538?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
+++ b/src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
@@ -172,13 +172,15 @@ public class AsyncNodeEntityFactory extends NodeEntityFactory {
     // Old cache and index pathway where we only cache sort fields
     @Deprecated
     protected void setUnCachedDataOld(List<Entity<TreeReference>> entities) {
-        for (int i = 0; i < entities.size(); i++) {
-            if (isCancelled) return;
-            AsyncEntity e = (AsyncEntity)entities.get(i);
-            for (int col = 0; col < e.getNumFields(); ++col) {
+        if (mEntityCache != null) {
+            for (int i = 0; i < entities.size(); i++) {
+                if (isCancelled) return;
+                AsyncEntity e = (AsyncEntity) entities.get(i);
+                for (int col = 0; col < e.getNumFields(); ++col) {
                     e.getSortField(col);
+                }
+                updateProgress(PHASE_UNCACHED_CALCULATION, i, entities.size());
             }
-            updateProgress(PHASE_UNCACHED_CALCULATION, i, entities.size());
         }
     }
 
--- a/src/test/java/org/commcare/cases/entity/AsyncNodeEntityFactoryTest.java
+++ b/src/test/java/org/commcare/cases/entity/AsyncNodeEntityFactoryTest.java
@@ -1,0 +1,77 @@
+package org.commcare.cases.entity;
+
+import org.commcare.suite.model.Detail;
+import org.commcare.suite.model.DetailField;
+import org.commcare.suite.model.Text;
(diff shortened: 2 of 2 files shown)


src/main/java/org/commcare/cases/entity/AsyncNodeEntityFactory.java
```diff
@@ -172,13 +172,15 @@ protected void setUnCachedData(List<Entity<TreeReference>> entities) {
     // Old cache and index pathway where we only cache sort fields
     @Deprecated
     protected void setUnCachedDataOld(List<Entity<TreeReference>> entities) {
-        for (int i = 0; i < entities.size(); i++) {
-            if (isCancelled) return;
-            AsyncEntity e = (AsyncEntity)entities.get(i);
-            for (int col = 0; col < e.getNumFields(); ++col) {
+        if (mEntityCache != null) {
```
MartinRiese, line 175: Not sure if we would want put this behind a feature flag for now.

esoergel: Looks like `commcare-core` doesn't have access to feature flags, is that right?  We could pass through a boolean if we needed to - might be nice to easily toggle on and off for performance testing, but coordinating across three repos is tedious, and Shubham seemed optimistic about this change, so maybe not worth the effort.  Dunno, I could go either way.

shubham1g5: yeah I am quite confident for this to be rolloed out without a feature flag.

MartinRiese:
@shubham1g5 Do you have a suggestion how to test the mobile app for regressions (making sure sort works as intended)?

shubham1g5:
@MartinRiese I don't think there is a need to do, the cache is always not-null for mobile when we use `AsyncNodeEntityFactory` so that change is essentially a no-op for mobile.

src/test/java/org/commcare/cases/entity/AsyncNodeEntityFactoryTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.ArrayList;
+import java.util.Hashtable;
+import java.util.List;
+import java.util.Vector;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Regression test for USH-6551: sort field expressions must not be evaluated during the cache
```
shubham1g5: Curious if it's a common practice on USS to add Jira ticket # in code comments, we have refrained from doing so on mobile team in past as the code is open source and putting these refrences make it cryptic for anyone external looking into the code.

shubham1g5 (dismissed):

esoergel (approved):
approving last commit

PR #1546, opened 2026-06-30
[SAAS-19868] Add face capture widget

avazirna:
## Summary
Adds a dedicated face-capture control type to the form engine, so the Android client can route `<upload>` questions to the new face-capture widget.

- New `Constants.CONTROL_FACE_CAPTURE` (= 15).
- `XFormParser.parseUpload` now maps `mediatype="image/*"` **with** `appearance="face"` to `CONTROL_FACE_CAPTURE`; plain `image/*` still maps to `CONTROL_IMAGE_CHOOSE`.
- Moved the `mediatype` `usedAtts` registration into `parseControl` so it's tracked whenever the attribute is present.

## Notes
This is the core-side dependency for the Android face-capture widget (`saas-19868`); the android PR builds on this control type.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for a new face-capture option in upload controls.
  * Upload controls can now better distinguish image uploads from face-capture uploads based on available settings.

* **Bug Fixes**
  * Improved handling of media-related upload settings so image, audio, video, and document uploads are recognized more consistently.
  * Reduced unnecessary warnings for valid upload attributes.

### DIFF
--- a/src/main/java/org/javarosa/core/model/Constants.java
+++ b/src/main/java/org/javarosa/core/model/Constants.java
@@ -80,6 +80,7 @@ public class Constants {
     public static final int CONTROL_AUDIO_CAPTURE = 12;
     public static final int CONTROL_VIDEO_CAPTURE = 13;
     public static final int CONTROL_DOCUMENT_UPLOAD = 14;
+    public static final int CONTROL_FACE_CAPTURE = 15;
 
     /**
      * constants for xform tags
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -87,6 +87,7 @@ public class XFormParser {
     private static final String BIND_ATTR = "bind";
     private static final String REF_ATTR = "ref";
     private static final String EVENT_ATTR = "event";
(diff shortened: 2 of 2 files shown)


shubham1g5:
Do we really need to add a new control type given all we are capturing is an Image at the end ? (And only difference is the UI we use to capture that)

avazirna:
> Do we really need to add a new control type given all we are capturing is an Image at the end ? (And only difference is the UI we use to capture that)

The intention is to have a cleaner path from the parser to widget rendering, considering that this widget will have its own representation on HQ, instead of just extending the widget selection in `WidgetFactory`. But I'm happy to reconsider this.

shubham1g5:
Think we normally don't create new control factors based on appearance attributes so the pattern used here seems to deviate from what we follow for variations based on appeatance attributes without a solid reason.

avazirna:
> Think we normally don't create new control factors based on appearance attributes so the pattern used here seems to deviate from what we follow for variations based on appeatance attributes without a solid reason.

Make sense. I will revisit this decision, and propose some documentation around this too

avazirna:
Based on the feedback, the widget will be routed through an image widget appearance attribute.

PR #1549, opened 2026-08-04
Add a function that gets the menu title for screen

MartinRiese:
## Product Description

getBestTitle is not always the menu title. Add a function that returns exactly that for logging to datadog and google analytics (https://github.com/dimagi/formplayer/pull/1794)

## Technical Summary

https://dimagi.atlassian.net/browse/USH-6605

### Safety story

Tested on staging

### Automated test coverage

Added tests

-  This PR can be deployed after merge with no further considerations.

### Rollback instructions

- The formplayer PR https://github.com/dimagi/formplayer/pull/1794 depends on this change and needs to reverted as well.

### Review

## Summary by CodeRabbit

* **New Features**
  * Added support for identifying the title of the most recently selected menu in the current session.
  * Returns no menu title when the session does not contain a matching menu selection.

### DIFF
--- a/src/cli/java/org/commcare/util/screen/ScreenUtils.java
+++ b/src/cli/java/org/commcare/util/screen/ScreenUtils.java
@@ -2,14 +2,18 @@ package org.commcare.util.screen;
 
 import org.commcare.modern.session.SessionWrapper;
 import org.commcare.session.SessionFrame;
+import org.commcare.suite.model.Menu;
 import org.commcare.suite.model.StackFrameStep;
+import org.commcare.suite.model.Suite;
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.condition.IFunctionHandler;
 import org.javarosa.core.model.data.GeoPointData;
 import org.javarosa.core.services.locale.Localization;
+import org.javarosa.core.services.locale.Localizer;
 import org.javarosa.core.util.NoLocalizedTextException;
 import org.javarosa.xpath.XPathException;
 
+import java.util.Hashtable;
 import java.util.Vector;
 
 /**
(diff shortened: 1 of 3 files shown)


src/cli/java/org/commcare/util/screen/ScreenUtils.java
```diff
@@ -74,6 +78,33 @@ public static String getBestTitle(SessionWrapper session) {
         return bestTitle;
     }
 
+    public static String getMenuTitle(SessionWrapper session) {
```
esoergel, line 82: I don't have enough context to speak confidently about how this should be done, but the current implementation looks pretty roundabout (I do see that it mirrors the function above).  I asked Claude to review this and it gave me back the following feedback.   Number 3 is kinda in the direction I was thinking, though I was hoping there'd be an even more direct route to know what menu we're in.

   1. **No exception handling.** `Menu.getName().evaluate()` calls `Localization.get()` (throws `NoLocalizedTextException`) or evaluates an XPath with a null `EvaluationContext` (throws
  `XPathException`) — see `Text.java:209-235`. `getBestTitle` catches both and returns null, with a comment about localization resources being absent mid-update. Since this method evaluates *every*
  menu title in the app, one bad title anywhere blows up a call whose only purpose is Datadog/GA logging.

  2. **`getName()` isn't null-safe.** `Menu.getDisplayText(ec)` (`Menu.java:214-219`) null-guards `display.getText()`; `getName().evaluate()` doesn't. Exposure is again app-wide because the map covers
  all menus.

  3. **The full map is unnecessary.** `Suite` already indexes menus by id — `getMenusWithId` is a hashtable lookup (`Suite.java:103`). You only need one id, and iterating the steps *backwards* lets you
  stop at the first menu found instead of evaluating N menu titles and scanning M steps.

  4. **`return menuTitle;` at the end** is a null return dressed up as a variable return.

  5. **No test**, though `MenuTests` / `AppStructureTests` have app fixtures that would make one cheap.

conroy-ricketts (commented):
No additional feedback from my side

src/cli/java/org/commcare/util/screen/ScreenUtils.java
```diff
@@ -74,6 +79,42 @@ public static String getBestTitle(SessionWrapper session) {
...
+        // Walk backwards looking for the most recent step that is an actual menu
+        Vector<StackFrameStep> steps = session.getFrame().getSteps();
+        for (int i = steps.size() - 1; i >= 0; i--) {
+            StackFrameStep step = steps.elementAt(i);
+            if (!SessionFrame.STATE_COMMAND_ID.equals(step.getType())) {
+                continue;
+            }
+
+            for (Suite s : session.getPlatform().getInstalledSuites()) {
```
esoergel, line 91: Nit, this duplicates this method - could pull that in here and save a few lines

https://github.com/dimagi/commcare-core/blob/130df00962a289381a8e0936c3ea5d3f53d96f73/src/main/java/org/commcare/session/CommCareSession.java#L155-L164

esoergel (approved):

